PR #14, opened 2015-07-31
pull jenkins rpm from rhui

bparees:


bparees:
@mfojtik ptal.

1.6/Dockerfile
```diff
@@ -24,31 +24,25 @@ LABEL k8s.io.description="Jenkins is a continuous integration server" \
...
-RUN  usermod -m -d "$JENKINS_HOME" jenkins && \
-  chown -R jenkins "$JENKINS_HOME" && \
-  chmod o+rw /var/lib/jenkins
+COPY run.sh /usr/local/bin/run.sh
+COPY ./contrib/ /opt/openshift
 
-RUN curl -L https://github.com/openshift/origin/releases/download/v1.0.2/openshift-origin-v1.0.2-325c7b7-linux-amd64.tar.gz | tar -zx && \
+RUN curl http://pkg.jenkins-ci.org/redhat/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
+    rpm --import http://pkg.jenkins-ci.org/redhat/jenkins-ci.org.key && \
+    yum install -y tar zip unzip java-1.7.0-openjdk jenkins-1.609 && \
```
mfojtik: do you need to specify the version of jenkins? it should pull the latest one, right?

1.6/Dockerfile
```diff
@@ -24,31 +24,25 @@ LABEL k8s.io.description="Jenkins is a continuous integration server" \
...
 #RUN  curl -L -o /tmp/git.hpi http://updates.jenkins-ci.org/latest/git.hpi
 #RUN  curl -L -o /tmp/git-client.hpi http://updates.jenkins-ci.org/latest/git-client.hpi
 #RUN  curl -L -o /tmp/scm-api.hpi http://updates.jenkins-ci.org/latest/scm-api.hpi
 #RUN  curl -L -o /tmp/ssh-credentials.hpi http://updates.jenkins-ci.org/latest/ssh-credentials.hpi
 #RUN  curl -L -o /tmp/credentials.hpi http://updates.jenkins-ci.org/latest/credentials.hpi
 
-RUN  usermod -m -d "$JENKINS_HOME" jenkins && \
-  chown -R jenkins "$JENKINS_HOME" && \
-  chmod o+rw /var/lib/jenkins
+COPY run.sh /usr/local/bin/run.sh
```
mfojtik: Why you're moving this up? Now when you change the run.sh you have to reinstall the jenkins on next docker build.

1.6/run.sh
```diff
@@ -24,7 +24,7 @@ fi
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_OPTS -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
+   exec java $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
```
mfojtik: just out of curiosity... what is this magic?

1.6/Dockerfile
bparees: i want it to match the rhel version, otherwise the centos version will be newer than what's in the rhel image.

(which is also why i then specify the rhel version, so it doesn't end up ahead of the centos version)

1.6/Dockerfile
bparees: i thought i had a reason, but i think it was just a result of squashing the other run commands.  i'll move it below the run.

1.6/run.sh
bparees: fixes this:
https://bugzilla.redhat.com/show_bug.cgi?id=1247894
see:
https://stackoverflow.com/questions/27960996/jenkins-console-output-characters

1.6/Dockerfile
mfojtik: can't you just disable centos repo (temporary) and pull the jenkins? I'm worried that if they change the version in upstream repo, this Dockerfile will be broken.

bparees: it's not getting it from the centos repo, it's getting it from the repo that i'm installing on lines 36-37, which has a historical set of images:

bparees:
COPY command moved.

mfojtik:
LGTM

PR #15, opened 2015-08-03
tag image on test success

bparees:


bparees:
@jhadvig we missed one :(

PR #20, opened 2015-08-26
Allow jenkins plugin installation

mfojtik:
This plugin allows to install additional Jenkins plugins the same way as the official Jenkins image https://github.com/jenkinsci/docker does. 

To produce a new Docker image with more plugins installed, just write Dockerfile like this:

```Dockerfile
FROM openshift/jenkins-16-centos7
COPY plugins.txt /opt/openshift/configuration/plugins.txt
RUN /usr/local/bin/plugins.sh /opt/openshift/configuration/plugins.txt
```

You can list the plugins you want to installed by `pluginID:pluginVersion` in `plugins.txt` file:

```
credentials:1.22
plain-credentials:1.1
ssh-agent:1.5
github:1.9.1
github-api:1.68
scm-api:0.2
git-client:1.10.1
git:2.3.4
ghprb:1.27
```

(this is an example of installing the Github Pull request plugin)


mfojtik:
@gabemontero @bparees PTAL

mfojtik:
[test]

mfojtik:
@gabemontero @bparees I've built my custom image with Github PR tracker plugin installed here:

```console
docker pull docker.io/mfojtik/jenkins-ci
```

mfojtik:
[test] again

bparees:
the new scripts belong in "contrib"

mfojtik:
@bparees do we want to copy that script to `/opt/openshift` ?

bparees:
probably not (though i don't love copying a "run.sh" into /usr/local/bin either.  can we name it better?)

maybe we need two sub-dirs under contrib.. contrib/openshift and contrib/jenkins ?  (so you can still do the COPY command as you are, roughly)

1.6/Dockerfile.rhel7
```diff
@@ -46,9 +45,11 @@ RUN yum install -y tar zip unzip java-1.7.0-openjdk jenkins-1.609.1 && \
     mv oc /usr/local/bin && \
     rm oadm && \
     rm openshift && \
-    chmod -R og+rw /opt/openshift
+    mkdir -p /opt/openshift && chmod -R og+rw /opt/openshift && \
+    mkdir -p /usr/share/jenkins && chmod -R og+rw /usr/share/jenkins
 
-VOLUME ["/var/lib/jenkins"]
+COPY ./contrib/openshift /opt/openshift
+COPY ./contrib/jenkins /usr/local/bin
```
bparees: same comment I made to @rhcarvalho.  moving these COPYs after the chmod of the dir means the files in the dir aren't getting chmod'd.  that seems potentially problematic.

rhcarvalho: Permissions in the original image:

```
$ ls -la /opt/openshift/
total 16
drwxrwxrwx. 3 root root 4096 Aug 27 13:10 .
drwxr-xr-x. 3 root root 4096 Aug 27 13:10 ..
drwxrwxrwx. 2 root root 4096 Aug 27 13:20 configuration
-rwxrwxrwx. 1 root root 2039 Aug 27  2015 password-encoder.jar
```

With COPY after RUN:

```
$ ls -la /opt/openshift/
total 16
drwxrwxrwx. 3 root root 4096 Aug 27 13:42 .
drwxr-xr-x. 3 root root 4096 Aug 27 13:41 ..
drwxrwxr-x. 4 root root 4096 Aug 27 13:22 configuration
-rwxrwxr-x. 1 root root 2039 Aug 27 13:22 password-encoder.jar
```

And I was able to, as UID 1001, make changes to `/opt/openshift/configuration/config.xml`.

What exactly could go problematic here?

bparees: what's the permissions of config.xml?

random uids will no longer be able to create new files under the configuration directory, not sure if that could be problematic, i'd have to revisit the directory structure.

what about the "jobs" directory?  isn't that going to also not be writable by uid 1001 (or any random uid) which means the jenkins process won't be able to write changes or jobs or create new jobs?

mfojtik: @bparees @rhcarvalho I think this problem is not very related to this PR ;-) Can we move it to issues and continue discussing it there :-) I really want to have plugin installation infrastructure in place to move on the Jenkins R&D ;-)

mfojtik:
@bparees your naming change almost killed me ;-) But I rebased my changes and hopefully I did not forget to move anything.

1.6/Dockerfile.rhel7
mfojtik: I added extra RUN chmod after we copy the scripts and config. This should allow to random UID to write to that files or override the files we copied there.

mfojtik:
[test]

mfojtik:
@bparees can you have a second look? I tested this here: https://github.com/mfojtik/jenkins-ci/tree/master/jenkins-master and everything works :-)

bparees:
@mfojtik yeah, i felt bad about the version/name change, but @danmcp pointed it out and i just wanted to get it done ASAP.  Sorry :(

changes lgtm, would like to see a test run though.
[test]

bparees:
[test]

1/Dockerfile.rhel7
```diff
@@ -46,9 +43,12 @@ RUN yum install -y tar zip unzip java-1.7.0-openjdk jenkins-1.609.1 && \
     mv oc /usr/local/bin && \
     rm oadm && \
     rm openshift && \
```
bparees: trailing && here is breaking things.

mfojtik:
[test]

mfojtik:
@bparees rebased, fixed... I think this is ready to go.

bparees:
lgtm.

PR #21, opened 2015-08-26
Disable binding on HTTP port and minor tweaks to entrypoint

mfojtik:
@bparees I spend 20 minutes (!) trying to figure out, why `docker run -it mfojtik/jenkins-ci /bin/bash` does not have the plugins installed and why the `plugins.txt` is not in `/opt/openshift/configuration` where I copied it in my Dockerfile, but in `/var/lib/jenkins`.

Then I discover, that the entrypoint is copying the files on your behalf.... Which is not nice to me. And to users ;-) I think we should print something out, so the future users that will try to layer on top of this image know what is going on when they execute "docker run ... /bin/bash" ;-)

Also, I while I'm playing with Jenkins, I found the issue with OpenShift router redirecting to HTTP **very** annoying. Basically every third click I made in 'Manage Jenkins' screen gives me 503 and I have to manually replace 'http://' in browser with 'https://'...

So I'm trying to disable the HTTP port. I also found that the Jenkins URL is wrong... It is pointing to 'http://' which might be related to this as well.

### DIFF
--- a/1.6/run.sh
+++ b/1.6/run.sh
@@ -13,15 +13,15 @@ unzip -q /usr/lib/jenkins/jenkins.war -d /tmp/war
 new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password}`
 
 if [ ! -e /var/lib/jenkins/configured ]; then
-  cp -r /opt/openshift/configuration/* /var/lib/jenkins
+  echo "[INFO] Installing Jenkins configuration and plugins to ${JENKINS_HOME} ..."
+  cp -r /opt/openshift/configuration/* ${JENKINS_HOME}
   rm -rf /opt/openshift/configuration/*
-  
+
+  echo "[INFO] Creating initial 'admin' user with password 'password'"
   sed -i "s,<passwordHash>.*</passwordHash>,<passwordHash>$new_password_hash</passwordHash>,g" "/var/lib/jenkins/users/admin/config.xml"
   echo $new_password_hash > /var/lib/jenkins/password
   touch /var/lib/jenkins/configured
-fi
-
-if [ -e /var/lib/jenkins/password ]; then
+else
   # if the password environment variable has changed, update the jenkins config.
   # we don't want to just blindly do this on startup because the user might change their password via
   # the jenkins ui, so we only want to do this if the env variable has been explicitly modified from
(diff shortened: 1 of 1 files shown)


1.6/run.sh
```diff
@@ -13,15 +13,15 @@ unzip -q /usr/lib/jenkins/jenkins.war -d /tmp/war
 new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password}`
 
 if [ ! -e /var/lib/jenkins/configured ]; then
-  cp -r /opt/openshift/configuration/* /var/lib/jenkins
+  echo "[INFO] Installing Jenkins configuration and plugins to ${JENKINS_HOME} ..."
+  cp -r /opt/openshift/configuration/* ${JENKINS_HOME}
   rm -rf /opt/openshift/configuration/*
-  
+
+  echo "[INFO] Creating initial 'admin' user with password 'password'"
```
bparees: it's not necessarily "password" and i don't think you want to print the actual password, so maybe just "creating initial admin user"

bparees:
I go through the service, not the router, so i don't have that problem :)

but is preventing jenkins from binding on http the right answer here?  is it going to bind https to port 8080?  or is it going to bind to some other port that the image doesn't EXPOSE and that the Service/Route aren't going to know about?

mfojtik:
@bparees it was just an attempt to fix it (or trick jenkins to not redirect to "http")... but it doesn't work...

bparees:
so are you going to remove that bit from this PR?

1.6/run.sh
bparees: still need to update this output

1.6/run.sh
```diff
@@ -34,6 +34,11 @@ if [ -e /var/lib/jenkins/password ]; then
 fi
 rm -rf /tmp/war
 
+# Disable binding on HTTP port
+if [ -v JENKINS_OPTS ]; then
+  JENKINS_OPTS="--httpPort=-1 "
+fi
+
```
bparees: and remove this?

mfojtik: I will, but want to get the other PR merged first to prevent conflicts ;-)

On Wed, Aug 26, 2015 at 4:55 PM, Ben Parees <notifications@github.com>
wrote:

> In 1.6/run.sh
> <https://github.com/openshift/jenkins/pull/21#discussion_r37991526>:
>
> > @@ -34,6 +34,11 @@ if [ -e /var/lib/jenkins/password ]; then
> >  fi
> >  rm -rf /tmp/war
> >
> > +# Disable binding on HTTP port
> > +if [ -v JENKINS_OPTS ]; then
> > +  JENKINS_OPTS="--httpPort=-1 "
> > +fi
> > +
>
> and remove this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/jenkins/pull/21/files#r37991526>.
>



-- 
-----------------------------------------------
Michal Fojtik <mi@mifo.sk <mfojtik@mifo.sk>>
Red Hat OpenShift, Engineering

mfojtik:
Moved the changes to the other PR. Closing this one.

PR #28, opened 2015-09-07
Report curl output when plugin download failed

mfojtik:
@bparees a minor change to Jenkins plugin download that will show you `curl` error(s) when a plugin download failed.

Also a question whether we want to fail the docker build of the Jenkins image when 1 plugin failed to download or just build the image and display error summary (# plugins failed to download:)...

(my opinion is to fail)


mfojtik:
@soltysh can you review this pls?

soltysh:
LGTM, I'd say fail, without the plugin, the image doesn't make sense, imho.

bparees:
agree, fail.

changes lgtm.

PR #35, opened 2015-10-07
WIP: Import some functionality from CI demos

mfojtik:
@bparees this will import:

* Pre-install kubernetes, git, promoted-builds, credentials, etc plugins
* Add auto-slave discovery based on image stream labels
* Auto configuration of Kubernetes plugin

Currently **only** for centos7 image, for RHEL7 I have to figure out how to install the plugins ;-)

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -43,5 +43,14 @@ COPY ./contrib/jenkins /usr/local/bin
 RUN chmod -R og+rw {/opt/openshift,/var/lib/jenkins}
 VOLUME ["/var/lib/jenkins"]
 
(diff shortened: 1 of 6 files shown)


bparees:
i'm a little worried about forking the centos vs rhel image behavior.... it's going to lead to a lot of questions and we don't really have the doc in place to support the centos image behavior.

mfojtik:
@bparees agree, we have to find a way to make this work in RHEL7 image

mfojtik:
[extended:core(s2i)]

mfojtik:
Closed in favor of using the S2I approach.

PR #36, opened 2015-10-08
Make the Jenkins image S2I compatible to install plugins and configuration

mfojtik:
See: https://gist.github.com/mfojtik/ff4ed568cfea5b8974d2

This will allow:

* Installing plugin binaries using `plugins/` directory in repository
* Installing plugins using `plugins.sh plugins.txt` (downloading from Jenkins update center)
* Installing custom Jenkins job definitions
* Installing additional files/configuration

This should work both on CentOS and RHEL7 image. The main reason for this is that we can't include the plugins in RHEL7 image as we can't download them from internet. This will allow users to rebuild the image and add the plugins from the repository (if they want to).
The image will still work out-of-box (just without plugins, kubernetes slave support and other things).


1/contrib/jenkins/run-jenkins
```diff
@@ -6,24 +6,17 @@
...
-# Generate passwd file based on current uid
-function generate_passwd_file() {
-  export USER_ID=$1
-  export GROUP_ID=$2
-  envsubst < /opt/openshift/passwd.template > /opt/openshift/passwd
-  export LD_PRELOAD=libnss_wrapper.so
-  export NSS_WRAPPER_PASSWD=/opt/openshift/passwd
-  export NSS_WRAPPER_GROUP=/etc/group
-}
+for name in $(get_is_names); do
```
bparees: doesn't this assume you're running the image in openshift?  you ought to be able to docker run the image outside of openshift and have it work.

bparees: meaning both "i can run this image outside openshift but have it connect to an openshift cluster" and "i can run this image outside openshift with no intent to use it with openshift"

bparees:
@smarterclayton @danmcp how do you guys feel about using s2i to layer plugins and job configuration into a jenkins image?  (ie users would construct an s2i build strategy that produces their customized jenkins image)?  personally i like it but i know we've had some debates before over what should be considered "source"

1/contrib/jenkins/run-jenkins
mfojtik: @bparees this should just return empty string in that case and be skipped :-)

mfojtik:
@smarterclayton @danmcp also we mainly doing this to offer a way for customers without external access to install plugins without downloading them from jenkins update center (they can make them part of the GIT repo)

1/contrib/jenkins/run-jenkins
mfojtik: fixed (added `has_service_account` function to detect openshift)

smarterclayton: I'm ok with that.  It fits other patterns we'd have (like Drupal, which
wants to add modules from a git repo).

bparees: Was this reply targeted at the main thread about s2i enabling Jenkins?

mfojtik:
@smarterclayton @danmcp ping :-)

danmcp:
@mfojtik @bparees I think this is a great option.  Storing plugins in git isn't my favorite path and doesn't promote a good maintenance strategy.  But I am fine if it's a last resort.

mfojtik:
@danmcp @bparees OK, I moved everything from my personal repo here, so my repo is not needed anymore. The master is now S2I compatible and the slave converter is also included in `examples/slave`.

I'm going to drop some decent README.md to both examples

mfojtik:
@bparees ready for the first round of reviews :-)

smarterclayton:
This is an excellent idea.

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
+# Jenkins Master example
+
+This directory contains an example of Jenkins setup, that is configured to
```
gabemontero: s/an example of Jenkins setup, that/an  example of a Jenkins setup that/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+# Now create the templates in the 'ci' and 'stage' projects
+$ oc create -f openshift -n ci
+$ oc create -f https://raw.githubusercontent.com/openshift/jenkins/master/examples/master/jenkins-master-with-k8s-plugin.json
+$ oc create -f https://raw.githubusercontent.com/openshift/jenkins/master/examples/slave/s2i-slave-template.json
+```
+
+## Instantiating templates from OpenShift web console
+
+Navigate to the OpenShift UI and choose the `ci` project we created in previous
```
gabemontero: s/created in previous/created in the previous/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+step. Now click on *Add to Project* button and then click on *Show All
```
gabemontero: s/click on/click the/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+Template*. You should see *jenkins-master* template and *s2i-to-jenkins-slave*.
+
+Note: To continue, you have to first create your Jenkins Slave image. Please see
+the [instructions](../slave/README.md) here.
+
+### Jenkins Master
+
+This template defines a
+[BuildConfig](https://docs.openshift.org/latest/dev_guide/builds.html#defining-a-buildconfig)
+that again, use the [Docker
```
gabemontero: s/that again, use/that uses/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+When you choose the `jenkins-master` template, you have to specify these parameters:
+
+* **JENKINS_SERVICE_NAME** - The name of the Jenkins service (default: *jenkins*)
+* **JENKINS_IMAGE** - The name of the original Jenkins image to use
+* **JENKINS_PASSWORD** - The Jenkins 'admin' user password
+
+Once you instantiate this template, you should see a new service *jenkins* in
+the overview page and a route *https://jenkins-ci.router.default.svc.cluster.local/*.
+
+You have to wait till OpenShift rebuilds the original image to include all
```
gabemontero: s/till/until/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+plugins and configuration needed.
+
+### Sample Application
+
+The last step is to instantiate the `sample-app` template. The [sample
+app](sample-app) here is a simple Ruby application
```
gabemontero: s/here is/is/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+that runs Sinatra and have one unit test defined to exercise the CI flow.
```
gabemontero: s/have one/has one/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+You have to instantiate the template in both `ci` and `stage` projects.
+
+## Workflow
+
+You can see [watch the youtube](https://www.youtube.com/watch?v=HsdmSaz1zhs)
```
gabemontero: s/You can see/You can/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+video that shows the full workflow. What happens in the video is:
+
+1. When the `sample-app-test` job is started it fetches the [sample-app](sample-app) sources,
+   install all required rubygems using bundler and then execute the sample unit tests.
```
gabemontero: s/install/installs/
s/execute/executes/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   In the job definition, we restricted this job to run only on slaves that has
```
gabemontero: s/has/have the/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   *ruby-20-centos7* label. This will match the Kubernetes Pod Template you seen
```
gabemontero: s/seen/see/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   in the Kubernetes plugin configuration. Once this job is started and queued,
+   the plugin connects to OpenShift and start the slave Pod using the converted
```
gabemontero: s/start/starts/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   S2I image. The job then run entirely on that slave.
```
gabemontero: s/run/runs/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   When this job finishes, the Pod is automatically destroyed by the Kubernetes
+   plugin.
+
+2. If the unit tests passed, the `sample-app-build` is started automatically via
+   the Jenkins [promoted builds](https://wiki.jenkins-ci.org/display/JENKINS/Promoted+Builds+Plugin)
+   plugin. This job will leverage the OpenShift Jenkins plugin that will start a
```
gabemontero: s/that will start/and start/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   build of the Docker image which will contain 'sample-app'.
+
+3. Once the new Docker image for the `sample-app` is built, the
+   `sample-app-stage` project will automatically deploy it into `stage` project
+   and notify the QA team about availability for testing.
+
+3. If the `sample-app` image pass the `stage` testing, you have to **manually
```
gabemontero: s/pass/passes/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+   promote** the `sample-app-build` build to be deployed to OpenShift. Since
+   re-deploying the application replaces the existing running application, human
+   intervention is needed confirm this step.
```
gabemontero: s/is needed confirm this step/is deemed necessary/

gabemontero:
@mfojtik - aside from the minor comments with the documentation, I'll officially add my +1 to this.

I think it should also be noted that aside from the cool s2i idea, these change from @mfojtik  are pulling in the use of the k8s cloudbees plugin and the master/slave scenario previously demonstrated (including call outs to the new v3 openshift jenkins plugin I've been working on).  Also cool I would claim.

And a heads up (or reminder), once the v3 openshift plugin is officially in the jenkins image, we'll be converting the existing sample job (currently untouched by @mfojtik 's changes) from its current shell script based job to a job solely composed of interactions with the v3 openshift plugin.

1/Dockerfile
```diff
@@ -14,12 +14,14 @@ MAINTAINER Ben Parees <bparees@redhat.com>
 
 ENV JENKINS_VERSION=1.609 \
     HOME=/var/lib/jenkins \
-    JENKINS_HOME=/var/lib/jenkins
+    JENKINS_HOME=/var/lib/jenkins \
+    STI_SCRIPTS_URL=image:///usr/libexec/s2i
```
bparees: i think we can safely remove this at this point, what do you think?  we deprecated it in s2i quite a while ago at this point.

1/contrib/s2i/assemble
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Replace the sample jenkins job if user has "jobs/" directory in repository
+if [ -d ${JENKINS_INSTALL_DIR}/jobs ]; then
+  echo "---> Removing sample Jenkins job ..."
+  rm -rf ${JENKINS_DIR}/configuration/jobs
+  echo "---> Installing $(ls -l ${JENKINS_INSTALL_DIR}/jobs | grep ^d | wc -l) Jenkins jobs ..."
+  mv ${JENKINS_INSTALL_DIR}/jobs ${JENKINS_DIR}/configuration/
+  chmod -R go+rw ${JENKINS_DIR}/configuration/jobs
+fi
```
bparees: should let the source repo supply a full jenkins config directory structure (so they can also supply a config.xml for example).

1/contrib/s2i/run
```diff
@@ -69,6 +60,12 @@ fi
 
 rm -rf /tmp/war
 
+echo "Processing Jenkins Kubernetes configuration (${CONFIG_PATH}) ..."
+envsubst < "${CONFIG_PATH}.tpl" > "${CONFIG_PATH}" && rm -f "${CONFIG_PATH}.tpl"
```
bparees: need to balance this against the ability for users to supply their own configuration file that should not be overwritten by this logic.  maybe only do this if a config.xml doesn't exist?  except i'm assuming you want to do this always so you can pick up env changes.

1/test/run
```diff
@@ -22,13 +22,8 @@ function cleanup() {
...
     exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
-    # jenkins "normally" exits with code 143:
-    # http://jenkins-ci.361315.n4.nabble.com/Jenkins-exit-code-143-due-to-Unexpected-termination-of-the-channel-td4652347.html
-    # tl;dr: 128+signal=return code, and "15" equates to SIGTERM
-    if [ "$exit_status" != "143" ]; then
-      echo "Dumping logs for $CONTAINER"
-      docker logs $CONTAINER
-    fi
+    echo "Dumping logs for $CONTAINER"
+    [[ "${exit_status}" != "0" ]] && docker logs $CONTAINER
```
bparees: did you make a change to how jenkins sets its exit code?

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
+# Jenkins Master example
+
+This directory contains an example of Jenkins setup, that is configured to
+demonstrate the CI/pipeline workflow for the sample-app application using the
+Jenkins Master/Slave setup and automatization done on OpenShift v3.
```
bparees: s/automization/automation/

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+$ oc new-project ci
+$ oc policy add-role-to-user edit system:serviceaccount:ci:default
+
+# Create the 'staging' project where we deploy the sample-app for testing
+$ oc new-project stage
+$ oc policy add-role-to-user edit system:serviceaccount:stage:default
+$ oc policy add-role-to-user edit system:serviceaccount:ci:default
+
+# Now create the templates in the 'ci' and 'stage' projects
+$ oc create -f openshift -n ci
```
bparees: what's "-f openshift" here?

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+$ oc create -f https://raw.githubusercontent.com/openshift/jenkins/master/examples/master/jenkins-master-with-k8s-plugin.json
```
bparees: why isn't jenkins created in the CI project?

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+$ oc create -f https://raw.githubusercontent.com/openshift/jenkins/master/examples/slave/s2i-slave-template.json
```
bparees: I would also expect this to be created in the ci project.

examples/master/README.md
```diff
@@ -0,0 +1,102 @@
...
+Strategy](https://docs.openshift.org/latest/dev_guide/builds.html#docker-strategy-options)
+to rebuild the official [OpenShift Jenkins Image](https://github.com/openshift/jenkins).
+This template also defines a [Deployment Configuration](https://docs.openshift.org/latest/dev_guide/deployments.html#creating-a-deployment-configuration) that will start just one instance
+of the Jenkins server.
+
+When you choose the `jenkins-master` template, you have to specify these parameters:
+
+* **JENKINS_SERVICE_NAME** - The name of the Jenkins service (default: *jenkins*)
+* **JENKINS_IMAGE** - The name of the original Jenkins image to use
+* **JENKINS_PASSWORD** - The Jenkins 'admin' user password
```
bparees: shouldn't the source uri for the jenkins plugin config be a parameter too?

examples/slave/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+* [run-jnlp-client](https://github.com/mfojtik/jenkins-ci/blob/master/jenkins-slave/contrib/openshift/run-jnlp-client)
+* [run-swarm-client](https://github.com/mfojtik/jenkins-ci/blob/master/jenkins-slave/contrib/openshift/run-swarm-client)
+
+Once you have this Docker Image, you have to manually configure Jenkins Master
+to use these images as a slaves. Follow the steps in the
+[jenkins-kubernetes-plugin](https://github.com/jenkinsci/kubernetes-plugin#running-in-kubernetes-google-container-engine)
+documentation.
+
+#### Tagging exiting ImageStream as Jenkins Slave
```
bparees: existing

examples/slave/README.md
```diff
@@ -0,0 +1,76 @@
...
+      "slave-label": "my-slave",
+      "slave-directory": "/opt/app-root/jenkins"
+    }
+  },
+  "spec": {}
+}
+```
+
+The `role=jenkins-slave` label is mandatory, but the annotations are optional.
+If the `slave-label` annotations is not set, Jenkins use the ImageStream name as
```
bparees: s/Jenkins use/Jenkins will use/

bparees:
i'm worried about the number of binary files being checked in here...  do we really need to support swarm and remoting?  what are the trade offs for one vs the other?

also how close are we to having the openshift plugin hpi installable as a dependent plugin instead of embedding the hpi in this repo?

1/Dockerfile.rhel7
```diff
@@ -42,10 +44,10 @@ RUN yum install -y git tar zip unzip java-1.7.0-openjdk jenkins-1.609.1 && \
 
 COPY ./contrib/openshift /opt/openshift
 COPY ./contrib/jenkins /usr/local/bin
+ADD ./contrib/s2i /usr/libexec/s2i
 
 RUN chmod -R og+rw {/opt/openshift,/var/lib/jenkins}
```
bparees: we should switch this to chgrp root since that's our pattern for enabling access now.

bparees:
@mfojtik rather than having a "slave builder template" how about including a jenkins job definition in the image that when run, takes a parameter (name of the image to slave-ify) and talks to openshift to build the slave/define the imagestream?  That makes jenkins one step closer to being self sufficient (it can construct its own slave images.  or at least request their construction).

1/Dockerfile
mfojtik: @bparees +1 but I would keep this in RHEL7 Dockerfile for backward compatibility with OSE 3.0.1

1/contrib/s2i/run
mfojtik: this is tricky... if you supply your own `config.xml` then you will disable the support for k8s slave detection, right? I think the ideal solution here would be support both, `config.xml` and `config.xml.tpl`. If the later exists, we also do processing using the environment variables. And since users can define their own variables, they can also use this mechanism to parametrize their own config template.

1/test/run
mfojtik: no, but I think even if this shows a lot of useless logs in stdout, it is very valuable to see the errors that might not be caused by jenkins itself, but the scripts/etc. For some reason, I'm not able to see any logs with the if set to != "143" ;-)

mfojtik:
@bparees I like that idea. what we need here is @gabemontero plugin to process the template with parameters set (image name) and then create the result. The template will be stored in the Jenkins image, so the plugin should be able to read it from disk location.

examples/master/README.md
mfojtik: create everything in "openshift" directory... left over from copy&paste

examples/master/README.md
mfojtik: @bparees I also need a sample-app with some fake unit test... any tip what repo to use? ;-) (and I will have to add example unit test there)

examples/master/README.md
mfojtik: @bparees what you mean with jenkins plugin config? The file that list the plugins that we install?

mfojtik:
@bparees the swarm vs. remoting ... we can download the remoting.jar from the Jenkins master (if it is running). We can do that during the build phase, but that means the Jenkins master must be up when the build is executed.

The swarm.jar is the same story... If the Jenkins master is up and running and has swarm plugin configured then you should be able to fetch the client from Jenkins master.

Maybe a good solution can be to fetch the slave jar (swarm/remoting) when we start the slave Pod (a shell script that will call back to Jenkins, fetch the appropriate client jar and run it).

1/test/run
bparees: checking for exit status 0 is misleading and pointless though, because jenkins never exits with RC=0.

so either remove the check entirely, or put it back to 143 because "143" is the successful exit code for jenkins when it was sigterm'd, per the comment i had in there.

examples/master/README.md
bparees: the repo containing the file listing the plugins and the actual plugin binaries (in the case that they are providing the plugin files directly)

bparees: put another way, the source repo for the buildconfig.

1/contrib/s2i/run
bparees: +1

examples/master/README.md
bparees: ruby-hello-world?  challenge there is it's a rails app so you need a DB.

bparees:
yes fetching from the master sounds like a better solution.

gabemontero:
So trying to formulate the new requirement on the plug-in per the exchange between @mfojtik and @bparees on "salve builder template" into a step-by-step recipe I can run with.  My initial attempt:
1. an openshift template json file will be included at a well known location in the openshift jenkins image
2. all the parameters in that template will have a default set
3. we add a new jenkins build step to the plug-in whose jenkins UI panel has analogous parameters to the parameters in the template that a customer could want to override (image name for example)
4. the new build step reads in the json from the file, updating the parameters with any values specified by the user, and does an `oc process -f <json> | oc create -`

@mfojtik @bparees : please comment / evaluate

bparees:
(3) input image/imagestream name, output imagestream name at a minimum.  I think it might be best to require the input be an imagestream (rather than an image).  But then we either need to require they create an imagestream, or the template should create an imagestream (and the user would go back to providing the image name).  The point being that by using an imagestream as the input we can more easily get the slave image rebuilt when the image it's based on changes.

1/test/run
mfojtik: @bparees hrm, I would rather remove the check and display all container output (it really helps debugging).

On the other hand, can we fix Jenkins to return 0 when sigtermed? I guess it won't be showing well in the console or in CloudForms when the container exit will not be graceful. However, I don't know how to do it 'smart' way with exec.

examples/master/README.md
mfojtik: I see, to parametrize the build. Gotcha and yes, that is very good idea. Do you also want to parametrize the 'git ref' ? It might help testing ;-)

mfojtik:
@gabemontero 

1) Correct
2) Correct, but the only really important parameter will be IMAGE_NAME. It really is an 'imagestream' name (not an image name). We can optionally add IMAGE_NAME and IMAGE_STREAM_NAME to allow specifying the full Docker pull spec as an image. This is not possible atm. because i'm using the IMAGE_NAME as the 'name' field (it can't contain `/` and other characters).
3) What I imagine this will do is following:
  * Create a parametrized Jenkins job named "add-jenkins-slave" (naming TBD)
  * The job will take 2 parameters: IMAGE_NAME and IMAGE_STREAM_NAME
  * When you start the job and fill out that two parameters, the job will process the template stored in the Jenkins image (well known location), using that two parameters.
  * The job will create the resulting list of resources in OpenShift
  * It will report the progress of build in the job console (build-logs)
  * Once it is done, it will automatically update the Jenkins configuration and add the new slave (or replace existing one if it already exists)

1/test/run
bparees: just remove the check i guess.

i'm not sure how to easily report an rc 0 either, I think you'd need to do some wrappering of the jenkins process and signal handling/passing rather than directly exec'ing jenkins.

bparees: but leave a comment about why you removed the check so someone doesn't come back later and try to naively add a check for rc=0.

examples/master/README.md
bparees: sure, as long as it has a sane default.

examples/master/README.md
```diff
@@ -0,0 +1,101 @@
...
+# Create project and allow Jenkins to talk to OpenShift API
+$ oc new-project ci
+$ oc policy add-role-to-user edit system:serviceaccount:ci:default
+
+# Create the 'staging' project where we deploy the sample-app for testing
+$ oc new-project stage
+$ oc policy add-role-to-user edit system:serviceaccount:stage:default
+$ oc policy add-role-to-user edit system:serviceaccount:ci:default
+
+# Now create the templates in the 'ci' and 'stage' projects
```
bparees: in the 'ci' project.  you're not creating anything in the stage project here.

examples/master/README.md
mfojtik: done.

mfojtik:
@bparees OK, I rebased this, fixed all permissions issues with copying configuration and jobs, fixed the slave template and removed all bundled jars (we now download the remoting.jar from jenkins master), also get rid of swarm (it is relatively easy to build swarm enable jenkins slave anyway) and update the docs....

This is ready for merge imho. Note that the current functionality should not be affected by this. If you don't want to use Jenkins image as a S2I builder, it will still work fine as standalone jenkins server.

1/contrib/s2i/assemble
```diff
@@ -0,0 +1,31 @@
...
+
+echo "---> Copying repository files ..."
+cp -Rf /tmp/src/. ${JENKINS_INSTALL_DIR}
+
+# Install jenkins plugins using plugin installer if user has "plugins.txt" file
+# in repository
+if [ -f ${JENKINS_INSTALL_DIR}/plugins.txt ]; then
+  echo "---> Installing Jenkins $(cat ${JENKINS_INSTALL_DIR}/plugins.txt|wc -l) plugins using ${JENKINS_DIR}/plugins.txt ..."
+  /usr/local/bin/plugins.sh ${JENKINS_INSTALL_DIR}/plugins.txt
+  chmod -R go+rw ${JENKINS_DIR}/plugins
```
bparees: just group and chgrp to root?

1/contrib/s2i/assemble
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+if [ -d ${JENKINS_INSTALL_DIR}/plugins ]; then
+  echo "---> Installing $(ls -l ${JENKINS_INSTALL_DIR}/plugins | grep ^- | wc -l) Jenkins plugins from plugins/ directory ..."
+  cp -R ${JENKINS_INSTALL_DIR}/plugins/* ${JENKINS_DIR}/plugins/
+  chmod -R go+rw ${JENKINS_DIR}/plugins
```
bparees: just group and chgrp to root?

1/contrib/s2i/assemble
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+if [ -d ${JENKINS_INSTALL_DIR}/configuration ]; then
+  if [ -d ${JENKINS_INSTALL_DIR}/configuration/jobs ]; then
+    echo "---> Removing sample Jenkins job ..."
+    rm -rf ${JENKINS_DIR}/configuration/jobs
+  fi
+  echo "---> Installing new Jenkins configuration ..."
+  chmod -R go+rw ${JENKINS_INSTALL_DIR}/configuration
```
bparees: just group and chgrp to root?

bparees: better still use this:
https://github.com/openshift/sti-base/blob/master/bin/fix-permissions

(you'll have to make a copy since this image isn't built on sti-base.  but maybe it should be now that it's an s2i builder?)

1/contrib/s2i/assemble
mfojtik: i tried that but it does not work for some reason :/

examples/master/README.md
bparees: shouldn't this doc be updated then?

bparees:
i think those are my only concerns.

1/contrib/s2i/assemble
mfojtik: non-priv users can't chgrp to root

examples/master/README.md
mfojtik: they are now :-) sorry

1/contrib/s2i/assemble
```diff
@@ -0,0 +1,32 @@
...
+
+echo "---> Copying repository files ..."
+cp -Rf /tmp/src/. ${JENKINS_INSTALL_DIR}
+
+# Install jenkins plugins using plugin installer if user has "plugins.txt" file
+# in repository
+if [ -f ${JENKINS_INSTALL_DIR}/plugins.txt ]; then
+  echo "---> Installing Jenkins $(cat ${JENKINS_INSTALL_DIR}/plugins.txt|wc -l) plugins using ${JENKINS_DIR}/plugins.txt ..."
+  /usr/local/bin/plugins.sh ${JENKINS_INSTALL_DIR}/plugins.txt
+  chmod -R g+rw ${JENKINS_DIR}/plugins
```
bparees: still no fix-permissions script?  i think the directories need to be marked executable if only because @csrwng found that necessary for the rsync logic to work.

though i guess fixing these permissions doesn't matter that much since the files end up copied into the persistent volume...but stil, presumably it would carry the permissions over.

mfojtik:
@bparees fix-permissions added, ship it :-)

bparees:
@mfojtik squash it.

bparees:
@mfojtik given the scope of the change i'd like to know that this scenario still works for you:
https://github.com/openshift/origin/blob/master/examples/jenkins/README.md

mfojtik:
@bparees just tested it, yes it does (new-app works, jenkins get started, no s2i build is needed).

aweiteka:
Nice! This is a cool pattern: optional S2I, otherwise "just works".

mfojtik:
As a follow up we should really add some extended tests for Jenkins that will:

* exercise the s2i build flow with installing some plugins
* exercise the kubernetes plugin slave auto registration
* exercise the jenkins slave stuff (fake job that builds on slave pod) (this one will be hard ;-)

mfojtik:
@bparees (do not merge this yet, need to test the last change)

mfojtik:
@bparees OK, ready for merge again... Changes:

* No more warning messages in the configuration page when running just standalone Jenkins
* Credentials are generated only when kubernetes slaves are configured

@bparees this is now working without any warnings or errors during build or during deployment/runtime.

PR #37, opened 2015-10-09
fix logic to find latest deployment

bparees:


bparees:
@mfojtik ptal
[test]

mfojtik:
lgtm

PR #43, opened 2015-11-09
Fix envsubst for credentials.xml in Jenkins

mfojtik:


mfojtik:
@bparees merge pls, I forgot to add this commit to the s2i branch...

PR #44, opened 2015-11-09
Switch the jenkins master to NORMAL mode

mfojtik:


mfojtik:
@bparees another fix (follow-up). Without this you will have to configure the jobs to run on master explicitly which breaks the "standalone" model. When you have slaves, then you will have to configure the jobs to run on slaves manually (or the job generator will have to set them as exclusive).

bparees:
i'm going to remember this next time you harass me to merge one of your PRs right before GA.

mfojtik:
@bparees i'm sorry :-(

PR #46, opened 2015-11-18
Pre-install kubernetes and openshift plugins in Jenkins image

mfojtik:
Do not merge this PR until we can see in Jenkins that the two packages are installed during the build.

For RHEL7 this will pull the two RPM's @tdawson created for us.
For Centos7 we just use the `plugins.sh` mechanism to install the plugins during the build.

We should keep the plugin versions in sync between rhel7 and centos7. (they will start to diverge at some point, so we need to add some tests to exercise them for both base images)

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -42,7 +42,8 @@ COPY ./contrib/openshift /opt/openshift
 COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 
-RUN /usr/local/bin/fix-permissions /opt/openshift && \
+RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
+    /usr/local/bin/fix-permissions /opt/openshift && \
     /usr/local/bin/fix-permissions /var/lib/jenkins
 
 VOLUME ["/var/lib/jenkins"]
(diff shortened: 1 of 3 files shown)


mfojtik:
[test]

mfojtik:
@gabemontero FYI (just want to see if they install properly)

The next step is to figure out how we install these in centos7. We could probably use plugins.txt because we can fetch plugins from the upstream.

gabemontero:
yep ... i plan to circle back to this soon (have 1 outstanding bug I want to wrap up first).

also note, as part of getting the openshift plugin hosted on the jenkins udpate center, I had to rename it.  So the sample jobs you added to the master/slave examples need some tweaking as well.

I've got a branch with those renames.

mfojtik:
[test]

bparees:
[test]

1/Dockerfile
```diff
@@ -42,7 +42,8 @@ COPY ./contrib/openshift /opt/openshift
 COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 
-RUN /usr/local/bin/fix-permissions /opt/openshift && \
+RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
```
bparees: why can't centos also install the plugins via rpm like rhel is?

mfojtik:
Closing this as this PR is outdated.

PR #55, opened 2015-12-09
Remove kubernetes plugin support

mfojtik:
This PR removes the included Kubernetes plugin (can be installed by custom assemble script) and the auto-slave-discovery code (can be done by the custom run script).

@bparees do you want to remove the examples/ as well? I think we should move them into a example repo with example S2I scripts.


mfojtik:
@gabemontero FYI

mfojtik:
[test]

bparees:
yeah we should remove those as well or we're just going to get questions about them.

mfojtik:
@bparees done

1/contrib/jenkins/jenkins-common.sh
```diff
@@ -1,48 +1,17 @@
...
-  </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
-  "
-  echo "${template}" > ${template_file}
-  $oc_cmd get is/${name} -o templatefile -t ${template_file}
-  rm -f ${template_file} &>/dev/null
-}
+if [ -f "${AUTH_TOKEN}" ]; then
+  export oc_auth="--token=$(cat $AUTH_TOKEN) --certificate-authority=${KUBE_CA}"
+  export oc_cmd="oc --server=$OPENSHIFT_API_URL ${oc_auth}"
+fi
```
bparees: is this needed for the openshift pipeline plugin?  or just a convenience?

gabemontero: Just convenience ... the pipeline plugin can find the token and cert on the files system if not passed in via param

bparees: so what is actually consuming these env variables?

gabemontero: let me clarify ... the pipeline plugin *CAN* consume AUTH_TOKEN, you are just not required / don't need to set AUTH_TOKEN in order for the pipeline plugin to work.

bparees: this isn't setting AUTH_TOKEN though, it's setting $oc_auth and $oc_cmd.  does the plugin use those?

gabemontero: Though AUTH_TOKEN must be previously set since the if check needs to pass to set oc_auth and oc_cmd ... I simply chose to look for AUTH_TOKEN as a convenience based on legacy use of that precise string; the plugin does not need/leverage oc_auth / oc_cmd out of the box.

mfojtik: @bparees just convenience. I thought it might be useful and you can consume those in job definitions.

bparees: got it, thanks @mfojtik

bparees:
lgtm i think.

mfojtik:
merge then I think :-)

PR #56, opened 2015-12-09
fixed typo

codenirvana:
Looks like ```mounting ``` is not spelled correctly.


mfojtik:
Thanks for spotting this @codenirvana !

PR #57, opened 2015-12-09
Initial addition of the S2I tests

mfojtik:


mfojtik:
[test]

@bparees @gabemontero PTAL

1/test/run
```diff
@@ -137,7 +137,40 @@ function run_tests() {
...
+  if [ "$(which s2i)" == "" ]; then
+    echo "WARNING: Please install the s2i binary to run the s2i tests."
+    return
+  fi
+  output_image="openshift/s2i-test-jenkins"
+  s2i build --context-dir="1/test/s2i" --loglevel=3 --pull-policy=never \
+    file://$(pwd)/.. $IMAGE_NAME $output_image
+  IMAGE_NAME="${output_image}" create_container s2i-jenkins-test $VOLUME
+  cid=s2i-jenkins-test
+  sleep 3 # Give jenkins time to copy all plugins/configs
```
bparees: retry logic with a duration would be better.

mfojtik: it can flake then... the first case can succeed but it will fail for the second one (the files are not yet copied)... unless you want to retry the whole test :-) i don't think 3 seconds are all that horrible.

bparees:
one nit and lgtm.

though one other thought would be to actually start the image and confirm it's accessible (as the existing test does) and that the test-job exists.  that would

1) further confirm the config was copied correctly 
2) confirm the built image still works (ie we aren't messing up the jenkins configuration via the assemble script, or set bad permissions, or.....)

mfojtik:
@bparees your comments addressed

1/test/run
```diff
@@ -137,7 +137,46 @@ function run_tests() {
...
+  test_get_job admin password sample-app-test
+  test_create_job admin password testJob
+  test_get_job admin password testJob
+}
+
 # Tests.
 
 # Normal tests
-run_tests jenkins_test
+# run_tests jenkins_test
```
bparees: re-enable this?

1/test/run
```diff
@@ -137,7 +137,46 @@ function run_tests() {
...
+    return 1
+  fi
+  docker exec ${cid} cat /var/lib/jenkins/config.xml | grep -q "s2i-test-config"
+  if [ "$?" != "0" ]; then
+    echo "Fail: The config.xml was not copied"
+    return 1
+  fi
+  set -e
+  test_get_job admin password sample-app-test
+  test_create_job admin password testJob
```
bparees: it's confusing that the job your config files provide is the same name as the job the other test creates/gets.  can you rename one of them?

bparees: oh, nm.  i misread this.

gabemontero:
fyi ... have looked, no comments (though looking was educational for me :-) )

mfojtik:
@bparees re-enabled original tests... leftover from debugging :-)

mfojtik:
@bparees and now I'm fighting with permissions again... https://gist.github.com/mfojtik/09251cf483bc712ccc07

Note that this happen on my local vagrant centos7 machine, but I can't see these errors in CI. My assumption is that all files in the image are owned by "root:root" (or at least from what I can see when I docker exec). When I start the S2I build however, it runs as "1001" user (defined in Dockerfile). I can't see how the CI build passes without errors as the "1001" user can't change ownership of the files owned by "root".

mfojtik:
@bparees the last commit will chown all files in /opt/openshift to 1001 (which I hope will enable us to run s2i build command)

bparees:
lgtm.

PR #59, opened 2015-12-10
fix rhel7 image to jdk1.8, link plugin rpm in jenkins startup path

gabemontero:
@bparees @mfojtik fyi / ptal

### DIFF
--- a/1/Dockerfile.rhel7
+++ b/1/Dockerfile.rhel7
@@ -33,7 +33,7 @@ LABEL BZComponent="openshift-jenkins-docker" \
 # 8080 for main web interface, 50000 for slave agents
 EXPOSE 8080 50000
 
-RUN yum install -y gettext git tar zip unzip java-1.7.0-openjdk jenkins-1.609.1 jenkins-plugin-openshift-pipeline && \
+RUN yum install -y gettext git tar zip unzip java-1.8.0-openjdk jenkins-1.609.1 jenkins-plugin-openshift-pipeline && \
     yum install -y --disablerepo="epel" --setopt=tsflags=nodocs nss_wrapper && \
     yum clean all  && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

1/Dockerfile.rhel7
```diff
@@ -46,7 +46,9 @@ COPY ./contrib/openshift /opt/openshift
 COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 
-RUN /usr/local/bin/fix-permissions /opt/openshift && \
+RUN mkdir -p /tmp/openshift/plugins && \
+    ln -s /usr/lib64/jenkins/openshift-pipeline.hpi /opt/openshift/plugins/openshift-pipeline.hpi && \
```
bparees: test doesn't seem to like this.  does the plugins directory exist at the point at which you are running this command in the docker build?

gabemontero: Well ... i just noticed that I had a typo ... /tmp/openshift/plugins vs. /opt/openshift/plugins ...the plugins subdir did not exist in /opt/openshift when I exec/ran the rhel7 container earlier, and on the centos side, the plugins.sh script that is runn is doing the mkdir -p /opt/openshift/plugins.

I'll try fixing the typo first, and then if it still complains, i'll move the `mkdir` and `ln` into a simple script analogous to the one ran in the centos image.

gabemontero:
[test]

gabemontero:
at least from a build perspective, switching /tmp/openshift/plugins to /opt/openshift/plugins seemed to do the trick:  https://ci.openshift.redhat.com/jenkins/job/jenkins/148/console

gabemontero:
I think the tests actually ran (https://ci.openshift.redhat.com/jenkins/job/jenkins/149/), but I will re [test] to be sure.

gabemontero:
@bparees - i think we are gtm unless i'm missing something.  Also, just confirming, assuming it merges, will the push_images job be kicked off automatically, or do we need to manually hit that button?

thx

bparees:
push_images runs every hour looking for repos that have commits that are newer than the latest built images for that repo.

so this should get built/pushed automatically the next time that job runs.

gabemontero:
Cool thx

PR #62, opened 2015-12-18
dummy change

bparees:


bparees:
@gabemontero just making this change to force a re-spin of the jenkins image so the rhel images will pick up the new plugin package.  (for reasons i won't go into, this is easier than the alternative)

gabemontero:
Roger that @bparees - thx

PR #74, opened 2016-02-02
remove problematic deploy verify step from sample job

gabemontero:
@bparees PTAL - related to latest discussion in bugzilla 1293859 (new defect most likely coming); also, I believe you and I discussed removing this step before Christmas, but I had put it lower on the to-do list, not realizing the ramifications until seeing the results after 1293859 was resolved.


bparees:
lgtm

PR #81, opened 2016-02-15
tune java heap, fix sample job config, install rsync

gabemontero:
@bparees PTAL

@abhgupta @danmcp fyi - this PR among other things will add auto-tuning of the java heap based on available memory to the openshift jenkins image

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -28,7 +28,7 @@ EXPOSE 8080 50000
 RUN curl http://pkg.jenkins-ci.org/redhat/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import http://pkg.jenkins-ci.org/redhat/jenkins-ci.org.key && \
     yum -y --setopt=tsflags=nodocs install epel-release && \
-    INSTALL_PKGS="gettext git tar zip unzip java-1.8.0-openjdk jenkins-1.642 nss_wrapper" && \
+    INSTALL_PKGS="rsync gettext git tar zip unzip java-1.8.0-openjdk jenkins-1.642 nss_wrapper" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all  && \
--- a/1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
+++ b/1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
@@ -13,29 +13,61 @@
   <builders>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftScaler>
-       <replicaCount>0</replicaCount>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <depCfg>frontend</depCfg>
+      <namespace>test</namespace>
+      <replicaCount>0</replicaCount>
+      <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftScaler>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftBuilder>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <bldCfg>frontend</bldCfg>
+      <namespace>test</namespace>
+      <authToken></authToken>
       <checkForTriggeredDeployments>true</checkForTriggeredDeployments>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftBuilder>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftDeploymentVerifier>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <depCfg>frontend</depCfg>
+      <namespace>test</namespace>
+      <replicaCount>0</replicaCount>
+      <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftDeploymentVerifier>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftScaler>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <depCfg>frontend</depCfg>
+      <namespace>test</namespace>
       <replicaCount>1</replicaCount>
+      <authToken></authToken>
       <verifyReplicaCount>true</verifyReplicaCount>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftScaler>
     
     <com.openshift.jenkins.plugins.pipeline.OpenShiftServiceVerifier>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <svcName>frontend</svcName>
+      <namespace>test</namespace>
+      <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftServiceVerifier>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftImageTagger>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <namespace>test</namespace>
+      <testTag>latest</testTag>
+      <prodTag>prod</prodTag>
+      <testStream>origin-nodejs-sample</testStream>
+      <prodStream>origin-nodejs-sample</prodStream>
+      <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftImageTagger>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftDeploymentVerifier>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
       <depCfg>frontend-prod</depCfg>
+      <namespace>test</namespace>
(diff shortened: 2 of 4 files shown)


1/Dockerfile
```diff
@@ -28,7 +28,7 @@ EXPOSE 8080 50000
 RUN curl http://pkg.jenkins-ci.org/redhat/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import http://pkg.jenkins-ci.org/redhat/jenkins-ci.org.key && \
     yum -y --setopt=tsflags=nodocs install epel-release && \
-    INSTALL_PKGS="gettext git tar zip unzip java-1.8.0-openjdk jenkins-1.642 nss_wrapper" && \
+    INSTALL_PKGS="rsync tar gettext git tar zip unzip java-1.8.0-openjdk jenkins-1.642 nss_wrapper" && \
```
bparees: tar is listed twice now

1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
```diff
@@ -13,29 +13,61 @@
...
+      <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftServiceVerifier>
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftImageTagger>
+      <apiURL>https://openshift.default.svc.cluster.local</apiURL>
+      <namespace>test</namespace>
+      <testTag>latest</testTag>
+      <prodTag>prod</prodTag>
+      <testStream>origin-nodejs-sample</testStream>
+      <prodStream>origin-nodejs-sample</prodStream>
```
bparees: is this the name used for the tagging feature?  I think something like "source" and "destination" would have been preferred as it is more generic.  People may use this tagging feature for a variety of scenarios.

1/contrib/s2i/run
```diff
@@ -12,6 +12,24 @@ source /usr/local/bin/kube-slave-common.sh
 jenkins_config_dir="/opt/openshift/configuration"
 jenkins_config_path="${jenkins_config_dir}/config.xml"
 
+
+CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+echo "Container memory in bytes ${CONTAINER_MEMORY_IN_BYTES}"
+DEFAULT_MEMORY_CEILING=$((2**30-1))
+echo "Default memory ceiling ${DEFAULT_MEMORY_CEILING}"
```
bparees: i don't think it makes sense to print this.

1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
gabemontero: No, this was just splitting our the images stream and tag name into separate parameters.

Yeah, source and destination would be better.  I would suggest doing that post 3.2, as it would require a rebuild of the plugin, which would delay delivery.

1/Dockerfile
gabemontero: pushed fix

1/contrib/s2i/run
gabemontero: removal of echo's pushed

1/contrib/s2i/run
```diff
@@ -12,6 +12,24 @@ source /usr/local/bin/kube-slave-common.sh
...
+if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
+    if [ -z $CONTAINER_HEAP_PERCENT ]; then
+	CONTAINER_HEAP_PERCENT=0.80
+    fi
+    
+    CONTAINER_MEMORY_IN_MB=$((${CONTAINER_MEMORY_IN_BYTES}/1024**2))
+    CONTAINER_HEAP_MAX=$(echo "${CONTAINER_MEMORY_IN_MB} ${CONTAINER_HEAP_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+    
+    echo "Going to set max heap to ${CONTAINER_HEAP_MAX}"
```
bparees: i don't think this needs to be printed either.  you can always rsh into the pod and run ps if you want to see what the java args were.  If you do want to print it (eg in case it's needed for debugging a non-starting container), make the message something more formal like "Java max heap=$FOO"

1/contrib/s2i/run
```diff
@@ -86,7 +104,7 @@ rm -rf /tmp/war
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
+   exec java $JAVA_MAX_HEAP_PARAM $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
```
bparees: if JAVA_OPTS contains a -Xmx argument, it will take precedence over the JAVA_MAX_HEAP_PARAM value, right?  Just want to make sure users have a way to explicitly override the max heap setting if they want to.

1/contrib/s2i/run
bparees: i'm also not sure where this value is coming from?  if -Xmx isn't set, doesn't java default the heap size to something like 1/4 of what "free" reports?  (and free is going to report the host memory, not the container memory).

1/contrib/s2i/run
gabemontero: there are other way to debug ... just removed print ... pushed

1/contrib/s2i/run
gabemontero: yes that was my thinking

1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
bparees: yes but we'll have to retain support for the "testTag/prodTag/testStream/prodStream" names when the change is made, for backwards compatibility.

bparees: (but yes, it should be post-3.2 at this point)

1/contrib/s2i/run
```diff
@@ -12,6 +12,21 @@ source /usr/local/bin/kube-slave-common.sh
 jenkins_config_dir="/opt/openshift/configuration"
 jenkins_config_path="${jenkins_config_dir}/config.xml"
 
+
+CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+DEFAULT_MEMORY_CEILING=$((2**30-1))
+if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
```
bparees: so isn't this effectively saying you can never set a memory limit higher than 1gig?  (if you set your memory limit to 1.5gigs, then no max heap value gets set)

(where 2**30==1gig if i've done my math correctly)

1/contrib/s2i/run
gabemontero: The logic from xpaas I borrowed this from originally had a higher number (2**60 -1) for the value.  I was afraid of setting heaps that were too large with larger VMs, but I suppose the flip is that "mid range" containers could still consume to much memory.

I'll look into redoing this logic, by always going into the max heap adjustment path, but  capping the max heap we'll set at say 1.5 GiB.  If users want more besides that, then can override via JAVA_OPTS.

Sound OK ?

1/contrib/s2i/run
gabemontero: See other comment ... redoing logic ...

bparees: (moving here so it's not a comment on an outdated diff).

I think we should always respect the memory limit the users chose, so capping at 1.5gigs (and forcing them to use a different mechanism) doesn't seem like the right thing to do to me.  If a user decides to run a jenkins container with a limit of 128gigs, we should assume that means they want ~128gigs of heap (w/ appropriate native memory reservations, as you're doing with the 80%). If they are doing something fancy where they want most of that memory for processes launched by jenkins, they can use the JAVA_OPTS override then.


The only case i'm particularly worried about is when they set no limit and therefore CONTAINER_MEMORY_IN_BYTES will be either "-1" or 9223372036854771712 and obviously using either of those to set a -Xmx value will be problematic (and if the value that represents "no limit" changes, any logic that uses those values to test for no limit will break).  

2**60 seems excessive, but 2**40 (~1 terabyte) is probably a safe upper limit.

gabemontero: OK ... i leave the general flow as is, but set the ceiling at entering at 2 ^ 40

gabemontero:
ok ... change to work off of 2 ^ 40 pushed

bparees:
lgtm

Issue #83, opened 2016-02-18
About the user for uid 1001


thincal:
If configuring the ssh key for jenkins container, it will report `No user exists for uid 1001` when running `ssh-keygen`.

some info FYI:

$ echo $USER
```

```

$ echo $UID
```
1001
```

$ id jenkins
```
uid=997(jenkins) gid=995(jenkins) groups=995(jenkins)
```

$ cat /opt/openshift/passwd
```
...
jenkins:x:1001:0:Jenkins Continuous Integration Server:/var/lib/jenkins:/bin/false
```

The problem should be there is no user created for the uid 1001, but it seems that it is trying to create a jenkins user with the uid 1001, right ? but the result shows jenkins user's uid is 997.

bparees:
how are you running the image?

thincal:
Followed the example https://github.com/openshift/origin/tree/master/examples/jenkins by running:
```
oc new-app jenkins-persistent-template.json
```

bparees:
when you run an image in openshift you are actually a random user (run whoami to see it), so the 1001 is irrelevant.

As for how to solve this problem, we'll have to use nss_wrapper to ensure we can always map the current random uid to jenkins for programs that require the current uid be resolvable in /etc/passwd.  This is something we've done in other images.

thincal:
$ whoami
whoami: cannot find name for user ID 1001

bparees:
fundamentally yes that is the issue.  you must have openshift configured with a RunAsAny policy though, since the image is running as 1001 and not a random id.

regardless, the fix is what i described, we need to use nss_wrapper to map whatever uid the container is running as, to the jenkins user.

thincal:
@bparees when will get the fix ? thanks.

mfojtik:
@thincal i'm working on the fix today.

mfojtik:
@thincal @bparees I think the problem is because of `docker exec`. It starts a shell but does not set the NSS_WRAPPER properly.

@thincal how exactly you invoking `docker exec` ? (I think something like: 

```console
$ docker exec 123456 /usr/libexec/s2i/run /bin/bash
```

should work). But I think the better solution would be to move the nss_wrapper into entrypoint wrapper.

thincal:
@mfojtik yes after running as following
```
$ docker exec 123456 /usr/libexec/s2i/run /bin/bash
``
it is working.

thincal:
[...]

PR #84, opened 2016-02-18
Move nss_wrapper to container entrypoint

mfojtik:
(Should) fix: https://github.com/openshift/jenkins/issues/83

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -51,4 +51,6 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
 VOLUME ["/var/lib/jenkins"]
 
 USER 1001
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 CMD ["/usr/libexec/s2i/run"]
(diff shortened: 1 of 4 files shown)


mfojtik:
[test]

mfojtik:
@bparees my faith here is that the entrypoint will always set the nss_wrapper right for any command that user run with "docker exec" or "oc rsh" or "oc exec". What do you think?

(need to test this in OpenShift)

rhcarvalho:
@mfojtik `oc rsh` replaces the entrypoint, by default with "/bin/bash", or with any other command you pass in the CLI. And `oc rsh` is using the same code path as `oc exec`, which is the one replacing the entrypoint, so no luck there AFAICT.

Last, `docker exec` is also ignoring the entrypoint. The entrypoint is only used as a default for `docker run`.

mfojtik:
@rhcarvalho yeah, you're right. I thought `oc rsh` respects the entrypoint. @bparees so there is no good
solution for this other than executing the entrypoint.sh with /bin/bash as argument...

mfojtik:
OK scratch this PR... this actually breaks the S2I build (assemble will generate passwd first and the run can't access that file anymore...).

PR #88, opened 2016-02-23
Bug 1311024 - Fix the sample job xml to match current example

mfojtik:


1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
```diff
@@ -57,11 +57,13 @@
...
       <prodTag>prod</prodTag>
-      <testStream>origin-nodejs-sample</testStream>
-      <prodStream>origin-nodejs-sample</prodStream>
+      <testStream>origin-ruby-sample</testStream>
+      <prodStream>origin-ruby-sample</prodStream>
       <authToken></authToken>
     </com.openshift.jenkins.plugins.pipeline.OpenShiftImageTagger>
 
+    <!--
+    FIXME: There is no such deployment config created by sample-app.
```
mfojtik: @gabemontero @bparees ok to remove this step? it will always fail because the sample-app in origin does not have that deployment config.

bparees: there isn't?  this job is supposed to be paired w/ this template, right?
https://github.com/openshift/origin/blob/master/examples/jenkins/application-template.json#L63-L67

mfojtik: @bparees oh, overseen that... i will change the PR.

mfojtik:
@bparees fixed... i think the sample-app job should now work with the sample-app source

1/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
```diff
@@ -22,7 +22,7 @@
 
     <com.openshift.jenkins.plugins.pipeline.OpenShiftBuilder>
       <apiURL>https://openshift.default.svc.cluster.local</apiURL>
-      <bldCfg>frontend</bldCfg>
+      <bldCfg>ruby-sample-build</bldCfg>
```
bparees: still wrong :(
https://github.com/openshift/origin/blob/master/examples/jenkins/application-template.json#L197

mfojtik: /me confused.

The Jenkins tutorial say to create the sample app so I think QE went ahead and create that instead of application-template.json. The sample is ruby but this app is nodejs :-) That means all these changes are not needed and we just need to instruct QE to use the application-template.json instead of sample-app template.

bparees: The jenkins tutorial says:
`$ oc new-app application-template.json`

where application-template.json is a file in the same dir as the tutorial readme.

so yeah, if they used the wrong template, their test instructions need to be updated.

PR #91, opened 2016-04-18
bump oc version

bparees:


bparees:
@gabemontero ptal

gabemontero:
The url `https://github.com/openshift/origin/releases/download/v1.1.6/openshift-origin-client-tools-v1.1.6-ef1caba-linux-64bit.tar.gz` looks correct given what is out there today.  And for the RHEL version, either there isn't a RPM for the `oc` binary, or it is not current enough, and so hence we are downloading?

bparees:
> The url https://github.com/openshift/origin/releases/download/v1.1.6/openshift-origin-client-tools-v1.1.6-ef1caba-linux-64bit.tar.gz looks correct given what is out there today.

not sure what you mean?  the binary file names changed from what they were in 1.0.8 which is why the url is different than it was.  (as is what is packaged, hence no longer needing to remove the oadm file)

>  And for the RHEL version, either there isn't a RPM for the oc binary, or it is not current enough, and so hence we are downloading?

I need to check on that.  At one point i think the rpm wasn't available in the context that the jenkins image builds run in, so we were doing it this way... the actual image that gets built in OSBS uses the rpm which is all that *really* matters, but i'll see if i can change this one to use the rpm for consistency.

gabemontero:
On Mon, Apr 18, 2016 at 3:42 PM, Ben Parees <notifications@github.com>
wrote:

> The url
> https://github.com/openshift/origin/releases/download/v1.1.6/openshift-origin-client-tools-v1.1.6-ef1caba-linux-64bit.tar.gz
> looks correct given what is out there today.
>
> not sure what you mean? the binary file names changed from what they were
> in 1.0.8 which is why the url is different than it was. (as is what is
> packaged, hence no longer needing to remove the oadm file)
>
Sorry, I was just trying to convey that there wasn't a typo in the URL or
anything like that, and that you grabbed the latest version.   And yep, I
inferred that the version and subdir structure of the tar changed.

We're good.

bparees:
> Sorry, I was just trying to convey that there wasn't a typo in the URL or anything like that, and that you grabbed the latest version. 

whoops, i read "looks correct" as "looks incorrect".  must be my inferiority complex acting up again.

bparees:
@gabemontero looks like we can in fact use the rpms now, so i've updated to that.  ptal.

gabemontero:
Looks good to me

PR #94, opened 2016-05-11
fix nss passwd update logic

bparees:


bparees:
(original version of this fix was a sad mess :(  )

PR #98, opened 2016-05-12
bump to 1.651.2 lts

bparees:

PR #99, opened 2016-05-17
move to 1.0.11 of openshift pipeline plugin

gabemontero:
@bparees PTAL

This version has 2 bugzilla fixes from QE testing of my 2 plugin cards this sprint, as well as an upstream ask around exiting failed builds more expediently, followed by an NPE fix that @sspeiche found.

### DIFF
--- a/1/contrib/openshift/base-plugins.txt
+++ b/1/contrib/openshift/base-plugins.txt
@@ -1,4 +1,4 @@
-openshift-pipeline:1.0.10
+openshift-pipeline:1.0.11
 durable-task:1.6
 kubernetes:0.5
 credentials:1.24


bparees:
[test]

bparees:
@gabemontero you'll have to rebase so you can pick up this change:
https://github.com/openshift/jenkins/pull/98

and then the test should pass.  but we probably want to wait to merge this until the rhel rpm is published so the rebuild/push picks up everything in one shot.

gabemontero:
@bparees just rebased off of https://github.com/openshift/jenkins/pull/98 's merging into master.

And yes, agreed on merge timing re: rhel ... that's why I pinged Troy while this was still in progress vs. waiting until this merged.

bparees:
sounds like the rpm is created.

PR #107, opened 2016-06-06
goto to v1.0.16 of plugin

gabemontero:
@bparees PTAL

this should be the last of the changes based on QE's testing of the centos image for the plugin DSL card based on their feedback in trello.


bparees:
lgtm

PR #108, opened 2016-06-07
go to v1.0.17 of openshift pipeline plugin

gabemontero:
@bparees PTAL

this version has the fix for the small nit QE reported in trello last night


bparees:
lgtm

PR #109, opened 2016-06-10
more fabric induced plugin upgrades

gabemontero:
@bparees PTAL

Based on analysis documented in https://trello.com/c/r7oLwXJJ/958-3-jenkins-pipeline-plugins-rpm-packaging I've upgraded versions of a couple of plugins to match the dependencies listed in the latest version of the sync plugin.


bparees:
lgtm

Issue #101, opened 2016-05-19
duplicated openshift-pipeline in /var/lib/jenkins/plugin folder


rafabene:
I have dumped my desired Jenkins dependencies in a plugin.txt file.

Following the instructions of the s2i custom-build, I ended with two openshift-pipeline.hpi in /var/lib/jenkins/plugin folder

41 May 19 03:21 openshift-pipeline.hpi -> /usr/lib64/jenkins/openshift-pipeline.hpi
4124260 May 19 03:21 openshift-pipeline.jpi

This also happened with:

- credentials.hpi
- durable-task.hpi
- kubernetes.hpi

bparees:
can you share the exact content of your plugin.txt and the buildconfig you used?

rafabene:
Sure. The resources are here: https://github.com/redhat-helloworld-msa/jenkins

bparees:
@gabemontero you might have some thoughts here since you've done more w/ installing plugins than i have.

rafabene:
It seems that using openshift/jenkins-1-centos7 is fine. The problem seems to be in the latest rhel image.

gabemontero:
A couple of thoughts (and possibly an answer):
- So the difference in centos / rhel behavior I believe makes sense since they seem to live in different spots based on what I'm seeing when I `docker run ... /bin/bash` those images without any changes.

For RHEL:
/usr/lib64/jenkins/openshift-pipeline.hpi - this comes from the RPM install that @tdawson does for us
/opt/openshift/plugins/openshift-pipeline.hpi - this is an `ln` from our RHEL Dockerfile for the Jenkins image  

For Centos:
/opt/openshift/plugins/openshift-pipeline.jpi

- I do know from the work I did for the jenkins extended test on origin that user the "jpi" suffix is the way to override an existing plugin install with a new version.

So I *suspect* the plugins.sh scirpt needs to do some special processing for the openshift-pipeline plugin where it 
- removes the `/opt/openshift/plugins/openshift-pipeline.hpi` link created by the RHEL dockerfile
- I suppose we can leave the RPM installed, but perhaps we should uninstall it ????
- create the `/opt/openshift/plugins/openshift-pipeline.jpi` file to override the existing plugin

bparees:
>    I do know from the work I did for the jenkins extended test on origin that user the "jpi" suffix is the way to override an existing plugin install with a new version.

[...]

PR #110, opened 2016-06-10
address double install with both hpi and jpi file being present

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/101

@bparees PTAL

### DIFF
--- a/1/Dockerfile.rhel7
+++ b/1/Dockerfile.rhel7
@@ -47,13 +47,13 @@ ADD ./contrib/s2i /usr/libexec/s2i
 # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
 # HPI file created by rpm to /opt/openshift/plugins folder.
 RUN mkdir -p /opt/openshift/plugins && \
-    ln -s /usr/lib64/jenkins/openshift-pipeline.hpi /opt/openshift/plugins/openshift-pipeline.hpi && \
-    ln -s /usr/lib64/jenkins/credentials.hpi /opt/openshift/plugins/credentials.hpi && \
+    ln -s /usr/lib64/jenkins/openshift-pipeline.hpi /opt/openshift/plugins/openshift-pipeline.jpi && \
+    ln -s /usr/lib64/jenkins/credentials.hpi /opt/openshift/plugins/credentials.jpi && \
     # Prefer the latest version of the credentials plugin over the bundled
     # version.
-    touch /opt/openshift/plugins/credentials.hpi.pinned && \
-    ln -s /usr/lib64/jenkins/durable-task.hpi /opt/openshift/plugins/durable-task.hpi && \
-    ln -s /usr/lib64/jenkins/kubernetes.hpi /opt/openshift/plugins/kubernetes.hpi && \
+    touch /opt/openshift/plugins/credentials.jpi.pinned && \
+    ln -s /usr/lib64/jenkins/durable-task.hpi /opt/openshift/plugins/durable-task.jpi && \
+    ln -s /usr/lib64/jenkins/kubernetes.hpi /opt/openshift/plugins/kubernetes.jpi && \
     /usr/local/bin/fix-permissions /opt/openshift && \
     chown -R 1001:0 /opt/openshift && \
     /usr/local/bin/fix-permissions /var/lib/jenkins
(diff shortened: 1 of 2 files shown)


bparees:
looks like we need to change the permissions on those files so they can be overwritten when the plugins.txt processing goes to install them again:

```
I0610 17:13:28.843292 04109 sti.go:603] ---> Copying repository files ...
I0610 17:13:28.853050 04109 sti.go:603] ---> Installing Jenkins 3 plugins using /opt/openshift/plugins.txt ...
I0610 17:13:28.991341 04109 sti.go:603] Downloading credentials-1.23 ...
I0610 17:13:29.445514 04109 sti.go:603] Failed to download credentials-1.23:
I0610 17:13:29.446093 04109 sti.go:603] curl: (23) Failed writing body (0 != 2664)
I0610 17:13:29.447064 04109 sti.go:603] Failed to install plugins.
```

gabemontero:
added a fix-permission call for `/opt/openshift/plugins`

gabemontero:
Same error even with the fix-permission call ...

gabemontero:
trying an explicit chmod on the jpi files

1/Dockerfile.rhel7
```diff
@@ -47,14 +47,16 @@ ADD ./contrib/s2i /usr/libexec/s2i
...
     # version.
-    touch /opt/openshift/plugins/credentials.hpi.pinned && \
-    ln -s /usr/lib64/jenkins/durable-task.hpi /opt/openshift/plugins/durable-task.hpi && \
-    ln -s /usr/lib64/jenkins/kubernetes.hpi /opt/openshift/plugins/kubernetes.hpi && \
+    touch /opt/openshift/plugins/credentials.jpi.pinned && \
+    ln -s /usr/lib64/jenkins/durable-task.hpi /opt/openshift/plugins/durable-task.jpi && \
+    ln -s /usr/lib64/jenkins/kubernetes.hpi /opt/openshift/plugins/kubernetes.jpi && \
     /usr/local/bin/fix-permissions /opt/openshift && \
+    /usr/local/bin/fix-permissions /opt/openshift/plugins && \
+    chmod -R g+rw /opt/openshift/plugins/*.jpi && \
```
bparees: that this appears to have worked means something is going wrong w/ the fix-permissions script..

in fact given that we're making the call on /opt/openshift, adding another call on /opt/openshift/plugins also shouldn't be necessary.

I wonder if chmod/chgrp -R doesn't follow symlinks or something.

that appears to be the case:
```
zahir:/tmp$ ln -s /tmp/b a/blink
zahir:/tmp$ ls -l a
total 0
lrwxrwxrwx. 1 bmp bmp 6 Jun 12 23:24 blink -> /tmp/b

zahir:/tmp$ ls -l b
-rw-rw-r--. 1 bmp bmp 0 Jun 12 23:24 b

zahir:/tmp$ chmod -R a+rwx /tmp/a
zahir:/tmp$ ls -l b
-rw-rw-r--. 1 bmp bmp 0 Jun 12 23:24 b

zahir:/tmp$ cd /tmp/a
zahir:/tmp/a$ chmod a+rwx *
zahir:/tmp/a$ ls -l /tmp/b
-rwxrwxrwx. 1 bmp bmp 0 Jun 12 23:24 /tmp/b
```

http://ubuntuforums.org/showthread.php?t=2046341

we may need to do this entirely via a "find" operation.

gabemontero: yep, good catch on the link thing ... and yep, I got a find / exec chmod to work using your sample above.

1/contrib/jenkins/fix-permissions
```diff
@@ -2,5 +2,6 @@
 # Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
 chgrp -R 0 $1
-chmod -R g+rw $1
+#chmod -R g+rw $1
+find $1 -exec chmod g+rw {} \;
```
bparees: guessing chgrp will need the same treatment?

gabemontero: yeah probably ... i'll wait until the upcoming test completes successfully for the find / chmod before officially pulling that trigger

gabemontero:
OK, all use of `-R` replaced with `find -exec`

bparees:
looks good to me, can you follow up w/ this same change to the other copies o fix-permissions?  (one in s2i-base, and possibly one in each of the DB images?)

gabemontero:
On Mon, Jun 13, 2016 at 9:22 AM, Ben Parees <notifications@github.com>
wrote:

> looks good to me, can you follow up w/ this same change to the other
> copies o fix-permissions? (one in s2i-base, and possibly one in each of the
> DB images?)
>
sure I'll take a look

PR #113, opened 2016-06-18
go to v1.0.19 of plugin

gabemontero:
@bparees PTAL

This is for the DSL `openShift..` to `openshift...` camel case related method name change requested by @smarterclayton


bparees:
lgtm

PR #115, opened 2016-06-22
need to pin ssh-credentials, script-security since they are bundled w…

gabemontero:
…ith jenkins core

@bparees PTAL - noticed as part of testing rhel image and fabric dependency bumps


gabemontero:
as an fyi, here is the latest bundle list from jenkins:  https://wiki.jenkins-ci.org/display/JENKINS/Bundled+Plugins

bparees:
lgtm

PR #116, opened 2016-06-23
enable slave configuration

bparees:


bparees:
@gabemontero ptal
@spadgett @tracyrankin with this change we should be able to use the normal openshift/jenkins-1-centos7 image for the pipeline sample, with slave launching, w/o any manual configuration of the image.  The k8s namespace will be automatically configured as well, so you don't have to name your project "pipelineproject" anymore.

bparees:
Note that the configuration will use the jenkinsci/jnlp-slave image by default.  That may change in the future, since among other things, @spadgett has found that image doesn't work for running shell commands, but i wanted to at least start w/ something more official than bparees/jenkinsslave.

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -83,17 +85,30 @@ function generate_kubernetes_config() {
     else
       return
     fi
-    [ -z "${slave_templates}" ] && return
     echo "
-    <org.csanchez.jenkins.plugins.kubernetes.KubernetesCloud plugin=\"kubernetes@0.5\">
+    <org.csanchez.jenkins.plugins.kubernetes.KubernetesCloud plugin=\"kubernetes@0.6\">
```
gabemontero: we should minimally add a comment in base-plugins.txt to update this line when we bump the kubernetes plugin version there.

extra credit would be to somehow grep/sed the version from base-plugin.txt and inject it into this echo

bparees: yeah i'm bothered we need this at all. i'm going to check if it's really required because it's annoying.

bparees:
@gabemontero ok, k8s plugin version has been removed and things still seem to work, ptal.

gabemontero:
Igtm

abdelhegazi:
Hi all

I am running jenkinsci/jenkins as my master jenkins node with kubernetes plugin/openshift plugin all added but I am not able to find the optipon of "restrict where this project can be run"  as mentioned in here https://blog.openshift.com/openshift-3-2-jenkins-s2i-slave-pods/ 

I do understand this will require some plugins, but googling this matter comes back with everyone is requiring installing kubernetes plugin thats it. I am aware that the slave jobs will be pulling this jenkinsci/jnlp-slave image by default, 

Can I ask am I missing other plugins for my jenkins to be able to run my projects in slave containers considering the jenkins version this master container has is 2.32

Any advices, thanks in advance,

bparees:
restrict where this project can be run is part of the jenkins job configuration page.  it's the mechanism by which jenkins assigns job instances to slaves (whether those slaves are run as k8s pods, or traditional jenkins slave instances).

I recommend you look at our jenkins tutorial here that uses our jenkins image which includes all necessarily plugins and predefined configuration:
https://github.com/openshift/origin/tree/master/examples/jenkins

abdelhegazi:
Thanks @bparees it seems like most of the configurations now work fines with all needed plugins installed i am expected it would behave similar to our old mesos jenkins cluster, one little issue with starting a build on a slave container.

- it only works with freestyle project, i can't restrict execution of pipeline projects to run on a separate container.
- Even the freestyle projects which I am executing using shell script, despite no other jobs are running which means "executes are idle" the job hasn't started yet saying waiting for available executor however no other jobs are running and has been like this some good time.
screenshot from 2017-03-02 00-34-34

bparees:
we have some example pipeline projects and a tutorial here that spawns slave pods:
https://github.com/openshift/origin/tree/master/examples/jenkins/pipeline

Issue #117, opened 2016-06-27
maven plugin failes


ullgren:
When using the maven plugin the build failes since the [maven plugin uses the default directory during the initial pom.xml scan](http://osdir.com/ml/java-jenkins-user/2012-03/msg00071.html). The default directory does not seem to be detected correctly in the openshift jenkins image thus the build fails with:

`ERROR: Failed to create ?/.m2`

A according to the post in the java-jenkins-user mailing list a workaround is to set  `-Duser.home=` when starting the jenkins server.

PR #120, opened 2016-06-27
set maven home to /var/lib/jenkins

bparees:
fixes https://github.com/openshift/jenkins/issues/117

@gabemontero ptal

### DIFF
--- a/1/contrib/s2i/run
+++ b/1/contrib/s2i/run
@@ -103,7 +103,7 @@ rm -rf /tmp/war
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_MAX_HEAP_PARAM $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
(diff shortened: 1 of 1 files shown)


gabemontero:
I saw mention of setting `MAVEN_HOME` as well in some stackoverflow / internet searches of the form "jenkins maven home doesn't exist" and similar permutations (along with configuring in the UI which doesn't apply here) as I attempted to get some more background on this.

This lead me to poke around the jenkins core code a bit ... a couple of points:
- it *looks* like MAVEN_HOME will got overridden with the Jenkins HOME location ... https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/tasks/Maven.java#L500
- based on what I see in Jenkins start up (https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/WebAppMain.java#L319-L370 and https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/WebAppMain.java#L405), we might be more *current* if we set JENKINS_HOME to ${HOME} .... `user.home` seems to be labeled as *legacy*

bparees:
@gabemontero hm. we already set JENKINS_HOME: https://github.com/openshift/jenkins/blob/master/1/Dockerfile#L17

so that doesn't seem to be sufficient?

gabemontero:
Heh....something else is going on then.  The code I found seemed straight
forward.

Go ahead and merge what you got.  Ill see if I can build a debug version of
Jenkins core, s2i extend our image with it, and discover the disconnect.

PR #121, opened 2016-07-08
update to 0.10 sync plugin

bparees:


bparees:
seems to be a problem w/ the new version as described here:
https://bugzilla.redhat.com/show_bug.cgi?id=1351129#c9

@jimmidyson

PR #122, opened 2016-07-08
create base and maven slave images

bparees:


bparees:
@gabemontero ptal

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -91,27 +91,28 @@ function generate_kubernetes_config() {
...
+          <label>maven</label>
           <volumes/>
           <envVars/>
           <nodeSelector/>
           <remoteFs>/tmp</remoteFs>
         </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
         <org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
-          <name>redhatdistortion-maven</name>
-          <image>docker.io/redhatdistortion/maven-jenkins-slave</image>
+          <name>nodejs</name>
```
gabemontero: having a nodejs template simply a function of the fact that we are providing a nodejs sample, correct?

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -91,27 +91,28 @@ function generate_kubernetes_config() {
       <name>openshift</name>
       <templates>
         <org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
-          <name>agent</name>
-          <image>jenkinsci/jnlp-slave</image>
+          <name>maven</name>
```
gabemontero: and we are now providing a new maven "sample" out of the box, correct?

1/contrib/jenkins/kube-slave-common.sh
bparees: right, this section is defining a new pod template which uses the new openshift/jenkins-slave-nodejs-centos7 image (which is introduced by this PR).  The label/name of "nodejs" means you can reference this pod template for slave-launching purposes by using the "nodejs" label name in your jenkinsfile/job configuration.

However you've made me realize this this hardcoded config means we're going to use the centos slave image even when we're running a rhel master, which is not what we want, so i need to rework this logic a little to be more customized.

slave-base/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    rpm -V $INSTALL_PKGS && \
+    yum clean all && \
+    mkdir -p /home/jenkins && \
+    chown -R 1001:0 /home/jenkins && \
+    chmod -R g+w /home/jenkins
+# Copy the entrypoint
+ADD contrib/bin/* /usr/local/bin/
+
+# Run the JNLP client by default
+# To use swarm client, specify "/opt/app-root/jenkins/run-swarm-client" as Command
```
gabemontero: Is swarm already getting installed somewhere else (if so, I'm not seeing it)?  If not, should we clarify that somehow (or is it too obvious)?

bparees: nope, that is a bad/leftover comment from something else.  will remove.

gabemontero:
ok ... only a few minor questions (@bparees has already started answering and is making changes for); looking forward to trying it all out

1/contrib/jenkins/kube-slave-common.sh
bparees: right, we're going to be providing two slave images on dockerhub.  one that includes maven, one that includes nodejs.  This config section references them.

bparees:
@gabemontero ok i've updated this to pick the right slave image based on whether the master is centos or rhel.  ptal.

Makefile
```diff
@@ -1,5 +1,5 @@
 SKIP_SQUASH?=0
-VERSIONS="1"
+VERSIONS="1 slave-base slave-maven slave-nodejs"
```
gabemontero: ah .. mean to ask about this before and got distracted .... so does 1 mean "master" image (the "traditional" image we have built), and the others obviously the slaves, where we couldn't change "1" to "master" because backward compatability?  Should we also have a "master" version or a "2" ?

bparees: roughly yes, but we wouldn't change "1" to "master" because at some point in the future we will have both "1" and "2" for master images.

But for the slaves, we aren't going to have multiple versions, so i felt (sort of) comfortable not including a version number in the directory names.

We could rename "1" to "master-1" but those directory names feed heavily into our build infrastructure in terms of dictating how the image ultimately gets named when tagged, so it's easier to leave it as "1".  (the image names become "jenkins-\<directoryname\>-centos7"). 

I also considered a whole new repo for the slave images to avoid this confusion, but it felt like overkill.  I admit we may live to regret this choice but I think we can always separate it out later if that proves the better course of action.

gabemontero: ok .. that helps, and yep, the pros/cons are such that there isn't a slam dunk choice for this.

How about adding a comment that lays out the rationals and implications you just detailed here for future reference?

bparees: done

gabemontero: Cool ... IGTM

PR #124, opened 2016-07-13
install oc binary in slaves and add tests

bparees:


bparees:
@gabemontero ptal, forgot this bit.

gabemontero:
lgtm

slave-base/Dockerfile
```diff
@@ -14,7 +14,11 @@ RUN yum install -y --setopt=tsflags=nodocs --enablerepo=centosplus epel-release
     yum clean all && \
     mkdir -p /home/jenkins && \
     chown -R 1001:0 /home/jenkins && \
-    chmod -R g+w /home/jenkins
+    chmod -R g+w /home/jenkins && \
+    curl -L https://github.com/openshift/origin/releases/download/v1.1.6/openshift-origin-client-tools-v1.1.6-ef1caba-linux-64bit.tar.gz | tar -zx && \
```
tnozicka: @bparees Isn't v1.1.6 a bit old? I would like to suggest to distribute latest stable. (1.2.1)

What would you think about adding also a latest version (v1.3.0-alpha.2) under different name like `oc-latest`. There will soon be a point when we need this for new features like `oc rollout`. But we can download it, although it will make Jenkinsfiles more complicated.

slave-base/Dockerfile
```diff
@@ -14,7 +14,11 @@ RUN yum install -y --setopt=tsflags=nodocs --enablerepo=centosplus epel-release
...
+    mv openshift*/oc /usr/local/bin && \
+    rm -rf openshift-origin-client-tools-*
+
```
tnozicka: @bparees  You could install the binary even without creating/deleting a directory https://github.com/tnozicka/dockerfiles/blob/master/openshift-maven-builder/Dockerfile#L3

bparees: thanks for the tip but I think this approach is syntactically clearer and it's not adding any layers.

slave-base/Dockerfile
bparees: yeah i should have bumped to 1.2.1.  I will do that.  I'm not inclined to start putting alpha stuff in the image though.

tnozicka: ok, that's fair.

slave-base/Dockerfile
tnozicka: ok

PR #125, opened 2016-07-14
move to v1.0.21 of pipeline plugin

gabemontero:
@bparees ptal

i was able to verify that the associated rhel rpm for this version is picked up during rhel build


bparees:
lgtm

Issue #130, opened 2016-07-28
Git is not working properly on slaves


tnozicka:
```bash
docker run -it --rm openshift/jenkins-slave-nodejs-centos7 bash
$ git clone https://github.com/tnozicka/nodejs-ex.git /home/jenkins/nodejs-ex
```
ends up with error:
```
Cloning into '/home/jenkins/nodejs-ex'...
remote: Counting objects: 393, done.
remote: Total 393 (delta 0), reused 0 (delta 0), pack-reused 393
Receiving objects: 100% (393/393), 119.53 KiB | 0 bytes/s, done.
Resolving deltas: 100% (161/161), done.
fatal: unable to look up current user in the passwd file: no such user
Unexpected end of command stream
```

Same for maven one:
```bash
docker run -it --rm openshift/jenkins-slave-maven-centos7 bash
$ git clone https://github.com/tnozicka/nodejs-ex.git /home/jenkins/nodejs-ex
```
```
Cloning into '/home/jenkins/nodejs-ex'...
remote: Counting objects: 393, done.
remote: Total 393 (delta 0), reused 0 (delta 0), pack-reused 393
Receiving objects: 100% (393/393), 119.53 KiB | 0 bytes/s, done.
Resolving deltas: 100% (161/161), done.
fatal: unable to look up current user in the passwd file: no such user
bash-4.2$ Unexpected end of command stream
```

Surprisingly, this is not a problem when running Jenkins on `oc cluster up`. When using ADB/CDK all attempts to checkout repositories on slaves fail with these errors.

(There is probably a different policy to get UID from or something like that.)

I am not sure about the proper solution here, but we had a hack in our images before summit to fix it: https://github.com/tnozicka/dockerfiles/blob/master/openshift-nodejs-builder/Dockerfile#L3

Running the commands like 
```bash
GIT_COMMITTER_NAME='<unknown>' GIT_COMMITTER_EMAIL='<unknown>' git clone https://github.com/tnozicka/nodejs-ex.git /home/jenkins/nodejs-ex
```
helps.

@rupalibehera @bparees

bparees:
The image isn't intended to be run that way, the problem is this:
```
$ docker run -it --rm openshift/jenkins-slave-nodejs-centos7 bash
bash-4.2$ whoami
whoami: cannot find name for user ID 1001
```

the slave startup script that is invoked when you use this image as a proper slave uses nss wrapper to setup a fake passwd file that ensures the uid can be resolved.

[...]

PR #131, opened 2016-07-28
allow nss to configure for uid 1001

bparees:
fixes https://github.com/openshift/jenkins/issues/130


bparees:
@tnozicka ptal.

bparees:
@gabemontero ptal

gabemontero:
lgtm

PR #133, opened 2016-08-08
add read permission to viewers

deads2k:
Adds view permission for the user or groups named "view".

@gabemontero there's no distinction between users and subjects.  This is going to create shadowing risks.  You should consider what you're doing.

### DIFF
--- a/1/contrib/openshift/configuration/config.xml.tpl
+++ b/1/contrib/openshift/configuration/config.xml.tpl
@@ -6,11 +6,11 @@
   <mode>NORMAL</mode>
   <useSecurity>true</useSecurity>
   <authorizationStrategy class="hudson.security.GlobalMatrixAuthorizationStrategy">
-    <permission>hudson.model.Computer.Configure:admin</permission>
     <permission>hudson.model.Computer.Configure:system_builder</permission>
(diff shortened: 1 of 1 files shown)


bparees:
@deads2k what does view permission actually allow and who's going to be in the group view?
(my main concern is around non-logged-in-users, so probably not a problem here, but still)

deads2k:
> @deads2k what does view permission actually allow and who's going to be in the group view?
> (my main concern is around non-logged-in-users, so probably not a problem here, but still)

Users who can "view" "jenkins.build.openshift.io" in the current namespace.  It would be a separate role.

deads2k:
> Users who can "view" "jenkins.build.openshift.io" in the current namespace. It would be a separate role.

Or (because jenkins is crazy) a user with the username "view".  I sent an email describing the slash option, which is probably important for all of our roles here.

bparees:
but what permissions does it grant them in jenkins?

deads2k:
According to the docs: https://wiki.jenkins-ci.org/display/JENKINS/Matrix-based+security, "View almost all pages within Jenkins." and "Grants read-only access to project configurations."

bparees:
thanks

PR #137, opened 2016-08-10
allow jnlp service name to be configured

bparees:


bparees:
@csrwng ptal

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -13,7 +13,15 @@
 
 export DEFAULT_SLAVE_DIRECTORY=/tmp
 export SLAVE_LABEL="jenkins-slave"
-export JENKINS_JNLP_SERVICE_PORT=${JENKINS_JNLP_SERVICE_PORT:-50000}
+JNLP_SERVICE_NAME=${JNLP_SERVICE_NAME:-JENKINS_JNLP}
+T=${JNLP_SERVICE_NAME}_SERVICE_HOST
+# the '!' handles env variable indirection so we can resolve the nested variable
+# see: http://stackoverflow.com/a/14204692
```
csrwng: 👍  this makes reviewing the code a lot easier, esp if you're not a bash expert.

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -13,7 +13,15 @@
...
+JNLP_HOST=${!T}
+T=${JNLP_SERVICE_NAME}_SERVICE_PORT
```
csrwng: Just a nit, why not use a different variable name instead of relying on the sequence of things for T to have the right value?

csrwng:
just one nit, otherwise LGTM

1/contrib/jenkins/kube-slave-common.sh
bparees: I didn't want to clutter things with multiple temp variables, it's the same way i'd have done it in "real" code..

csrwng: that's fine

bparees: updated to use 2 vars.

1/contrib/jenkins/kube-slave-common.sh
bparees: yeah, i had to google to learn how do to that, so i figured i better make a note to myself :)

PR #140, opened 2016-08-11
reformat service name to match expected variable name

bparees:


bparees:
@csrwng ptal, this is needed if we're going to let people define their service name as "jenkins-jnlp"

csrwng:
makes sense, LGTM

Issue #128, opened 2016-07-21
Jenkins service account is not propagated to slaves


tnozicka:
Jenkins slaves are missing service account **jenkins** and have only the default one. They should have the same one as master.

My current workaround is:
```groovy
#!/usr/bin/groovy
def k8s_token = '<unknown>'

node ('master') {
    k8s_token = readFile '/var/run/secrets/kubernetes.io/serviceaccount/token'
}

node ('nodejs') {
    // use ${k8s_token}
}
```

I would like to get rid of it.

bparees:
the k8s plugin claims to support setting a service account in the pod template definition for the node:
https://github.com/jenkinsci/kubernetes-plugin/commit/49cd23fd47e66d79744d5fa6a95767ee33a7d436#diff-7ce44bbe5d164ad0631710f06a682472

but i'm not seeing the config field in the jenkins configuration page.  will investigate further.

bparees:
@iocanel you added the service account config support to the k8s plugin, are you aware of any issues with it not showing up on the jenkins config page?

bparees:
looks like we need a new release of the plugin to get the service account feature, i've requested one:
https://issues.jenkins-ci.org/browse/JENKINS-36891

PR #143, opened 2016-08-16
bump to latest k8s plugin

bparees:
fixes https://github.com/openshift/jenkins/issues/128

### DIFF
--- a/1/contrib/openshift/base-plugins.txt
+++ b/1/contrib/openshift/base-plugins.txt
@@ -2,7 +2,7 @@ openshift-pipeline:1.0.21
 
 # kubernetes plugin - https://wiki.jenkins-ci.org/display/JENKINS/Kubernetes+Plugin
 durable-task:1.10
-kubernetes:0.7
+kubernetes:0.8
 credentials:2.1.0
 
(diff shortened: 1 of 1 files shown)


bparees:
@tnozicka fyi when this goes in you should be able to specify the service account to use as part of the k8s cloud plugin configuration on the manage jenkins page.

bparees:
autoconfiguring it will be slightly more complicated but i'll look into it.

tnozicka:
@bparees I have tested it with our pipelines and here are my findings:

- If you specify only the service account name e.g. "jenkins", it works
- If you want to use namespace+SA name e.g. "myproject:jenkins", it doesn't work
- if you use the full account spec e.g. "system:serviceaccount:myproject:jenkins", it doesn't work

But for us to use it, it has to be pre-configured by default. Using `oc whoami` I am able to get current service account on master and could use it to fill the config template. I will send PR for it next week.

bparees:
@tnozicka thanks.  I think it might be better to configure it from an env variable, and that env variable can be defined in the template that instantiates jenkins.

(the jenkins logic that creates the initial config from a template could consume the env variable in question).

PR #147, opened 2016-08-24
remove base-plugins.txt from rhel image

bparees:


bparees:
@gabemontero ptal

gabemontero:
LGTM

PR #150, opened 2016-09-07
don't override credentials if provided by user

bparees:


bparees:
@gabemontero ptal

gabemontero:
LGTM

btw, did we do more than just fix the credentials piece, given that the config xml check changed and we lost the "Skipping kubernetes plugin configuration..." message ?

bparees:
yeah i reworked it to just always avoid overwriting specific files, as opposed to the existing logic that would either overwrite all or no files, depending on whether a config.xml was present.  As for the lost message, I didn't feel it was all that critical of a message to have.

gabemontero:
ok cool

PR #151, opened 2016-09-09
Add serviceaccount auto-configuration for slaves

tnozicka:
It works only for service account in the same namespace, but that is the limitation of kubernetes cloud plugin in 0.8 as metioned in https://github.com/openshift/jenkins/pull/143#issuecomment-241013800

I have considered [using it from env. variable](https://github.com/openshift/jenkins/pull/143#issuecomment-241089186), but this seems architecturally cleaner. I don't like coupling it to the template and it should be the same service account for master and slaves.

@bparees PTAL

### DIFF
--- a/1/contrib/jenkins/kube-slave-common.sh
+++ b/1/contrib/jenkins/kube-slave-common.sh
@@ -58,6 +58,7 @@ function has_service_account() {
 if has_service_account; then
   export oc_auth="--token=$(cat $AUTH_TOKEN) --certificate-authority=${KUBE_CA}"
   export oc_cmd="oc --server=$OPENSHIFT_API_URL ${oc_auth}"
+  export oc_serviceaccount_name="$(expr "$(oc whoami)" : 'system:serviceaccount:\w\+:\(\w\+\)' || true)"
 fi
 
 # get_imagestream_names returns a list of image streams that match the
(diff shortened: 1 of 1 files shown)


tnozicka:
Fixes https://github.com/openshift/jenkins/issues/128

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -121,6 +122,7 @@ function generate_kubernetes_config() {
           <envVars/>
           <nodeSelector/>
           <remoteFs>/tmp</remoteFs>
+          <serviceAccount>${oc_serviceaccount_name}</serviceAccount>
```
bparees: add this to the "convert is to slave" function too, up on line 87.

bparees:
[test]

1/contrib/jenkins/kube-slave-common.sh
tnozicka: @ Is that function ever called? https://github.com/openshift/jenkins/search?utf8=%E2%9C%93&q=convert_is_to_slave&type=Code

bparees: it used to be. i think it's a bug if it's not anymore.

tnozicka: @bparees It is called here https://github.com/tnozicka/jenkins/blob/master/1/contrib/jenkins/kube-slave-common.sh#L102 .  I am not sure why github search failed there...

I will add it there as well.

bparees: https://github.com/openshift/jenkins/blob/master/1/contrib/jenkins/kube-slave-common.sh#L102

bparees: thanks, you had me panicked :)

bparees:
lgtm
[test]

1/contrib/jenkins/kube-slave-common.sh
tnozicka: sorry about that; I will not relly on GitHub's search any more :(

PR #152, opened 2016-09-09
introduce jenkins v2 image

bparees:
@gabemontero ptal.  rhel image won't merge/pass tests until the new plugin is rpm-packaged and the jenkins 2.7 package exists.

otherwise things went pretty smoothly.


2/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    mv openshift*/oc /usr/local/bin && \
+    rm -rf openshift-origin-client-tools-*
+
+COPY ./contrib/openshift /opt/openshift
+COPY ./contrib/jenkins /usr/local/bin
+ADD ./contrib/s2i /usr/libexec/s2i
+
+RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
+    # need to create <plugin>.pinned files when upgrading "core" plugins like credentials or subversion that are bundled with the jenkins server
+    # Currently jenkins v2 does not embed any plugins, but for reference:
```
gabemontero: How about rewording the comment slightly:
# Currently jenkins v2 does not embed any plugins, so there is no longer a need to create <plugin>.pinned files to upgrade "core" plugins like credentials, but for reference:

2/Dockerfile.rhel7
```diff
@@ -0,0 +1,66 @@
...
+
+# Remove the base-plugins.txt file because it's only used for Centos
+# and its presence in the rhel image is confusing.
+RUN rm /opt/openshift/base-plugins.txt && \ 
+    mkdir -p /opt/openshift/plugins && \
+    # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
+    # HPI file created by rpm to /opt/openshift/plugins folder.
+    for FILENAME in /usr/lib64/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
+    # need to create <plugin>.pinned files when upgrading "core" plugins like credentials or subversion that are bundled with the jenkins server
+    # Currently jenkins v2 does not embed any plugins, but for reference:
```
gabemontero: same comment reword as Dockerfile

2/contrib/jenkins/install-plugins.sh
```diff
@@ -0,0 +1,344 @@
...
+# COPY plugins.txt /plugins.txt
+# RUN /usr/local/bin/install-plugins.sh /plugins.txt
+#
+# The format of 'plugins.txt. is:
+#
+# pluginId:pluginVersion
+
+set -o pipefail
+
+# BEGIN - From https://raw.githubusercontent.com/jenkinsci/docker/master/jenkins-support
```
gabemontero: So I'm assuming that jenkinsci updated their plugin install script since we originally created the V1 version, and that is why they differ ... that correct?

2/contrib/jenkins/fix-permissions
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of 
+# regular files and execute of directories.
+find $1 -exec chgrp 0 {} \;
+find $1 -exec chmod g+rw {} \;
+find $1 -type d -exec chmod g+x {} +
```
gabemontero: since this is the same as the v1 version, can we use symlinks?

2/contrib/jenkins/jenkins-common.sh
```diff
@@ -0,0 +1,41 @@
...
+function obfuscate_password {
+    local password="$1"
+    local salt="$2"
+    local acegi_security_path=`find /tmp/war/WEB-INF/lib/ -name acegi-security-*.jar`
+    local commons_codec_path=`find /tmp/war/WEB-INF/lib/ -name commons-codec-*.jar`
+
+    # source for password-encoder.jar is inside the jar.
+    # acegi-security-1.0.7.jar is inside the jenkins war.
+    java -classpath "${acegi_security_path}:${commons_codec_path}:/opt/openshift/password-encoder.jar" com.redhat.openshift.PasswordEncoder $password $salt
+}
```
gabemontero: same can we use symlinks question?

2/contrib/jenkins/kube-slave-common.sh
```diff
@@ -0,0 +1,168 @@
...
+      <java.util.concurrent.CopyOnWriteArrayList>
+        <org.csanchez.jenkins.plugins.kubernetes.ServiceAccountCredential plugin=\"kubernetes@0.4.1\">
+          <scope>GLOBAL</scope>
+          <id>1a12dfa4-7fc5-47a7-aa17-cc56572a41c7</id>
+          <description></description>
+        </org.csanchez.jenkins.plugins.kubernetes.ServiceAccountCredential>
+      </java.util.concurrent.CopyOnWriteArrayList>
+    </entry>
+    "
+}
```
gabemontero: I think these are the same ... if so, same symlink question

2/contrib/openshift/base-plugins.txt
```diff
@@ -0,0 +1,63 @@
...
+pipeline-utility-steps:1.1.5
+script-security:1.19
+
+# seem to be missing from the Jenkins docs but required to run these plugins..
+structs:1.1
+git-client:1.19.6
+git-server:1.6
+plain-credentials:1.2
+ace-editor:1.1
+matrix-auth:1.4
```
gabemontero: I *could* see the plugin versions diverging over time (if they have not already) ... at least the "core" jenkins plugins ... maybe the openshift ones could be in a common file that is shared, but I don't feel strongly either way

2/contrib/openshift/configuration/config.xml.tpl
```diff
@@ -0,0 +1,62 @@
...
+      <properties/>
+    </hudson.model.AllView>
+  </views>
+  <primaryView>All</primaryView>
+  <slaveAgentPort>${JNLP_PORT}</slaveAgentPort>
+  <label>master</label>
+  <nodeProperties/>
+  <globalNodeProperties/>
+  <noUsageStatistics>true</noUsageStatistics>
+</hudson>
```
gabemontero: could share via symlink if we make the version an env var, no? (assuming the symlink theme I've introduced holds up)

2/contrib/openshift/configuration/credentials.xml.tpl
```diff
@@ -0,0 +1,6 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<com.cloudbees.plugins.credentials.SystemCredentialsProvider plugin="credentials@1.23">
+  <domainCredentialsMap class="hudson.util.CopyOnWriteMap$Hash">
+    ${KUBERNETES_CREDENTIALS}
+  </domainCredentialsMap>
+</com.cloudbees.plugins.credentials.SystemCredentialsProvider>
```
gabemontero: symlink (with understanding, maybe even a comment hinting at, that the symlink might have to be broken if credentials version diverges over time)

slave-maven/Dockerfile.rhel7
```diff
@@ -2,7 +2,7 @@ FROM openshift/jenkins-slave-base-rhel7
 
 MAINTAINER Ben Parees <bparees@redhat.com>
 
-ENV MAVEN_VERSION=3.3.9
+ENV MAVEN_VERSION=3.0.5
```
gabemontero: interesting ... we had to go down/backward in the maven version

2/contrib/openshift/configuration/jobs/OpenShift Sample/config.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no ?

2/contrib/openshift/configuration/users/admin/config.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no?

2/Dockerfile
bparees: I did it this way so we can just delete line 47-48 if we need to add a pin line and be back to where we were.

2/contrib/s2i/assemble
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
gabemontero: symlink yes/no?

2/contrib/s2i/run
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
gabemontero: symlink yes/no?

2/test/run
```diff
@@ -0,0 +1,189 @@
+#!/bin/bash
```
gabemontero: symlink yes/no?

2/test/s2i/configuration/config.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no?

2/test/s2i/configuration/jobs/sample-app-test/config.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no?

2/test/s2i/configuration/jobs/sample-app-test/promotions/Verified/config.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no?

2/test/testjob.xml
```diff
@@ -0,0 +1,198 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gabemontero: symlink yes/no?

2/contrib/jenkins/fix-permissions
bparees: we don't do that in any of our other image repos that share files between versions today, and i'm kind of leery of symlinks 

1) in git repos in general
2) when we start docker add'ing files that are symlinks

i know it's kind of painful (and has been the source of complaints before) but i'd rather keep the same model here until/unless we adopt a wholesale solution.

2/contrib/jenkins/install-plugins.sh
bparees: yeah, they claim plugins.sh is deprecated and recommend this instead.

gabemontero:
You are already responding :-) but to summarize, a couple of minor comment suggestions and questions, along with the "should we use symlinks for files which seemed the same between v1 and v2?".

Otherwise, totally cool that things went smoothly - IGTM

2/contrib/openshift/base-plugins.txt
bparees: matrix-auth is already unique to the v2 list as a result of it being a core plugin in v1.

2/Dockerfile
gabemontero: OK, sure, it is still easy enough to discern what is up, I'm fine with it.

slave-maven/Dockerfile.rhel7
bparees: nah, we were just never actually at 3.3.9. We are installing 3.3.9 in centos, but the rhel package is 3.0.5, so i was just fixing the misinformation.

gabemontero: 10-4

2/contrib/jenkins/fix-permissions
gabemontero: I see ... I was able to use them in jenkins-plugin with no ill-affects (re: point 1)), but yeah I've experienced enough now that 2) resonates (and 2) did not apply in the jenkins-plugin case)

I'll defer.

2/contrib/jenkins/install-plugins.sh
gabemontero: ok cool

2/contrib/openshift/base-plugins.txt
gabemontero: yeah i figured i may have missed a diff when just doing an eye inspection ... sounds good

gabemontero:
OK ... symlink question resolved (we can always change it later if updating 2 versions gets too onerous and we satisfy our Docker build concerns).

LGTM

bparees:
my usual process for managing the version diffs is to run a diff between the "1" and "2" directories and make sure the only deltas are ones that make sense.  It is definitely not free, but so far it's been a manageable process and really it's one you need to do anyway since there could be files that have small differences and thus can't be symlinked anyway, but still need to stay largely the same.

bparees:
[test]

bparees:
[test]

bparees:
[test]

bparees:
[test]

bparees:
[test]

bparees:
[test]

PR #159, opened 2016-10-07
show download error if any

akostadinov:
When launching slave it is likely to hit errors like non-trusted jenkins server certificate.
Without the `-S` option, curl exits with some status but no indication as to
what the problem actually is. `-S` option only tells curl to print things when
an error is hit. Not output will be present (as before) when things go smoothly.

btw, how do I take care of bad certificate? Any way I can setup server cert
so it is accepted without modifying Dockerfile?

### DIFF
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -31,7 +31,7 @@ if [[ $# -gt 1 ]]; then
   fi
 
   echo "Downloading ${JENKINS_URL}/jnlpJars/remoting.jar ..."
-  curl -s ${JENKINS_URL}/jnlpJars/remoting.jar -o ${JAR}
+  curl -sS ${JENKINS_URL}/jnlpJars/remoting.jar -o ${JAR}
 
   # if -tunnel is not provided try env vars
   if [[ "$@" != *"-tunnel "* ]]; then


bparees:
[test]

bparees:
@akostadinov no i don't think there's a way to do that currently w/o augmenting the slave image.

akostadinov:
I think probably we could make that possible using secret volume and then use `update-ca-trust` with certs from secret volume to allow curl and jnlp slave connect flawlessly. Hope I have time some day to take a crack at it.
I think most people will augment image anyway but having ability to share image publicly without exposing internal details would be useful for community to  share images.

bparees:
@akostadinov you'll need to take that request to the jenkins kubernetes plugin which is what this image relies on for launching slaves.  Any configuration of the slave pod has to be exposed by that plugin, as part of the kubernetes cloud plugin slave pod template configuration in jenkins.  Right now the kubernetes plugin allows you to list volumes to be mounted but it doesn't allow direct reference of secrets to be mounted, so you'd have to manually create the secret volume and then reference that volume in the slave pod template definition.

akostadinov:
@bparees thanks for clarification. I will file a bug.

btw `-sS` seems not enough to show if file has been downloaded. If request returns HTTP 403, then curl exits with no output and zero status. I can look at proper curl parameters to catch such situations and print some output. Would you agree with such a change?

P.S. I find k8s slaves plug-in working fast and reliably, kudos to all involved in its development

bparees:
> Would you agree with such a change?

sure

PR #169, opened 2016-10-20
enable the ose33 repo so the rhel images can be built by users

bparees:


bparees:
@tdawson ptal

bparees:
[test]

tdawson:
LTGM :+1:

PR #173, opened 2016-10-26
add blue ocean plugin to centos v2 and bump sync plugin

bparees:


bparees:
@gabemontero ptal

bparees:
fyi i'm not intending to have @tdawson  package blueocean for rhel given its beta state, but we will need him to repackage the new sync plugin version.

gabemontero:
ah ... sorry for the delay @bparees

PR #176, opened 2016-10-28
remove -f on docker tag

bparees:


bparees:
@gabemontero let's see how this goes, but we'll have to update the vagrant plugin to to make the actual push jobs work.

bparees:
[test]

PR #179, opened 2016-11-01
Updating labels on Dockerfile.rhel7

tdawson:


bparees:
lgtm, thanks @tdawson!

PR #180, opened 2016-11-01
go to v0.6 of openshift login

gabemontero:
@bparees fyi

v0.6 of openshift-login includes debug around problems accessing the master for defaults when within a pod that we discussed in scrum, as well as closing a window for improperly defaulting to oauth that surfaced when helping out @oatmealraisin this morning.


gabemontero:
[test]

PR #188, opened 2016-11-07
go to v0.8 of login plugin

gabemontero:
v0.7 somehow missed a previous fix around inadvertent defaulting; also, 1 more small tweak in the logging was needed to clearly diagnose the issue in @csrwng 's env

with both changes, we were able to a) successfully *NOT* default to oauth, b) confirm a config/permissions issue existed in @csrwng 's env

@bparees fyi


gabemontero:
test job passed ... merging

Issue #189, opened 2016-11-08
duplicate <remoteFs> tags


jim-minter:
<remoteFs> tag is duplicated at https://github.com/openshift/jenkins/blob/master/1/contrib/jenkins/kube-slave-common.sh#L119-L125 and https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L119-L125

This causes a "You have data stored in an older format and/or unreadable data." error at https://jenkins-demo.router.default.svc.cluster.local/manage .

jim-minter:
It's not instantly clear to me what the correct value of <remoteFs> should be - /home/jenkins doesn't exist inside the pod, and is referenced in other places, so a global search/replace may be necessary in addition to removing the duplicated XML.

bparees:
/home/jenkins is created in slave-base, so it will exist in those slave images:
https://github.com/openshift/jenkins/blob/master/slave-base/Dockerfile#L14

but i think /tmp is a better path so i'll remove the /home/jenkins references from the podtemplate definition.

PR #190, opened 2016-11-08
remove duplicated remotefs definition

bparees:
fixes https://github.com/openshift/jenkins/issues/189

@jim-minter ptal


bparees:
[test]

jim-minter:
@bparees change looks good to me assuming it works on /tmp :)

PR #196, opened 2016-11-14
Added cccp.yml files for building on container pipeline.

mohammedzee1000:
This is for the purpose of building these containers on the
CentOS Container Pipeline.

### DIFF
--- a/1/cccp.yml
+++ b/1/cccp.yml
@@ -1,0 +1,41 @@
+# This is for the purpose of building containers on the CentOS Community Container 
+# Pipeline. The containers are built, tested and delivered to registry.centos.org and
+# lifecycled as well. A corresponding entry must exist in the container index itself, 
+# located at https://github.com/CentOS/container-index/tree/master/index.d
+# You can know more at the following links:
+# * https://github.com/CentOS/container-pipeline-service/blob/master/README.md
+# * https://github.com/CentOS/container-index/blob/master/README.rst
+# * https://wiki.centos.org/ContainerPipeline
+
+# This will be part of the name of the container. It should match the job-id in index entry
+job-id: jenkins-1-centos7 
+
+#the following are optional, can be left blank
+#defaults, where applicable are filled in
+#nulecule-file   : nulecule
+
+# This flag tells the container pipeline to skip user defined tests on their container
+test-skip       : True
+
+# This is path of the script that initiates the user defined tests. It must be able to
+# return an exit code.
+test-script     : null
+
+# This is the path of custom build script.
+build-script    : null
+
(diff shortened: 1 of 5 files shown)


mohammedzee1000:
Linked with https://github.com/CentOS/container-index/pull/75

bparees:
Are these images being built on the centos container pipeline?  what are they being published as?

this repo is already built+published via openshift CI/CD tooling, as openshift/jenkins-XXXX-centos7

mohammedzee1000:
@bparees they will be, once we have cccp.yml entries in. I could go ahead and build from my fork for now, but we should eventually build from upstream repos. 

Any container build on the pipeline are delivered to registry.centos.org. There are documentation links of the process in the cccp.yml files itself that should give you more information. 

The openshift containers themselves are currently already being delivered through the container pipeline.

bparees:
@mohammedzee1000 what are they being published as?  when do they get rebuilt/updated?

mohammedzee1000:
@bparees As of now they will get updated whenever there a change in the repository. 

We are working on job dependency side as well which will cascade a rebuild whenever a base image gets rebuilt for any reason and should have that soon.

Also in pipeline, is rebuilding containers if rpms get updated in the repositories.

bparees:
@mohammedzee1000 I still haven't seen where you are publishing them.  What's the registry hostname? what's the image repository name?  how will they be tagged?

i'm concerned that we already publish these images as docker.io/openshift/jenkins-XXXX-centos7 so what's the value (vs confusion) if we start publishing them in two places?

mohammedzee1000:
@bparees https://github.com/CentOS/container-index/blob/master/index.d/index_template.yml is how an entry in the container index looks like. The container is named based as registry.centos.org/app-id/job-id:desired-tag. 

The container is life-cycled regularly and also scan functionality has been added along with a dockerfile linter which checks that a dockerfile is following container best practices. 

You can know more about the scanner in the blog post here https://seven.centos.org/2016/11/introducing-centos-container-image-scanners/. These scan will run at build and also weekly, to ensure containers do not have known issues or vulnerabilities and users will be informed of the same.

The name can be kept same, if necessary (openshift/jenkins-xxxx-centos7) by adding the entry to https://github.com/CentOS/container-index/blob/master/index.d/openshift.yml.

I understand your concern about duplication, but r.c.o is the official centos registry and future centos releases will automatically goto r.c.o for their content. Also, docker hub is, for all practical purposes, under control of Docker company, while r.c.o is a community registry. 

Also, we can simultaneously push the built container to docker hub as well, if needed.

@kbsingh @bamachrn , if you wish to add anything, your welcome

PR #197, opened 2016-11-15
Adding low memory profile for running Jenkins

jupierce:
See openshift/origin#11795 for rationale. 
@bparees @gabemontero ptal


bparees:
lgtm

PR #198, opened 2016-11-16
go to v1.0.34 of openshift-pipeline

gabemontero:
pulls in the openshift-pipeline fix for https://bugzilla.redhat.com/show_bug.cgi?id=1395555 for jenkins-*-centos7

@bparees FYI

@tdawson we'll need a new RPM for v1.0.34 of openshift-pipeline (the other RPMs are up to date) and a respin of jenkins-*-rhel7 on brew-pulp - thanks

@dongboyan77 FYI

### DIFF
--- a/1/test/run
+++ b/1/test/run
@@ -11,6 +11,8 @@ shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/jenkins-1-centos7-candidate}
 
+FAILURE=true
+
 CIDFILE_DIR=$(mktemp --suffix=jenkins_test_cidfiles -d)
 VOLUME=`mktemp -d`
 chmod a+rwx $VOLUME
@@ -22,10 +24,8 @@ function cleanup() {
     echo "Stopping and removing container $CONTAINER..."
     docker stop $CONTAINER
     exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
-    echo "Dumping logs for $CONTAINER"
-    # Dump the logs always (jenkins returns 143 when killed), so we can
-    # identify the failures in Jenkins
-    [[ "${exit_status}" != "143" ]] && docker logs $CONTAINER
+    [[ "${FAILURE}" == "true" ]] && echo "Dumping logs for $CONTAINER"
+    [[ "${FAILURE}" == "true" ]] && docker logs $CONTAINER
     docker rm $CONTAINER
     rm $cidfile
     echo "Done."
@@ -49,7 +49,7 @@ function get_container_ip() {
 test_connection() {
   set +e
   echo "Testing HTTP connection..."
-  local max_attempts=10
+  local max_attempts=20
   local sleep_time=6
   local attempt=1
   local result=1
@@ -137,7 +137,6 @@ function run_tests() {
   set -e
   # Stop the container so next tests can run
   docker stop $(get_cid $name)
-  echo "  Success!"
 }
 
 function run_s2i_tests() {
@@ -185,5 +184,7 @@ run_tests jenkins_test
 
 # S2I tests
 run_s2i_tests
+
+FAILURE=false
 
 echo "SUCCESS!"
--- a/1/contrib/openshift/base-plugins.txt
+++ b/1/contrib/openshift/base-plugins.txt
@@ -1,4 +1,4 @@
-openshift-pipeline:1.0.33
+openshift-pipeline:1.0.34
 openshift-login:0.9
 
 
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -1,4 +1,4 @@
-openshift-pipeline:1.0.33
+openshift-pipeline:1.0.34
 openshift-login:0.9
 
 
--- a/2/test/run
+++ b/2/test/run
@@ -11,6 +11,8 @@ shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/jenkins-1-centos7-candidate}
 
+FAILURE=true
+
 CIDFILE_DIR=$(mktemp --suffix=jenkins_test_cidfiles -d)
 VOLUME=`mktemp -d`
 chmod a+rwx $VOLUME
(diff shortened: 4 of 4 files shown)


gabemontero:
[test]

gabemontero:
on these intermittent test job failures, we are not dumping the container logs cause of that rc == 143 check after the Testing HTTP Connection... does not get the expected results ... I'm leaning toward adding a change to our test script that just always dumps the jenkins container.

bparees:
I really don't want to do that, it makes the runs incredibly messy.  Let's make it conditional.  set the conditional to false at the end of the test script (before we print SUCCESS at the very end) so we don't dump the container if all the tests were successful.

we should also get rid of the echo "  Success!" within the run_tests function, because it's weird to print success in multiple places (one for each test, one overall).

if we want to print something for each test, let's make the two messages obvious that one is test case success and one is overall success.

tdawson:
rpm built:
  jenkins-plugin-openshift-pipeline-1.0.34-1.el7
images built:
  openshift3/jenkins-1-rhel7:1.651.2-34
  openshift3/jenkins-2-rhel7:2.7-19

gabemontero:
thanks @tdawson !

bparees:
[test]

gabemontero:
running some local tests with the conditional dumping of container logs as prescribed by @bparees

gabemontero:
Interesting ... I saw an issue with "Testing HTTP Connection..." locally when testing the conditional container dump.  Not sure what to make of it, but it would be curious if the same thing is happening with the PR testing:

```
INFO: Started all plugins
Nov 17, 2016 12:57:12 AM hudson.ExtensionFinder$GuiceFinder$FaultTolerantScope$1 error
WARNING: Failed to instantiate optional component org.jenkinsci.plugins.workflow.steps.scm.GitStep$DescriptorImpl; skipping
Nov 17, 2016 12:57:12 AM io.fabric8.kubernetes.client.Config tryServiceAccount
WARNING: Error reading service account token from: [/var/run/secrets/kubernetes.io/serviceaccount/token]. Ignoring.
```

gabemontero:
rhel mirror repo issues now ... trying one more time ... [test]

gabemontero:
still with the mirrors ... will try again either later tonight or tomorrow AM

tdawson:
There was a problem when syncing to the mirrors.  It's been fixed and things should work now.

gabemontero:
thanks for the info @tdawson ... [test]

1/test/run
```diff
@@ -11,6 +11,8 @@ shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/jenkins-1-centos7-candidate}
 
+FAILURE=`echo true`
```
gabemontero: It was causing funny highlighting in my emacs bash editor and i wanted to make sure i got something functional to test ... I'll work on switching

bparees:
it looks to me like we're failing because jenkins doesn't quite start in 1 minute (perhaps due to our heap changes, new plugins, etc):

```
Testing HTTP connection...
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Stopping and removing container 8f323239f946244d9193d136dcb73d61ded0c871dc43b858129271a7370bd858...
8f323239f946244d9193d136dcb73d61ded0c871dc43b858129271a7370bd858
Dumping logs for 8f323239f946244d9193d136dcb73d61ded0c871dc43b858129271a7370bd858
```

I suggest we bump the number of retries to 12 (it waits 10 seconds between each retry)

gabemontero:
The new form of dumping the container on an error appears to have been informative.  Best as I can tell, the http test connection simply gave up too soon.  There were two instances.  In the second, it seemed obvious that Jenkins was in the middle of its startup.  The first, it seemed to come up, but can't say for sure if it finished coming up between when the test stopped and when the container was dumped.

I'm going to try and increase the time period, just to see what effect it has (assuming the consistent failures this evening continue)

bparees:
heh, sounds like we're on the same page :)

gabemontero:
ok ... our common analysis passed through the ether :-)

bparees:
one other strange thing is it appears the output is comingled, not sure why that is but it makes it a lot harder to follow what failed and what logs are from what test.

bparees:
either that or we're actually getting the wrong container id, which would explain the failures better.

see:
```
Created container e763f2e93b2962138c7fefd0a0a523611e341e0a89d5883294270afbd3439e7d
Testing HTTP connection...
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Sending GET request to http://172.17.0.2:8080/
Stopping and removing container 8f323239f946244d9193d136dcb73d61ded0c871dc43b858129271a7370bd858...
8f323239f946244d9193d136dcb73d61ded0c871dc43b858129271a7370bd858
```

gabemontero:
looks like cleanup() dumps *every* container that was created

bparees:
hm. good catch.

gabemontero:
we *could* dump the container in test_connection if $result is non-zero ... be more precise on which container we had trouble with

bparees:
at a minimum we should be printing a message in there indicating things failed(ie it's giving up after 10 retries), because right now it takes too much detective work to realize that what failed was it hitting its limit of 10 retries and bombing out.

trying to dump the specific container might be too much work (given all the ways/places things could fail.  it'd be easy to solve for this one, but not generally for all the tests)

gabemontero:
yes agreed easy for this case (since the container name is passed into test_connection by its callers, even if it is not used currently) but in general more complicated

btw, it passed with the longer grace period

merge now to get the centos images updated, and continue tweaking the debug in a separate pull ?

bparees:
yeah. (merged)

PR #200, opened 2016-11-18
bump to LTS v2.19.3

bparees:


bparees:
@gabemontero @jupierce fyi in case we see any weirdness after this change...
@tdawson fyi, updates our centos image.  once the rpm is available i can update the dockerfile.rhel7 if you care, though i know you're carrying patches anyway.

tdawson:
Changes are in the last jenkins-2 build
  openshift3/jenkins-2-rhel7:2.7-20

bparees:
@tdawson why is it tagged as :2.7-20 if it's v2.19?

tdawson:
*Facepalm* That's what I get for manually merging ... one fix, coming up.

tdawson:
Note: You only updated Dockerfile, and not Dockerfile.rhel7

bparees:
yeah i'm doing the rhel7 file over here, because i didn't know if the rpm was ready yet and didn't want this PR to fail testing because it couldn't install the new rpm:
https://github.com/openshift/jenkins/pull/201

PR #203, opened 2016-11-19
re-bump to LTS v2.19.3, update related plugins, and add tests to confirm plugins load successfully

bparees:
This reverts commit 9ff112fbae7bc66c13def4091459b1dd47ba81cd.


bparees:
@tdawson the changes to base-plugin.txt should reflect the plugin version that need to be repackaged for rhel.

PR #214, opened 2016-12-07
Simple check to allow for s2i build

luigizuccarelli:
Motivation:
A need for the 'slave-base' to be used as a base for a s2i build
The s2i build for some reason passes more than 2 parameters when executed and causes an error when trying to curl https://github.com/openshift/jenkins/blob/master/slave-base/contrib/bin/run-jnlp-client#L34

Changes:
A simple check for the JENKINS_URL envar (if empty will not error for s2i builds)

Please could you review and comment

### DIFF
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -17,7 +17,8 @@ export HOME=${JENKINS_HOME}
 set -e
 
 # if `docker run` has 2 or more arguments the user is passing jenkins launcher arguments
-if [[ $# -gt 1 ]]; then
+# also check the JENKINS_URL  envar as this is useful for s2i builds
+if [ $# -gt 1 ] && [ ! -z $JENKINS_URL ]; then
   JAR="${JENKINS_HOME}/remoting.jar"
   PARAMS=""
 


bparees:
I need to better understand what you're trying to do.  If you're using slave-base as an s2i builder image, then you need to be supplying assemble and run scripts(this image is not by itself enabled for s2i builds).  And then any issue with how run-jnlp-client is being invoked would be related to your run script, since your run script will now be the entrypoint for the image instead of the run-jnl-client script.

luigizuccarelli:
@bparees - I have created an image with the relevant assemble and run script etc for s2i builds. I based it on your slave-base. Your suggestion makes sense, I can simply include the contents of the slave-base Dockerfile and add the scripts folder *contrib* to my current *contrib* folder and add the relevant entry point - wdyt ?

bparees:
You shouldn't need to do that.  Having your image extend this image is the right way to build your s2i builder image, but again, whatever problem you're having is related to the content of the "run" script you are introducing and what it is doing when it invoked the run-jnlp-client script.  Can you show me your run script?

luigizuccarelli:
Actually I have deleted the run script as it does not start any service.

bparees:
you can't have an s2i builder image w/o a run script.

luigizuccarelli:
I have added a script that now checks for an envar and if its not set then call the orignal endpoint i.e /usr/local/bin/run-jnlp-client with parameters , busy checking that now.  

I will close this PR. 

Thnaks again for you help , much appreciated

PR #215, opened 2016-12-14
Update to oc whoami regex to not exclude project names containing '-'.

etsauer:
The current `kube-slave-common.sh` script uses a regular expression to strip a service account name like 'system:serviceaccount:namespace:default' down to just the username, 'default'. However, the existing regex does not support all possible namespace/project names. Namely, a namespace like 'my-namespace' is a valid resource name, but '-' does not match the expression used in that script. The regex in this change more closely resembles the one used to validate namespace names in k8s, (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`)

We can't quite match exactly what k8s uses because `expr` doesn't support extended regular expressions, but this is very close.

@bparees

### DIFF
--- a/1/contrib/jenkins/kube-slave-common.sh
+++ b/1/contrib/jenkins/kube-slave-common.sh
@@ -58,7 +58,7 @@ function has_service_account() {
 if has_service_account; then
   export oc_auth="--token=$(cat $AUTH_TOKEN) --certificate-authority=${KUBE_CA}"
   export oc_cmd="oc --server=$OPENSHIFT_API_URL ${oc_auth}"
-  export oc_serviceaccount_name="$(expr "$(oc whoami)" : 'system:serviceaccount:\w\+:\(\w\+\)' || true)"
+  export oc_serviceaccount_name="$(expr "$(oc whoami)" : 'system:serviceaccount:[a-z0-9][-a-z0-9]*:\([a-z0-9][-a-z0-9]*\)' || true)"
 fi
 
 # get_imagestream_names returns a list of image streams that match the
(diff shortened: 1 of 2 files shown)


etsauer:
If we wanted to match exactly what kube validation is, we could replace `expr` with `sed -r`, but I don't know if there was a reason `expr` was chosen.

jkupferer:
Here is what the command would look like in sed if we want to do full validation of the namespace and service account names:
whoami | sed -r 's/^system:serviceaccount:([a-z0-9]([-a-z0-9]*[a-z0-9])?):([a-z0-9]([-a-z0-9]*[a-z0-9])?)/\3/'

bparees:
can we not just change the regex to:
`'system:serviceaccount:.\+:\(.\+\)'`

?

etsauer:
We could, but what's in the PR is already less restrictive than the
kubernetes rule, so it's not necessary to be that wide open. If you want it
that way for simplicity I can make the change.

bparees:
yeah i was just trying to simplify/future-proof us, but let's go w/ what you've got.  thanks!

bparees:
looks like one of the plugins in our image has rotted and needs updating before this will pass our CI.  @gabemontero can you look into bumping our plugin versions and getting @tdawson to package the new ones?  See the errors here:

```
SEVERE: Failed Loading plugin Pipeline: Model Definition v0.7.1 (pipeline-model-definition)
java.io.IOException: Pipeline: Model Definition v0.7.1 failed to load.
 - Pipeline: API v2.5 is older than required. To fix, install v2.7 or later.
```

(and when you bump the pipeline-api plugin, who knows what else will fail)

bparees:
ok plugins bumped here https://github.com/openshift/jenkins/pull/217
[test]

etsauer:
@bparees how do I see what version of the plugin this will show up in?

bparees:
@etsauer version of the plugin? this was a change to the image.  it's already in the centos image on dockerhub.  @tdawson would be the one to know if it will be in the upcoming rhel image that will go out with OCP3.4.

tdawson:
This change is in the upcoming images that come out with openshift 3.4, which should be images
openshift3/jenkins-1-rhel7:1.651.2-45
openshift3/jenkins-2-rhel7:2.19-12

etsauer:
@tdawson thanks.

PR #217, opened 2016-12-16
update workflow-api plugin version

bparees:


bparees:
@tdawson fyi.  again not needed for 3.4, but probably for 3.5 we'll want to include this update.

PR #219, opened 2017-01-04
Added ruby slave based on enabling scl repos

AdamSaleh:
Hi,

my team is planning to use Jenkins on Openshift, and we found a need to have a ruby based slave.
So I made one :-) 

So far 

```
make test VERSION=slave-ruby TARGET=centos7
```

ended up with 

```
2.5.1
sh: cannot set terminal process group (-1): Inappropriate ioctl for device
sh: no job control in this shell
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
SUCCESS!
```

where "2.5.1" and "ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]" are expected versions of both `gem --version` and `ruby --version` and I can report that we already are testing the image internally in our proof-of-concept jenkins pipe-lines.

So, anything more that needs to be done to have this in the official repo? 

Adam Saleh

### DIFF
--- a/slave-ruby/Dockerfile
+++ b/slave-ruby/Dockerfile
@@ -2,16 +2,6 @@ FROM openshift/jenkins-slave-base-centos7
 
 MAINTAINER Adam Saleh <asaleh@redhat.com>
 
-# Labels consumed by Red Hat build service
-LABEL com.redhat.component="jenkins-slave-ruby-rhel7-docker" \
-      name="openshift3/jenkins-slave-ruby-rhel7" \
-      version="1" \
-      architecture="x86_64" \
-      release="1" \
-      io.k8s.display-name="Jenkins Slave Ruby" \
-      io.k8s.description="The jenkins slave ruby image has the ruby tools on top of the jenkins slave base image." \
-      io.openshift.tags="openshift,jenkins,slave,ruby"
-
 ENV RUBY_VERSION=2.3 \
     BASH_ENV=/usr/local/bin/scl_enable \
     ENV=/usr/local/bin/scl_enable \
--- a/slave-ruby/Dockerfile
+++ b/slave-ruby/Dockerfile
@@ -1,0 +1,32 @@
+FROM openshift/jenkins-slave-base-centos7
+
+MAINTAINER Adam Saleh <asaleh@redhat.com>
+
+# Labels consumed by Red Hat build service
+LABEL com.redhat.component="jenkins-slave-ruby-rhel7-docker" \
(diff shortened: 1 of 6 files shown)


gabemontero:
To be consistent with our other two slave images, a k8s plug pod template config would be added to the default cloud config the image provides.  See https://github.com/openshift/jenkins/blob/master/1/contrib/jenkins/kube-slave-common.sh#L99-L153 and https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L99-L153.

I'll defer to @bparees on long term desire to carry a ruby slave image.

FYI - the PR test failure seems orthogonal to this change (a plugin could not be loaded because of version issues).

slave-ruby/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM openshift/jenkins-slave-base-centos7
+
+MAINTAINER Adam Saleh <asaleh@redhat.com>
+
+# Labels consumed by Red Hat build service
+LABEL com.redhat.component="jenkins-slave-ruby-rhel7-docker" \
+      name="openshift3/jenkins-slave-ruby-rhel7" \
+      version="1" \
+      architecture="x86_64" \
+      release="1" \
```
bparees: all the labels above this line should be removed (they're only relevant to the rhel dockerfile).

slave-ruby/test/run
```diff
@@ -0,0 +1,18 @@
...
+#
+# Test the Jenkins image.
+#
+# IMAGE_NAME specifies the name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+# make sure oc is installed
+docker run ${IMAGE_NAME} oc
+# make sure npm and nodejs are available
```
bparees: comment needs to be updated.

danmcp:
@bparees Still sounds right to me.

AdamSaleh:
@bparees our usecase?

We have large parts of our deployment and infrastructure automation scripts written in ruby (knife and rest of the chef tooling mostly). We want to have it  as a part of our continuous deployment pipeline.

AdamSaleh:
> I'm definitely wary of us ending supporting/maintaining a large set of slave images (today we have nodejs and maven, likely if we go down this road we'll end up adding ruby, perl, php, and python).

@bparees, I assumed that you want to have it this way, based on the fact that you have python and maven separately :-) I have no problem with having one image that would support all of them.

PR #223, opened 2017-01-10
update dependent plugin versions

bparees:


bparees:
@gabemontero fyi, more plugin versions have bit-rotted.

@tdawson post 3.4, we're going to need to refresh/update a bunch of the dependent plugins we package.

gabemontero:
@bparees 10-4

PR #224, opened 2017-01-10
update login plugin with bearer token support, update readme for bear…

gabemontero:
…er token support

@bparees FYI


bparees:
lgtm

PR #225, opened 2017-01-13
bump k8s plugin to 0.10

bparees:


bparees:
@smarterclayton @gabemontero fyi

Issue #220, opened 2017-01-07
change .metadata.annotations  to .metadata.tags 


mykemd:
Had a small issue with autodiscovery jenkins slaves from image streams in the project.
In the template annotations is used for optional slave-label annotation.
Maybe it should be changed to tags since tags are always present in the metadata of the image stream.

bparees:
are you referring to this logic which finds imagestreams with the slave label?
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L68

mrmikedoe:
Yes.
My jenkins was bombing with null error because this portion of the template  was failing

<label>{{if index .metadata.annotations \"slave-label\"}}{{index .metadata.annotations \"slave-label\"}}{{else}}${name}{{end}}</label>

It's not a big deal - just have to make sure annotations are present - or else template default logic to set label to image name won't work

bparees:
hm, well we'll definitely want to fix the template to at least not bomb. i'll look into it, thanks for the report!

bparees:
i'm not able to recreate this, i defined an imagestream with just the role=slave-label label:

```
apiVersion: v1
kind: ImageStream
metadata:
  annotations:
    openshift.io/display-name: Ruby
    openshift.io/image.dockerRepositoryCheck: 2017-01-12T21:04:58Z
  creationTimestamp: 2017-01-12T21:04:56Z
  generation: 2
  labels:
    role: jenkins-slave
  name: ruby
  namespace: test
  resourceVersion: "983"
  selfLink: /oapi/v1/namespaces/test/imagestreams/ruby
  uid: c59c1ac5-d90a-11e6-9c9d-507b9d27b5d9
```

and my jenkins config was populated correctly:
```
  <org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
    <name>ruby</name>
    <image>172.30.19.155:5000/test/ruby</image>
    <privileged>false</privileged>
    <command></command>
    <args></args>
    <instanceCap>5</instanceCap>
    <volumes/>
    <envVars/>
    <nodeSelector/>
    <serviceAccount>jenkins</serviceAccount>
    <remoteFs>/tmp</remoteFs>
    <label>ruby</label>
  </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
```

can you provide any additional details about how to recreate this, or what version of the jenkins image you are were using?

mykemd:
[...]

PR #226, opened 2017-01-13
tolerate empty imagestream annotations field for slave lookup

bparees:
fixes https://github.com/openshift/jenkins/issues/220


bparees:
@mfojtik you did the original I think and probably know templates better than me, ptal.

mfojtik:
LGTM (I thought annotations are defaulted...)

bparees:
@mfojtik i'm at least find the that the repository check annotation immediately gets added to any imagestream i create, so it's pretty hard to hit this, but the user who hit this was on 3.3, perhaps behavior was different then.  I was able to recreate it by deleting the annotation block from the IS and then quickly running the template, so it's definitely not defaulted.

PR #228, opened 2017-01-13
disable sshd in jenkins image

gabemontero:
see https://github.com/openshift/origin/issues/12476

@bparees ptal - this is for disabling sshd per the discussion in the referenced issue


bparees:
[test]

bparees:
in theory lgtm, but the test failed, looks like maybe a network flake while installing the jenkins plugins.

gabemontero:
test passed ... merged

PR #229, opened 2017-01-15
indicate slave images are jnlp images

bparees:


bparees:
@gabemontero @mmilata @smarterclayton fyi.  this may fix some of the extended test issues we've seen w/ pipelines (I haven't audited the failures to see if any of them were related to us trying to run steps in the wrong container now that the k8s plugin was putting 2 containers in the pod)

PR #230, opened 2017-01-16
optimize jvm gc/heap settings for lower memory usage

bparees:


bparees:
@gabemontero ptal

gabemontero:
Wow, some serious and specific `-XX` usage.  Did it originate from a recommendation from someone, or from GC analysis of note?

bparees:
yeah it was recommendations we got from the JDK team back in the v2 days, but for some reason we didn't carry it forward into v3, see:
https://github.com/openshift/origin-server/blob/master/cartridges/openshift-origin-cartridge-jenkins/bin/control#L50-L58

gabemontero:
i see ... ok, cool ... lgtm

PR #231, opened 2017-01-16
set jnlp args on slave pod templates

bparees:


bparees:
@gabemontero ptal

gabemontero:
lgtm

PR #232, opened 2017-01-16
plugin version bumping

bparees:


bparees:
@gabemontero i'm not sure how jenkins plugin authors keep updating their plugin dependencies w/o updating their plugin version, lately every time i go to rebuild jenkins I discover another plugin that fails to load. (luckily our tests now check for this.  unluckily I merged that other PR w/o waiting for tests to complete).

anyway, fyi.

PR #233, opened 2017-01-19
fix -z option in $JAVA_GC_OPTS condition

h-sano:


h-sano:
fix -z option in $JAVA_GC_OPTS condition

bparees:
@h-sano thanks, looks like there are a few other spots w/ this problem too, can you fix them all?
(run-jnlp-client and the v2 run script)

h-sano:
fix this problem in run-jnlp-client and the v2 run script.

bparees:
[test]

PR #234, opened 2017-01-20
bump durable task plugin version

bparees:


bparees:
@tdawson @gabemontero more plugin dep fun.

PR #238, opened 2017-01-25
add java i386 arch

gabemontero:
@bparees PTAL

*** NOTE *** 

I wanted to see what it would take to use `alternatives` to manage the java version as the `1001` user, since `alternatives` is a more complete switch (handles everything, not just the `java` binary).

Here are the updates for centos and jenkins v1 and v2.

Based on how the review goes, I'll circle back and start looking into updating the rhel image.

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -28,7 +28,7 @@ EXPOSE 8080 50000
 RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
-    INSTALL_PKGS="rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk-devel jenkins-1.651.3-1.1 nss_wrapper" && \
+    INSTALL_PKGS="rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-1.651.3-1.1" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all  && \
@@ -48,6 +48,36 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/subversion.jpi.pinned && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
+    chmod 775 /etc/passwd && \
+    chmod -R 775 /etc/alternatives && \
+    chmod -R 775 /var/lib/alternatives && \
+    chmod -R 775 /usr/lib/jvm && \
+    chmod 775 /usr/bin && \
+    chmod 775 /usr/lib/jvm-exports && \
+    chmod 775 /usr/share/man/man1 && \
+    unlink /usr/bin/java && \
+    unlink /usr/bin/jjs && \
+    unlink /usr/bin/keytool && \
+    unlink /usr/bin/orbd && \
+    unlink /usr/bin/pack200 && \
+    unlink /usr/bin/policytool && \
+    unlink /usr/bin/rmid && \
+    unlink /usr/bin/rmiregistry && \
+    unlink /usr/bin/servertool && \
+    unlink /usr/bin/tnameserv && \
+    unlink /usr/bin/unpack200 && \
+    unlink /usr/lib/jvm-exports/jre && \
+    unlink /usr/share/man/man1/java.1.gz && \
+    unlink /usr/share/man/man1/jjs.1.gz && \
+    unlink /usr/share/man/man1/keytool.1.gz && \
+    unlink /usr/share/man/man1/orbd.1.gz && \
+    unlink /usr/share/man/man1/pack200.1.gz && \
+    unlink /usr/share/man/man1/policytool.1.gz && \
+    unlink /usr/share/man/man1/rmid.1.gz && \
+    unlink /usr/share/man/man1/rmiregistry.1.gz && \
+    unlink /usr/share/man/man1/servertool.1.gz && \
+    unlink /usr/share/man/man1/tnameserv.1.gz && \
+    unlink /usr/share/man/man1/unpack200.1.gz && \
     chown -R 1001:0 /opt/openshift && \
     rmdir /var/log/jenkins && \
     /usr/local/bin/fix-permissions /opt/openshift && \
--- a/1/contrib/s2i/run
+++ b/1/contrib/s2i/run
@@ -9,38 +9,55 @@
 source /usr/local/bin/jenkins-common.sh
 source /usr/local/bin/kube-slave-common.sh
 
+#get the fully qualified paths to both 32 and 64 bit java
+JVMPath32bit=`alternatives --display java | grep family | grep i386 | awk '{print $1}'`
+JVMPath64bit=`alternatives --display java | grep family | grep x86_64 | awk '{print $1}'`
+
+
+# set the java version used based on OPENSHIFT_JENKINS_JVM_ARCH
+if [ -z $OPENSHIFT_JENKINS_JVM_ARCH  ]; then
+    echo "Using 64 bit Java since OPENSHIFT_JENKINS_JVM_ARCH is not set (historic setting)"
+    alternatives --set java $JVMPath64bit
+elif [ "${OPENSHIFT_JENKINS_JVM_ARCH}" == "x86_64"  ]; then
+    echo "64 bit Java explicitly set in OPENSHIFT_JENKINS_JVM_ARCH"
+    alternatives --set java $JVMPath64bit
+else
+    echo "OPENSHIFT_JENKINS_JVM_ARCH is set to ${OPENSHIFT_JENKINS_JVM_ARCH} so using 32 bit Java"
+    alternatives --set java $JVMPath32bit
+fi
+
 image_config_dir="/opt/openshift/configuration"
 image_config_path="${image_config_dir}/config.xml"
-
 
 CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
 DEFAULT_MEMORY_CEILING=$((2**40-1))
 if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
 
     if [ -z $CONTAINER_HEAP_PERCENT ]; then
-        LOW_MEMORY_THRESHOLD=$((640*2**20)) # 640M
-        if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${LOW_MEMORY_THRESHOLD}" ]; then
-            # If the user is not actively tailoring environment variables to adjust memory use,
-            # then tailor the runtime for limited memory automatically.
-            CONTAINER_HEAP_PERCENT=0.40
-            LOW_MEMORY_PARAMS="-Xss500K"
-        else
-            CONTAINER_HEAP_PERCENT=0.50
-        fi
+        CONTAINER_HEAP_PERCENT=0.50
     fi
     
     CONTAINER_MEMORY_IN_MB=$((${CONTAINER_MEMORY_IN_BYTES}/1024**2))
+    #if machine has 4GB or less, meaning max heap of 2GB given current default, force use of 32bit to save space unless user
+    #specifically want to force 64bit
+    HEAP_LIMIT_FOR_32BIT=$((2**32-1))
+    HEAP_LIMIT_FOR_32BIT_IN_MB=$((${HEAP_LIMIT_FOR_32BIT}/1024**2))
     CONTAINER_HEAP_MAX=$(echo "${CONTAINER_MEMORY_IN_MB} ${CONTAINER_HEAP_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+    if [[ -z $OPENSHIFT_JENKINS_JVM_ARCH && "${CONTAINER_HEAP_MAX}" -lt "${HEAP_LIMIT_FOR_32BIT_IN_MB}"  ]]; then
+	echo "max heap in MB is ${CONTAINER_HEAP_MAX} and 64 bit was not explicitly set so using 32 bit Java"
+	alternatives --set java $JVMPath32bit
+    fi
     
     JAVA_MAX_HEAP_PARAM="-Xmx${CONTAINER_HEAP_MAX}m"
 fi 
 
 if [ -z $JAVA_GC_OPTS ]; then
-  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MaxPermSize=100m -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
+  # note - MaxPermSize no longer valid with v8 of the jdk ... used to have -XX:MaxPermSize=100m
+  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
 fi
 
 # Since OpenShift runs this Docker image under random user ID, we have to assign
-# the 'jenkins' user name to this UID. For that we use nss_wrapper
+# the 'jenkins' user name to this UID.
 generate_passwd_file
 
 mkdir /tmp/war
@@ -126,7 +143,7 @@ fi
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_GC_OPTS $JAVA_MAX_HEAP_PARAM $LOW_MEMORY_PARAMS -Duser.home=${HOME} $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
+   exec java $JAVA_GC_OPTS $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
 fi
 
 # As argument is not jenkins, assume user want to run his own process, for sample a `bash` shell to explore this image
--- a/1/test/run
+++ b/1/test/run
@@ -103,11 +103,10 @@ function test_get_job() {
 
 function run_tests() {
   local name=$1 ; shift
-  CONTAINER_ARGS="-u 10000:0"
   create_container $name $VOLUME
   CONTAINER_IP=$(get_container_ip $name)
   test_connection $name
- 
+
   # Make sure no plugins failed to load
   set +e
   CONTAINER=$(cat $cidfile)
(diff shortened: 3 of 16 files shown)


1/Dockerfile
```diff
@@ -49,6 +49,35 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
     chown -R 1001:0 /opt/openshift && \
+    chown -R 1001:0 /etc/alternatives && \
+    chown -R 1001:0 /var/lib/alternatives && \
+    chown -R 1001:0 /usr/lib/jvm && \
```
bparees: what are we doing at assemble time (that's the only thing that runs under uid 1001) that needs ownership of these directories?

1/Dockerfile
```diff
@@ -49,6 +49,35 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
...
+    chmod 775 /usr/bin && \
+    chmod 775 /usr/lib/jvm-exports && \
+    chmod 775 /usr/share/man/man1 && \
```
bparees: what are we doing in these dirs that the existing permissions were problematic?

bparees: (and when are we doing it?  assemble or run time?)

1/Dockerfile
```diff
@@ -49,6 +49,35 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
...
+    unlink /usr/share/man/man1/jjs.1.gz && \
+    unlink /usr/share/man/man1/keytool.1.gz && \
+    unlink /usr/share/man/man1/orbd.1.gz && \
+    unlink /usr/share/man/man1/pack200.1.gz && \
+    unlink /usr/share/man/man1/policytool.1.gz && \
+    unlink /usr/share/man/man1/rmid.1.gz && \
+    unlink /usr/share/man/man1/rmiregistry.1.gz && \
+    unlink /usr/share/man/man1/servertool.1.gz && \
+    unlink /usr/share/man/man1/tnameserv.1.gz && \
+    unlink /usr/share/man/man1/unpack200.1.gz && \
```
bparees: do we need to unlink all these files?  the man pages in particular i'd guess we don't care about.  for the bin directory, does "alternatives" actually setup content in the bin directory?  Could we simply make sure the actual JVM paths appear in $PATH before /usr/bin?

bparees:
I think most of my questions boil down to "exactly what does "alternatives -set java" do?

1/test/run
```diff
@@ -103,7 +103,7 @@ function test_get_job() {
 
 function run_tests() {
   local name=$1 ; shift
-  CONTAINER_ARGS="-u 10000:0"
+  CONTAINER_ARGS="-u 1001:0"
```
bparees: (from your IRC query) we're setting it to 10000 to ensure it can run as a random uid, not the uid we built the image as.

if it only passes when you set the container uid to 1001, then it's going to fail when run on openshift.

gabemontero:
OK, so the `alternatives` command (do a `man alternatives` on say your fedora laptop, or see https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6/html/Installation_Guide/Install_OpenJDK_on_Red_Hat_Enterprise_Linux.html) allows you to manage all the binaries of a given "application" and switch between versions when multiple versions can exist on a system.
Java is such an "application".

Essentially every binary I am unlinking in the Dockerfile is a binary under the Java umbrella that is managed by `alternatives`.

Per the doc noted above, it is assumed you'll be root or using sudo, so all this was the attempt to work around that.

Ultimately, `/usr/bin/java` is linked to `/etc/alternatives/java`, which is then linked to the specific version of java you want (i.e. i386 vs. x86_64, or specifhttps://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6/html/Installation_Guide/Install_OpenJDK_on_Red_Hat_Enterprise_Linux.htmlic versions of either of those two arch's).  The `alternatives` command will replace the links for all the files listed under the Java "family" to the version you specify.

Simply changing the permissions was not sufficient though.  `alternatives` does some `ln` type operations under the covers.  Evidently when running as root, you can existing links.  But you can't `ln` over an existing file when running as non-root, independent of the permission of the file.  So the `ln` portion of `alternatives` would fail.  Unlinking in the Dockerfile prior to running `alternatives` in the run script was the workaround.

As to the random ID, to work around that, we would ditch the `chown` calls and replace them with `chmod 777` to allow the random IDs to manipulate these files.

So the question:  is this extent of permission change the lesser or greater evil, as compared to using JAVA_HOME or fully resolved path to `java` in our various scripts, but having all the other java-related binaries point to the other arch.  My concern with the latter is how assured can we be the Jenkins or some plugin will pull in won't have a dependency on one of those other binaries.

Or we simply just switch to 32 bit and don't provide a dynamic switch.

gabemontero:
OK I've reverted the test run scripts to use the random ID and updated the Dockerfiles to leverage chmod to open things up further to users beyond root or 1001.  `make test` passes locally for me.

bparees:
> So the question: is this extent of permission change the lesser or greater evil, as compared to using JAVA_HOME or fully resolved path to java in our various scripts, but having all the other java-related binaries point to the other arch. My concern with the latter is how assured can we be the Jenkins or some plugin will pull in won't have a dependency on one of those other binaries.

yeah i certainly like using `alternatives` better than relying on java_home (it's never been clear to me to what extent that actually controls the classpath and other aspects of the application runtime)

1/Dockerfile
```diff
@@ -48,6 +48,35 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/subversion.jpi.pinned && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
+    chmod -R 777 /etc/alternatives && \
+    chmod -R 777 /var/lib/alternatives && \
+    chmod -R 777 /usr/lib/jvm && \
+    chmod 777 /usr/bin && \
+    chmod 777 /usr/lib/jvm-exports && \
+    chmod 777 /usr/share/man/man1 && \
```
bparees: chgrp'ing to 0 should also work (as long as the dir has group rwx permissions). 

I don't know that it's actually any more secure at the end of the day, though it feels slightly more secure.

1/contrib/s2i/run
```diff
@@ -9,6 +9,29 @@
 source /usr/local/bin/jenkins-common.sh
 source /usr/local/bin/kube-slave-common.sh
 
+#display the available i386 and x86_64 java's to confirm their presence and version
+alternatives --display java | grep family | grep i386
+alternatives --display java | grep family | grep x86_64
```
bparees: is this just for debugging purposes?  doesn't seem like output we'd likely want in the final image.

1/contrib/s2i/run
```diff
@@ -9,6 +9,29 @@
...
+elif [ "${OPENSHIFT_JENKINS_JVM_ARCH}" == "x86_64"  ]; then
+    echo "64 bit Java explicitly set in OPENSHIFT_JENKINS_JVM_ARCH"
+    alternatives --set java $JVMPath64bit
+else
+    echo "OPENSHIFT_JENKINS_JVM_ARCH is set to something other than x86_64 so using 32 bit Java"
+    alternatives --set java $JVMPath32bit
+fi
+
+#confirm what version of java is being used
+ls -la /etc/alternatives/java
```
bparees: same question here (debugging?)

bparees:
we should add a test that brings up the image using the 32 bit jvm.

1/contrib/s2i/run
gabemontero: Originally debug, but then I left in cause I thought it might be useful when debugging the odd ball customer issue that could be a JDK issue ... channeling my prior existence as an app server developer where knowing the precise JDK version was important :-)

1/contrib/s2i/run
gabemontero: Basic / same response as above  .... could see getting rid of the ones above and just keeping this one for that debug note on which JDK the customer went with

1/contrib/s2i/run
bparees: yeah, I can certainly appreciate the "let's dump all the debug info we want by default" attitude, but i think we should be able to get that info if we really need it relatively easily in this case.

1/Dockerfile
gabemontero: Gotcha ... let me see about getting a stable PR test run with the current approach.  If we get to that point, I'll see about switching up.

1/contrib/s2i/run
bparees: how about dump both, but expand the echo statements above?  "Using JDK $FOO based on OPENSHIFT_JENKINS_JVM_ARCH value of $OPENSHIFT_JENKINS_JVM_ARCH"

gabemontero:
The 32 bit jvm comes up, but an initial foray into a 32 bit jvm test brought up a potential issue.  The password related stuff failed, and I now noticed this in the jenkins log:

```
ERROR: ld.so: object 'libnss_wrapper.so' from LD_PRELOAD cannot be preloaded: ignored.
```

Looks like we'll need to see if a 32 bit version of the nss stuff is out there.

gabemontero:
OK I've pushed an update that 
- adds a 32 bit test
- tweaks the debug along the lines discussed
- installs (albeit via rpmfind.net which may not be a long term solution) 32 bit nss for centos

`make test` passes for me locally

I'll circle back to the `chgrp` comment, then look at 32-bit for `slave-base` and the headless java stuff.

1/Dockerfile
gabemontero: turns out the group only has r-x permissions:

```
drwxr-xr-x.  2 root root 4096 Jan 25 19:01 man1
```

Leaving things as they are wrt chmod, unless I'm misinterpreting your previous comment.

bparees: yeah i meant that we could chmod them to 775 (since they are already in the root group, there is no need to chgrp them)

gabemontero:
On Fri, Jan 27, 2017 at 4:37 PM Ben Parees <notifications@github.com> wrote:

>
> In 1/Dockerfile <https://github.com/openshift/jenkins/pull/238>:
>
> > @@ -48,6 +48,35 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
>      touch /opt/openshift/plugins/subversion.jpi.pinned && \
>      touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
>      touch /opt/openshift/plugins/script-security.jpi.pinned && \
> +    chmod -R 777 /etc/alternatives && \
> +    chmod -R 777 /var/lib/alternatives && \
> +    chmod -R 777 /usr/lib/jvm && \
> +    chmod 777 /usr/bin && \
> +    chmod 777 /usr/lib/jvm-exports && \
> +    chmod 777 /usr/share/man/man1 && \
>
> yeah i meant that we could chmod them to 775 (since they are already in
> the root group, there is no need to chgrp them)
>
ah ok i originally tried 775 and it didn't work with the random 10000 uid

bparees:
> ah ok i originally tried 775 and it didn't work with the random 10000 uid

strange.  the container user is always a member of the root group.

gabemontero:
Pretty sure that was the case, but I did iterate over the solution quickly.  Maybe I misinterpreted something.  I'll retry when I get the chance to make sure.

Also, I just pushed the slave-base changes for 32 bit centos.

1/Dockerfile
```diff
@@ -28,9 +28,12 @@ EXPOSE 8080 50000
 RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
-    INSTALL_PKGS="rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk-devel jenkins-1.651.3-1.1 nss_wrapper" && \
+    INSTALL_PKGS="rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-1.651.3-1.1 nss_wrapper wget" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
+    wget ftp://rpmfind.net/linux/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/n/nss_wrapper-1.1.3-1.fc25.i686.rpm && \
```
bparees: yeah i definitely want to work out a better solution with @tdawson (but it's fine for testing/proving it can work).

1/Dockerfile
gabemontero: OK ... switching to 775 did in fact work for me when running the tests.  Must have misinterpreted something earlier.

Pushed the updates.

1/Dockerfile
gabemontero: I incrementally added each `unlink` until I was able to get the `alternatives` command to work.  It manipulates all of the above file system content.

1/test/run
```diff
@@ -102,12 +102,16 @@ function test_get_job() {
 }
 
 function run_tests() {
-  local name=$1 ; shift
+  local name=$1
   CONTAINER_ARGS="-u 10000:0"
+  if [ $# -gt 1 ];
+  then
+      CONTAINER_ARGS+=" -e OPENSHIFT_JENKINS_JVM_ARCH=i386 "
+  fi
```
bparees: we handle this more generically in other test code:
https://github.com/sclorg/mongodb-container/blob/master/3.2/test/run#L163-L167
https://github.com/sclorg/mongodb-container/blob/master/3.2/test/run#L291-L299

gabemontero: Didn't rework to exactly like the mongo run's (would have lead to a much broader structural change) but I reworked it to something I think is similar.  Changes pushed.

bparees:
looks ok to me, but i still want us to pursue the centos nss wrapper rpm before merging.

gabemontero:
Absolutely - I've also started rhel changes.  Or do we only want to merge
centos initially?

bparees:
starting rhel makes sense.

gabemontero:
OK, I've pushed the rhel changes after some verification on a local AWS rhel instance I have.

@tdawson - if you recall the 32 bit nss_wrapper rpms for centos7 and rhel7 discussion you @bparees and I had on IRC a few days ago, do you have a preference on how we "officially" engage with you to get those RPMs available on a repo for our jenkins build to access on your plate?  Or is it too soon to "officially" engage :-) ?

thanks

gabemontero:
Hmm ... the use of `alternatives` failed in the PR test for rhel even though it passed for me in my local aws instance.

Going to temporarily add some debug and see what is up.

gabemontero:
Also taking out SKIP_SQUASH based on our recent email thread.

gabemontero:
Ah .. did not pull all the changes from my AWS instance over to my local repo (since I can't get vagrant-openshift sync for the jenkins repo to work yet :-( )

Pushing fix in a sec.

gabemontero:
OK, now pushed update to dynamically set to 32bit if mem is less that 4GB and the env var is not explicitly set to use 64 bit.  PTAL @bparees

1/contrib/s2i/run
```diff
@@ -9,11 +9,36 @@
...
+JVMPath64bit=`alternatives --display java | grep family | grep x86_64 | awk '{print $1}'`
+
+#if machine has 4GB or less, going to force use of 32bit to save space unless user
+#specifically want to force 64bit
+CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+MEMORY_LIMIT_FOR_32BIT=$((2**32-1))
+
+# set the java version used based on OPENSHIFT_JENKINS_JVM_ARCH
+if [ -z $OPENSHIFT_JENKINS_JVM_ARCH  ]; then
+    if [ "${CONTAINER_MEMORY_IN_BYTES}" -gt "${MEMORY_LIMIT_FOR_32BIT}" ]; then
```
bparees: unfortunately it's more complicated than this.

What the check really needs to be based on is whether the computed heap size is greater than what a 32 bit jvm allows.  Depending on CONTAINER_HEAP_PERCENT, the computed heap could be significantly less than the container memory.

also what is the actual max value a 32 bit jvm on linux allows the heap to be set to?  is it 2**32-1?  i would think it might be a bit lower than that since some process memory needs to be reserved for native usage.

bparees: btw i'm perfectly ok w/ not doing this magic.

gabemontero:
Ok I can see about adding more sophistication.  Though I wonder if it
really needs to be that precise.  If the user is manipulating the heap with
the other props and hitting edge cases he can just add the prop to
explicitly turn on 32 bit.

But yeah if it becomes a rabbit hole we can just punt.

gabemontero:
Pushed the changes for the 32 vs. 64 bit auto tuning per our discussion in scrum this AM.

gabemontero:
OK pushed the changes to remove nss_wrapper as a dependency per discussion in scrum this AM.

Assuming this holds we won't need our previous ask of @tdawson regarding getting 32 bit nss_wrapper available on our repos during image build.

hack/build.sh
```diff
@@ -27,9 +27,6 @@ function docker_build_with_version {
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-version=\"${git_version}\"" >> "${dockerfile}.version"
   docker build -t ${IMAGE_NAME} -f "${dockerfile}.version" .
-  if [[ "${SKIP_SQUASH}" != "1" ]]; then
-    squash "${dockerfile}.version"
```
bparees: you can remove the actual squash function too.

1/Dockerfile
```diff
@@ -48,6 +48,36 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/subversion.jpi.pinned && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
+    chmod 775 /etc/passwd && \
```
bparees: i'm having second thoughts about this since it means anyone who can run a job on jenkins can update the passwd file which in turn means they can have free run of the jenkins container, including changing the jenkins config to give themselves more access.  (for example I think they could add/update an existing user with a different password and then su to that user)

it would be better if we can "close the door behind us" after we update the file on startup but i don't think that's going to be possible.

going to have to think about this one some more.

hack/build.sh
gabemontero: update pushed

tdawson:
Thanks for removing the nss_wrapper dependency.  Every scenario I could think of was rather messy.

1/Dockerfile
gabemontero: @bparees and I had a confab with Clayton ... that coupled with some experimentation leads us to believe we are OK, but are still going to sit and ponder for a bit since this PR is post 3.5.

1/test/run
```diff
@@ -192,6 +191,16 @@ function run_s2i_tests() {
 # Tests.
 
 # Normal tests
+echo "Normal test: 64 bit"
+CONTAINER_ARGS="-u 10000:0"
```
bparees: should set the JVM explicitly to 64bit, because if we (somehow) run this test on a machine w/ low memory it's going to go through the 32 bit path.

2/test/run
```diff
@@ -192,6 +191,16 @@ function run_s2i_tests() {
 # Tests.
 
 # Normal tests
+echo "Normal test: 64 bit"
+CONTAINER_ARGS="-u 10000:0"
```
bparees: same thing here

bparees:
@gabemontero one last test related nit and then lgtm.  We'll hold this until the 3.5 jenkins image goes out the door though.

1/test/run
gabemontero: good catch ... part of next push

gabemontero:
Updates to test pushed.

btw, more scm-api plugin churn seen during the v2 test:

```
SEVERE: Failed Loading plugin GitHub Branch Source Plugin v2.0.3 (github-branch-source)
java.io.IOException: GitHub Branch Source Plugin v2.0.3 failed to load.
 - SCM API Plugin v2.0.3 is older than required. To fix, install v2.0.4 or later.
```

I'll create a separate PR, then rebase this PR after that PR merges.

gabemontero:
ok test passing again after plugin update rebase

gabemontero:
@bparees  bump ... are we close enough to 3.5 GA to update the centos image and have @tdawson cook up a rhel7 image on brew for QA to use?  If not there yet, are we close enough that a precise target date is known?

thanks

bparees:
@gabemontero probably the end of this sprint, which is when we hit codefreeze for 3.5.

gabemontero:
On Tue, Feb 21, 2017 at 11:41 AM Ben Parees <notifications@github.com>
wrote:

> @gabemontero <https://github.com/gabemontero> probably the end of this
> sprint, which is when we hit codefreeze for 3.5.
>
makes sense ok thx for update

PR #239, opened 2017-01-27
backlevel the scm-api plugin version

bparees:


bparees:
@gabemontero @tdawson I tested building w/ the 1.3 plugin and everything seems to come up (no plugin load errors) so let's go with this for 3.5.

PR #241, opened 2017-02-01
README: fix script path in layered example

csrwng:
example has wrong path to for install-plugins script


csrwng:
@bparees ptal

bparees:
lgtm

PR #242, opened 2017-02-06
Change the default Kubernetes cloud name to 'kubernetes' 

peterzhao:
The old name 'openshift' of the kubernetes cloud is not recognized by the Kubernetes plugin. When a job with a node label 'maven' or 'nodejs' is run, there will be an error from the Kubernetes plugin saying cannot find the cloud 'kubernetes'.  After rename the cloud to 'kunbernetes' this issue is gone. Tested with Jenkins 2.7.3 and Kubernetes plugin 0.10.

### DIFF
--- a/1/contrib/jenkins/kube-slave-common.sh
+++ b/1/contrib/jenkins/kube-slave-common.sh
@@ -126,7 +126,7 @@ function generate_kubernetes_config() {
     local crt_contents=$(cat "${KUBE_CA}")
     echo "
     <org.csanchez.jenkins.plugins.kubernetes.KubernetesCloud>
-      <name>openshift</name>
+      <name>kubernetes</name>
       <templates>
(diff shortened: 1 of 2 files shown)


bparees:
@peterzhao we haven't experienced any issue w/ the existing name and given that the plugin itself is allowing the user to specify the name, it seems unlikely that the name is forced to any particular value.  Can you provide config files and logs from when you hit the issue?

You mention jenkins 2.7.3, but our image is on jenkins 2.19 currently, did you see the issue on that version?

peterzhao:
Hi Ben,

We are using a derived image from openshift/jenkins-2-centos7, which has a
different version of Jenkins. I did not test this issue on jenkins 2.19.
I tested this issue again this morning but cannot reproduce it anymore. I
think it probably has been fixed through a newer base image.
Anyway since this issue is gone, please discard the pull request.

Thanks

Peter

PR #243, opened 2017-02-06
bump to LTS 2.32

bparees:


bparees:
@tdawson fyi.  looks like the scm-api 2.0.2 plugin is available now too.

bparees:
@tdawson turns out even more plugins needed to be updated.

tdawson:
rpms made.  Yum repos/puddles updated and pushed to mirrors.
Images might not get done today.

PR #244, opened 2017-02-07
allow for permission to see, easy enablement of, access logs

gabemontero:
@bparees ptal

@jupierce fyi

### DIFF
--- a/1/Dockerfile
+++ b/1/Dockerfile
@@ -49,8 +49,10 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
     chown -R 1001:0 /opt/openshift && \
+    rmdir /var/log/jenkins && \
     /usr/local/bin/fix-permissions /opt/openshift && \
-    /usr/local/bin/fix-permissions /var/lib/jenkins
+    /usr/local/bin/fix-permissions /var/lib/jenkins && \
+    /usr/local/bin/fix-permissions /var/log
 
 VOLUME ["/var/lib/jenkins"]
 
--- a/1/contrib/s2i/run
+++ b/1/contrib/s2i/run
@@ -50,6 +50,9 @@ if [ -e ${JENKINS_HOME}/password ]; then
 fi
 new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
 
+# finish the move of the default logs dir, /var/log/jenkins, to the volume mount
+mkdir ${JENKINS_HOME}/logs
+ln -sf ${JENKINS_HOME}/logs /var/log/jenkins
 
 if [ ! -e ${JENKINS_HOME}/configured ]; then
   # This container hasn't been configured yet, so copy the default configuration from
@@ -116,9 +119,14 @@ fi
 
 rm -rf /tmp/war
 
+# default log rotation in /etc/logrotate.d/jenkins handles /var/log/jenkins/access_log
+if [ ! -z "${OPENSHIFT_USE_ACCESS_LOG}" ]; then
+    JENKINS_ACCESSLOG="--accessLoggerClassName=winstone.accesslog.SimpleAccessLogger --simpleAccessLogger.format=combined --simpleAccessLogger.file=/var/log/jenkins/access_log"
+fi
+
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_GC_OPTS $JAVA_MAX_HEAP_PARAM $LOW_MEMORY_PARAMS -Duser.home=${HOME} $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
+   exec java $JAVA_GC_OPTS $JAVA_MAX_HEAP_PARAM $LOW_MEMORY_PARAMS -Duser.home=${HOME} $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
 fi
 
 # As argument is not jenkins, assume user want to run his own process, for sample a `bash` shell to explore this image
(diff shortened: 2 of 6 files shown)


1/Dockerfile
```diff
@@ -48,9 +48,11 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/subversion.jpi.pinned && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
+    chmod 775 /var/log/jenkins && \
```
bparees: seems like logs should go into a volume dir, can we change the jenkins config to put the logs in /var/lib/jenkins/logs or something?

bparees: (or symlink /var/log/jenkins into a /var/lib/jenkins/logs)

1/Dockerfile
```diff
@@ -48,9 +48,11 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
...
     chown -R 1001:0 /opt/openshift && \
     /usr/local/bin/fix-permissions /opt/openshift && \
-    /usr/local/bin/fix-permissions /var/lib/jenkins
+    /usr/local/bin/fix-permissions /var/lib/jenkins && \
+    /usr/local/bin/fix-permissions /var/log/jenkins
```
bparees: fix-permissions already sets group rwx, so the 775 above seems redundant.

1/Dockerfile
gabemontero: I'll have to figure out how `/etc/logrotate.d/jenkins` is getting created and based on that see if we can modify.  Assuming we can say manipulate it in the dockerfiles, then yes I think we can make this change.

1/Dockerfile
gabemontero: Yeah I was wondering about that and was going to look/try and then forgot.  I'll confirm / change.

1/Dockerfile
gabemontero: update pushed

1/Dockerfile
gabemontero: update pushed

1/Dockerfile
```diff
@@ -49,6 +49,9 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
     touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
     touch /opt/openshift/plugins/script-security.jpi.pinned && \
     chown -R 1001:0 /opt/openshift && \
+    rmdir /var/log/jenkins && \
+    mkdir -p /var/lib/jenkins/logs && \
+    ln -sf /var/lib/jenkins/logs /var/log/jenkins && \
```
bparees: i think we can't actually create the link here, because when you actually mount an empty volume into /var/lib/jenkins it's going to shadow the logs directory.

the symlink needs to be created in the same place where we populate the rest of /var/lib/jenkins (in the run script, during startup)

gabemontero:
On Tue, Feb 7, 2017 at 12:57 PM Ben Parees <notifications@github.com> wrote:

>
> In 1/Dockerfile
> <https://github.com/openshift/jenkins/pull/244#pullrequestreview-20574809>
> :
>
> > @@ -49,6 +49,9 @@ RUN /usr/local/bin/plugins.sh /opt/openshift/base-plugins.txt && \
>      touch /opt/openshift/plugins/ssh-credentials.jpi.pinned && \
>      touch /opt/openshift/plugins/script-security.jpi.pinned && \
>      chown -R 1001:0 /opt/openshift && \
> +    rmdir /var/log/jenkins && \
> +    mkdir -p /var/lib/jenkins/logs && \
> +    ln -sf /var/lib/jenkins/logs /var/log/jenkins && \
>
> i think we can't actually create the link here, because when you actually
> mount an empty volume into /var/lib/jenkins it's going to shadow the logs
> directory.
>
> the symlink needs to be created in the same place where we populate the
> rest of /var/lib/jenkins (in the run script, during startup)
>
Interesting.  Ok I will leave the rmdir/mkdir here and move the ln.  Thanks.

bparees:
> Interesting.  Ok I will leave the rmdir/mkdir here and move the ln.  Thanks.

Sorry, you can't make the /var/lib/jenkins/logs dir here either.  It'll still be shadowed by the volume mount.

gabemontero:
OK ... pushed updates ... simply remove /var/log/jenkins, then make sure the random uid has permission to link in /var/log, in the Dockerfiles.

Dir creation and linking done in run scripts.

1/contrib/s2i/run
```diff
@@ -50,6 +50,9 @@ if [ -e ${JENKINS_HOME}/password ]; then
 fi
 new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
 
+# finish the move of the defaul logs dir, /var/log/jenkins, to the volume mount
```
bparees: s/defaul/default/

2/Dockerfile.rhel7
```diff
@@ -55,6 +55,7 @@ RUN rm /opt/openshift/base-plugins.txt && \
     # need to create <plugin>.pinned files when upgrading "core" plugins like credentials or subversion that are bundled with the jenkins server
     # Currently jenkins v2 does not embed any plugins, but for reference:
     # touch /opt/openshift/plugins/credentials.jpi.pinned && \
+    rmdir /var/log/jenkins && \
```
bparees: doesn't this one also need fix-permissions on /var/log?

2/contrib/s2i/run
```diff
@@ -50,6 +50,9 @@ if [ -e ${JENKINS_HOME}/password ]; then
 fi
 new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
 
+# finish the move of the defaul logs dir, /var/log/jenkins, to the volume mount
```
bparees: defaul->default here too.

1/contrib/s2i/run
gabemontero: update pushed

2/Dockerfile.rhel7
gabemontero: yep update pushed

2/contrib/s2i/run
gabemontero: update pushed

bparees:
lgtm, have you been able to test this in a cluster w/ a volume?

gabemontero:
Cool - no, I've only tested with jenkins-ephemeral

bparees:
ephemeral should be fine, there's still a volume in play in that scenario (an emptydir volume)

gabemontero:
ah - yep, that's true re: ephemeral

ok i'll hit the merge button if the test passes

PR #247, opened 2017-02-14
Bump BlueOcean to version 1.0.0-b23

csrwng:
Current version of BlueOcean is broken


csrwng:
@bparees ptal

bparees:
lgtm pending test pass

PR #248, opened 2017-02-15
more jenkins plugin churn (scmapi, jquery-detached)

gabemontero:
@bparees  - more scm-api churn on v2

also, jquery-detached failed to download in v2 testing.  Uncovered that
a) https://updates.jenkins-ci.org/download/plugins/jquery-detached-plugin now gets a 404 / not found
b) per https://wiki.jenkins-ci.org/display/JENKINS/jQuery+Detached+Plugin you should not install this plugin directory

Interestingly, even after clearing my docker image cache, jquery-detached still downloaded and built with v1.  At first blush don't understand why.  We'll see how subsequent tests turn out.

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -32,14 +32,13 @@ cloudbees-folder:5.17
 
 # Pipeline stage view plugin - https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Stage+View+Plugin
 pipeline-stage-view:2.2
-jquery-detached:1.2.1
 handlebars:1.1.1
 pipeline-rest-api:2.2
 momentjs:1.1.1
 
 # remote loader https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Remote+Loader+Plugin
 workflow-remote-loader:1.2
-scm-api:2.0.3
+scm-api:2.0.4
 branch-api:2.0.2
 
 # Pipeline SCM Step Plugin - https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+SCM+Step+Plugin


bparees:
Hmm. Cesar hit and fixed the scm one earlier today (if you rebase you should see it), but the jquery one didn't come up.

bparees:
Sorry, nm, the one Cesar hit was a different plugin. This stuff is certainly churning fast.

bparees:
i find it odd that this isn't failing v1 testing since it installs the same set of plugins:
https://github.com/openshift/jenkins/blob/master/1/contrib/openshift/base-plugins.txt#L35

1) it would be good to understand why the v1 test didn't fail in this PR
2) assuming it should be removed from v1 also, we should do so.

gabemontero:
Yep I echo'ed the same curiosity / concern in the description.  As you can see though, the tests passed and I see jquery-detached referenced in the v1 image output.

My first suspicion on test pass is it is muddled in either the pinned / embedded (i.e. what is included in the respective jenkins.war files) differences between v1 and v2 or the different plugin install script you put in for v2 vs. v1 (https://github.com/openshift/jenkins/blob/master/1/contrib/jenkins/plugins.sh vs. https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/install-plugins.sh)

I tried removing it and while the image builds, it fails when you start it up, citing the missing dependency.  So we need to leave it in for now.

gabemontero:
there is a `set +e` just prior to the wait for downloads to complete section in the v1 plugins.sh .... I'm doing some local runs to examine some more.

bparees:
@gabemontero there are some differences between the plugin download script in the v1 image and the one in the v2 image, i'm guessing that is the reason why one can download it and the other gets a 404.

gabemontero:
Need to circle back to the v2 image, but the v1 image plugin.sh is actually curling against https://updates.jenkins-ci.org/download/plugins/jquery-detached instead of https://updates.jenkins-ci.org/download/plugins/jquery-detached-plugin and that still works.  I can bring the former URL in my browser and it works.

Given the "support plugin" verbiage on the wiki for jquery-detached, perhaps there has been a recent, subtle change in the file paths for updates.jenkins-ci.org ?

I'll report back once I've debugged the v2 build.

gabemontero:
Well now the V2 download of jquery-detached is working as well, with the correct URL !?!?!

Best guess is that there must have been a http hiccup at home when I ran the test last night, and then I botched the PR when testing it in my brownser (though I though I copied/pasted).  Chalk it up to working later in the evening and missing something I suppose.

I still am sure based on the logs today that the dependency downloads work differently / more enhanced in v2 jenkins, and the image still works when we don't explicitly down, where I see mention of that dependency getting pulled down:

```
 > workflow-cps depends on workflow-api:2.6,workflow-scm-step:1.15,workflow-step-api:2.5,workflow-support:2.11,scm-api:1.1,script-security:1.24,structs:1.5,support-core:2.32;resolution:=optional,ace-editor:1.0.1,jquery-detached:1.2.1
```

And I do see that file there when I bash into a image built where that plugin is not in base-plugins.txt:

```
gmontero ~/go/src/github.com/openshift/jenkins  (bumpSCMAPI)$ docker images | grep jenkins
openshift/jenkins-2-centos7              latest              acf4a778f72f        About a minute ago   718.6 MB
gmontero ~/go/src/github.com/openshift/jenkins  (bumpSCMAPI)$ docker run -it acf4a778f72f /bin/bash
bash-4.2$ 

bash-4.2$ 

bash-4.2$ ls -la /opt/openshift/plugins/jquery*
ls -la /opt/openshift/plugins/jquery*
-rw-rw-r--. 1 1001 root 211945 Feb 15 16:26 /opt/openshift/plugins/jquery-detached.jpi
bash-4.2$ 
```
As such, I say we leave it removed, and let Jenkins manage the dependency, vs. us having to update jqeury-detached's version in base-plugin.txt when it changes.

bparees:
> As such, I say we leave it removed, and let Jenkins manage the dependency, vs. us having to update jqeury-detached's version in base-plugin.txt when it changes.

agreed, if we don't have to call it out explicitly, let's not.

PR #249, opened 2017-02-15
Bump Pipeline Shared Groovy Libraries Plugin to version 2.6

csrwng:
Version 2.5 of the plugin breaks the openshift client plugin
Requires change from https://github.com/openshift/jenkins/pull/248


csrwng:
@gabemontero @bparees ptal

bparees:
lgtm but let's let @gabemontero finish his investigation on https://github.com/openshift/jenkins/pull/248 before we merge this jquery change.

PR #254, opened 2017-03-02
bump sync plugin to 0.1.10

gabemontero:
@bparees FYI

@tdawson - can we use this PR to initiate the generation of a new RPM for the openshift-sync-plugin?  If not of course just let me know how you want us to register the ask.

However initiated, once that new RPM is ready, and after we merge https://github.com/openshift/jenkins/pull/238 tomorrow (march 3) we'll want to also have a new set of jenkins images created on brew-pulp to facilitate QE's testing of few features we have ready (and with 3.5 code freeze tomorrow march 3, we feel safe in doing so now)

thanks

PR #257, opened 2017-03-10
Bump sync

gabemontero:
... as well as the next round of upstream jenkins plugin churn

@tdawson - in addition to the bug fix for openshift-sync there I mentioned [here](https://github.com/openshift/jenkins/pull/254#issuecomment-285424673), the now expected churn in upstream jenkins plugins for our jenkins 2 image occurred.  We'll need RPMs for those built in addition to the new openshift-sync RPM as part of building the 3.6 rhel jenkins images.  Thanks.

@bparees fyi

@tdawson @bparees As a side note, upstream jenkins appears to have started making changes wrt how plugn dependencies are loaded...let's not hold off on the items I noted above, but I'm going to investigate whether we can reduce the list of plugins we explicitly download / install.

### DIFF
--- a/1/contrib/openshift/base-plugins.txt
+++ b/1/contrib/openshift/base-plugins.txt
@@ -9,7 +9,7 @@ kubernetes:0.10
 credentials:2.1.9
 
 # fabric8 openshift sync
-openshift-sync:0.1.10
+openshift-sync:0.1.11
 
 # Pipeline plugin - https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Plugin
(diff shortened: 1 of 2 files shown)


gabemontero:
test passed ... merging

gabemontero:
btw, early results on my reduction to 2/contrib/openshift/base-plugins.txt are encouraging... if unit testing and extended testing looks good, I'll craft a PR (and hopefully we'll have less RPMs for @tdawson to deal with and less jenkins upstream plugin churn for @bparees and I to deal with)

PR #258, opened 2017-03-10
reduce v2 plugin list, list parent plugins download

gabemontero:
@tdawson @bparees the origin extended tests and jenkins v2 images test both look good.  I was also able to confirm from the make logs that the plugins removed from the list that we care about were in fact still downloaded (subversion wasn't but I was wondering if we really need that one)

When all is said and down then, I suspect the larger list for v1 jenkins will remain constant for plugins we don't own, while the smaller list of plugins we don't own for v2 can still change.  

Also, for @bparees and I, I believe this will shield from the scenario where a plugin changes dependency versions but the plugin itself does not change its version.  Since we are only including the "parent" plugins now.

I won't change the rhel Dockerfile until a) this PR merges, and b) we confirm with @tdawson that we have new RPMs for the remaining plugins (I *think* some of them have still changed with the recent PRs)

Even if the test passes, I won't merge until @bparees gets back from PTO and has a chance to digest this (and perhaps argues on inclusion of subversion)

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -4,48 +4,17 @@ openshift-client:0.9.2
 
 
 # kubernetes plugin - https://wiki.jenkins-ci.org/display/JENKINS/Kubernetes+Plugin
-durable-task:1.13
 kubernetes:0.10
-credentials:2.1.13
 
 # fabric8 openshift sync
 openshift-sync:0.1.11
 
 # Pipeline plugin - https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Plugin
-pipeline-build-step:2.1
-pipeline-graph-analysis:1.3
-pipeline-input-step:2.5
-pipeline-stage-step:2.2
 workflow-aggregator:2.1
-workflow-api:2.12
-workflow-basic-steps:2.3
-workflow-cps:2.29
-workflow-cps-global-lib:2.7
-workflow-durable-task-step:2.9
-workflow-job:2.10
-workflow-step-api:2.9
-workflow-support:2.13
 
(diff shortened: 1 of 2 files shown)


bparees:
@gabemontero my main concern with this is that if a "parent" plugin changes in the future to no longer require a particular dependent plugin, our image will suddenly stop including it, potentially breaking people's configurations.

Obviously we're already exposed to that since we aren't declaring every plugin that ultimately gets installed, but seeing some of the git and SCM plugins being removed from this list does leave me worried.  So I guess i do wonder if there is a more surgical approach we should take regarding which plugins we explicitly provide and which we implicitly pick up via dependencies.

Of course the rhel based image wouldn't get broken since in that case all the plugins are expliciltly installed by rpm.

gabemontero:
Especially if we stayed more aggressive with reducing what we explicitly install, I could add to our tests and confirm that certain plugins are still installed.  We would at least be more assured that we would find out quickly that certain plugins were in fact missing.

Otherwise, as far as the *churn* goes, while we have had to bump the git related plugins on occasion, the workflow/pipeline plugins have been the greater offenders.  Conceivably in the spirit of being more surgical as you say, we could initially limit the implicit pickup to those, and hopefully only have to occasionally deal with the churn for git, etc.

bparees:
> Especially if we stayed more aggressive with reducing what we explicitly install, I could add to our tests and confirm that certain plugins are still installed. 

I like that approach.

gabemontero:
OK I added subversion/mapdb back in and then added the tests that confirm that non-pipeline plugins that we used to explicitly pull in are still pulled in by pipeline aggregator/parent plugins.

Don't test for workflow/pipeline under the premise that if they change the relevant list for that umbrella, then we defer to them.

bparees:
lgtm

PR #259, opened 2017-03-14
Update Pipeline Groovy libraries

jmcshane:
I tried building this image from a dockerfile into OpenShift with the latest Jenkins image and received an error that the pipeline groovy libraries needed to be updated to 2.7.


gabemontero:
Yeah this has been a common occurrence recently with upstream jenkins churn.  We may ultimately address it with https://github.com/openshift/jenkins/pull/258, but we can merge this if it passes our tests in the interim.

[test]

PR #263, opened 2017-03-28
bump sync and blueocean plugins

gabemontero:
@openshift/devex fyi

@spadgett @jwforres this will lead to the centos image on docker hub having the changes for the new log url annotations


gabemontero:
@tdawson fyi as well ... new versions of openshift-sync and blueocean

PR #264, opened 2017-03-29
Added support of MAVEN_MIRROR_URL to the maven slave image

siamaksade:


bparees:
lgtm but we need to understand the issue you're having w/ the home directory/default user not getting set for you.  how were you testing?

siamaksade:
It is solved now by 8ae0cec976ac0c5827ae298d74c9a2e1491327ac. I built the image locally and configured Kubernetes Jenkins plugin running on my local cluster (oc cluster) to refer to that image.

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -4,6 +4,14 @@
 # will use for the auto-discovery of this slave.
 #
 
+# The directory that Jenkins will execute the builds and store cache files.
+# The directory has to be writeable for the user that the container is running
+# under.
+export JENKINS_HOME=/home/jenkins
+
+# Make sure the Java clients have valid $HOME directory set
+export HOME=${JENKINS_HOME}
```
bparees: i don't understand why this should be necessary:
https://github.com/openshift/jenkins/blob/master/slave-base/Dockerfile#L5

PR #265, opened 2017-03-30
switch osbs version to align with openshift versions

bparees:


bparees:
@tdawson @gabemontero ptal

PR #266, opened 2017-03-31
update maven slave to maven v3.3

bparees:


ALRubinger:
Subscribe :)

bparees:
notes to self:
```
/usr/local/bin/generate_container_user: line 7: /etc/passwd: Permission denied
container_linux.go:247: starting container process caused "exec: \"mvn\": executable file not found in $PATH"
/usr/bin/docker-current: Error response from daemon: invalid header field value "oci runtime error: container_linux.go:247: starting container process caused \"exec: \\\"mvn\\\": executable file not found in $PATH\"\n".
```

bparees:
@gabemontero ptal.  Some of this is changes to cleanup the 32/64bit jdk enablement, looks like we weren't setting the permissions correctly.  guess it got missed when we added 32bit support to the slave images?

also note that i had to hackily uninstall the 1.7 jdk that gets dragged in w/ the maven rpms.

gabemontero:
Ah - yeah in hindsight it makes that similar permission changes would be needed for the maven image.  On brief review, it seems like our existing maven test does not exercise the run-jnlp-client script (which I did change) to the point where we would have seen a error ... is that what you see?

On the change itself, I would expect the dirs to change would match what we did in the base image.  Looks like you did that from what I can tell.

And noted on the 1.7 jdk.

bparees:
> On brief review, it seems like our existing maven test does not exercise the run-jnlp-client script (which I did change) to the point where we would have seen a error ... is that what you see?

yeah our tests are pretty minimal on the slave images.

there's definitely room to add some more extensive tests to the slave-base test script to have it actually exercise the 2 JDKs + heap settings but i don't think i'm going to tackle that in this PR.

bparees:
@stevekuznetsov @csrwng something is really bad in our amis now, we can't even build all the jenkins images:

```
devmapper: Thin Pool has 2800 free data blocks which is less than minimum required 3035 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior
```

SlavoKrupa:
So which version of maven will be available in maven slaves? I believe we need 3.3.9 for [obsidian](https://github.com/openshiftio/booster-parent/blob/master/pom.xml#L17) because `vertx-maven-plugin` requires this version.

stevekuznetsov:
@bparees @csrwng that `jenkins` job that failed due to the `devicemapper` issue ... it's the same AMi we had before, with the same setup... no clue

csrwng:
taking a look

bparees:
> So which version of maven will be available in maven slaves? I believe we need 3.3.9 for obsidian because vertx-maven-plugin requires this version.

@SlavoKrupa it will be 3.3.9, that's what the SCL package contains.

bparees:
[test]

bparees:
[test]

dnguyen9:
@bparees  - Ben what was the rationalize to unlink a lot of the packages from /usr/bin. I'm just more curious than anything. 

https://github.com/openshift/jenkins/commit/5dbd8cfda01f6943257ef3768c7cef03cc251e19

bparees:
@dnguyen9 my understanding is it was necessarily to allow "alternatives" (used to switch between the 64 and 32 bit jdks) to be run as non-root.

@gabemontero can provide more background.

gabemontero:
@bparees is correct - that is exactly where it stemmed from; when the non-root user executes https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/run as part of jenkins startup those unlinked paths get populated

dnguyen9:
Thanks @bparees @gabemontero, understood now. Funny enough, I actually encountered a minor bug in one of the method used in the jnlp-run script. Will raise it as a separate issue

PR #268, opened 2017-04-03
Optionally install plugins on startup

csrwng:
Add support for an environment variable (INSTALL_PLUGINS) that allows
installing a set of plugins on image startup. The format of the variable
is a set of comma-separated strings of the form [plugin-id]:[version].
Example:
INSTALL_PLUGINS=groovy:1.30,github-pr-comment-build:1.1

When used with persistent storage, plugin download/installation would
only happen the very first time the image is run.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -103,6 +103,25 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
   cp -r /opt/openshift/configuration/* ${JENKINS_HOME}
   rm -rf /opt/openshift/configuration/*
 
+
+  # If the INSTALL_PLUGINS variable is populated, then attempt to install
+  # those plugins before copying them over to JENKINS_HOME
+  # The format of the INSTALL_PLUGINS variable is a comma-separated list
+  # of pluginId:pluginVersion strings
+  if [[ -n "${INSTALL_PLUGINS:-}" ]]; then
+    echo "Installing additional plugins: ${INSTALL_PLUGINS} ..."
+
+    # Create a temporary file in the format of plugins.txt
+    plugins_file=$(mktemp)
+    IFS=',' read -ra plugins <<< "${INSTALL_PLUGINS}"
+    for plugin in "${plugins[@]}"; do
+        echo "${plugin}" >> "${plugins_file}"
+    done
+
+    # Call install plugins with the temporary file
+    /usr/local/bin/install-plugins.sh "${plugins_file}"
+  fi
+
   if [ "$(ls -A /opt/openshift/plugins 2>/dev/null)" ]; then
     mkdir -p ${JENKINS_HOME}/plugins
     echo "Copying $(ls /opt/openshift/plugins | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
--- a/2/test/run
+++ b/2/test/run
@@ -201,6 +201,20 @@ function run_s2i_tests() {
     return 1
   fi
 
+  if [[ -n "{VERIFY_INSTALLED:-}" ]]; then
+	  if ! docker exec ${cid} ls /var/lib/jenkins/plugins/${VERIFY_INSTALLED}.jpi; then
+		  echo "Fail: the ${VERIFY_INSTALLED} plugin was not installed."
+		  return 1
+      fi
+  fi
+
+  if [[ -n "{VERIFY_NOT_INSTALLED:-}" ]]; then
+	  if docker exec ${cid} ls /var/lib/jenkins/plugins/${VERIFY_NOT_INSTALLED}.jpi; then
+		  echo "Fail: the ${VERIFY_NOT_INSTALLED} plugin was installed."
+		  return 1
+      fi
+  fi
+
   test_get_job admin password sample-app-test
   test_create_job admin password testJob
   test_get_job admin password testJob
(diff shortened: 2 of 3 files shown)


csrwng:
@bparees @gabemontero ptal
/cc @stevekuznetsov

2/contrib/s2i/run
```diff
@@ -103,6 +103,25 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
...
+
+	# Create a temporary file in the format of plugins.txt
+	plugins_file=$(mktemp)
+	IFS=',' read -ra plugins <<< "${INSTALL_PLUGINS}"
+	for plugin in "${plugins[@]}"; do
+		echo "${plugin}" >> "${plugins_file}"
+	done
+
+	# Call install plugins with the temporary file
+    /usr/local/bin/install-plugins.sh "${plugins_file}"
```
bparees: alignment.

csrwng:
readme updated and alignment fixed

bparees:
can we get a test for this behavior?

csrwng:
@bparees in origin extended tests ?

bparees:
> @bparees in origin extended tests ?

no, in here: https://github.com/openshift/jenkins/blob/master/2/test/run

2/contrib/s2i/run
```diff
@@ -103,6 +103,25 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
...
+  # If the INSTALL_PLUGINS variable is populated, then attempt to install
+  # those plugins before copying them over to JENKINS_HOME
+  # The format of the INSTALL_PLUGINS variable is a comma-separated list
+  # of pluginId:pluginVersion strings
+  if [[ ! -z ${INSTALL_PLUGINS:-} ]]; then
+    echo "Installing additional plugins: ${INSTALL_PLUGINS} ..."
+
+    # Create a temporary file in the format of plugins.txt
+    plugins_file=$(mktemp)
+    IFS=',' read -ra plugins <<< "${INSTALL_PLUGINS}"
```
stevekuznetsov: Don't leak `$IFS`:


```sh
old_ifs="${IFS}"
IFS=','
for ...
end
IFS="${old_ifs}"
```

stevekuznetsov: No need to `read` here, `for` will honor `IFS`:

```sh
IFS=','
for plugin in ${INSTALL_PLUGINS}; do

done
```

2/test/run
```diff
@@ -201,6 +201,12 @@ function run_s2i_tests() {
     return 1
   fi
 
+  if [[ ! -z "{VALIDATE_PLUGIN:-}" ]]; then
```
stevekuznetsov:
```sh
if [[ -n "${VALIDATE_PLUGIN:-}" ]]; then
```

2/test/run
```diff
@@ -220,6 +226,15 @@ VOLUME=`mktemp -d`
 chmod a+rwx $VOLUME
 CONTAINER_ARGS="-u 10000:0 -e OPENSHIFT_JENKINS_JVM_ARCH=i386 "
 run_tests jenkins_test
+cleanup
+sleep 5
+echo "Test: install plugins on start"
+CIDFILE_DIR=$(mktemp --suffix=jenkins_test_cidfiles -d)
+VOLUME=`mktemp -d`
```
stevekuznetsov: Prefer `$()` over back-ticks

2/contrib/s2i/run
```diff
@@ -103,6 +103,25 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
   cp -r /opt/openshift/configuration/* ${JENKINS_HOME}
   rm -rf /opt/openshift/configuration/*
 
+
+  # If the INSTALL_PLUGINS variable is populated, then attempt to install
+  # those plugins before copying them over to JENKINS_HOME
+  # The format of the INSTALL_PLUGINS variable is a comma-separated list
+  # of pluginId:pluginVersion strings
+  if [[ ! -z ${INSTALL_PLUGINS:-} ]]; then
```
stevekuznetsov:
```sh
if [[ -n "${INSTALL_PLUGINS:-}" ]]; then
```

2/contrib/s2i/run
csrwng: why would the statement as it is leak IFS? It's only setting it for the read statement. I think that's cleaner than setting it and cleaning it up after the loop.

csrwng:
@stevekuznetsov addressed your comments except for the IFS one. I think it's cleaner the way it is.

2/contrib/s2i/run
stevekuznetsov: Ah, sorry -- misread it as being on it's own line. I think I've seen `$IFS` set/reset and used with a bare `for` much more often than `read` but I don't feel strongly about it, both will work.

csrwng:
@bparees test added as well

2/test/run
```diff
@@ -216,10 +222,19 @@ cleanup
...
 run_tests jenkins_test
+cleanup
+sleep 5
+echo "Test: install plugins on start"
+CIDFILE_DIR=$(mktemp --suffix=jenkins_test_cidfiles -d)
+VOLUME=$(mktemp -d)
+chmod a+rwx $VOLUME
+VALIDATE_PLUGIN=ghprb
+CONTAINER_ARGS="-u 10000:0 -e OPENSHIFT_JENKINS_JVM_ARCH=x86_64 -e INSTALL_PLUGINS=ghprb:1.35.0"
+run_tests jenkins_test
```
bparees: lest we start installing this plugin by default, it would be good for the test to first confirm the plugin is not normally there.

csrwng:
test updated

bparees:
lgtm, will merge once tests finish.

csrwng:
@bparees just realized I missed actually failing the test when the plugin test fails ... updated.

bparees:
[test]

bparees:
[test]

PR #272, opened 2017-04-10
cleanup invalid config content

bparees:


bparees:
@tdawson fyi, another day another plugin version bump.

@gabemontero ptal, the xml change removes an error i was seeing in the jenkins console related to invalid syntax.

Issue #279, opened 2017-04-24
Missing UTF-8 support, environment variable LANG not set


ropsahl:
Environment variable LANG not set, leading to missing UTF-8 support and Jenkins unable to handle non ASCII characters.
Workaround To get UTF-8 support add following to Dockerfile
ENV LANG="en_US.UTF-8"

bparees:
You can also just set that env variable on the pod/container you're running, I would assume?

bparees:
In general we don't explicitly set a LANG on any of our images.

ropsahl:
I can make my own jenkins, with a dockerfile like:
```
FROM openshift/jenkins-2-centos7
COPY plugins.txt /opt/openshift/configuration/plugins.txt
RUN /usr/local/bin/install-plugins.sh /opt/openshift/configuration/plugins.txt
ENV LANG="en_US.UTF-8"
```
I would prefer using using the solution right out of the box, and I expect more users not to familiar with such configurations ending up spending much time chasing down the problem and finding a solution.

In the container UTF-8 is already set for Jenkins, but the shell where it is running is not. The effect is that I can make a new project in Jenkins with non ASCII characters like Å (LATIN CAPITAL LETTER A WITH RING ABOVE) and it will show correctly in the GUI, however I can not delete it as java.nio access to delete the work area will throw an exception, the same goes for working with source files having UNICODE letters in their path.

A short example from  running bash in a openshift/jenkins-2-centos7 container:
(Probably need UNICODE support in browser to )
Create file and show ls output:
```
$ docker exec -i -t $(docker ps|grep jenkins|awk '{print $1}') bash
bash-4.2$ touch /tmp/æøå_ÆØÅ.tmp
bash-4.2$ ls -lrt /tmp/*.tmp
-rw-r--r--. 1 default root   90382 Apr 25 05:13 jffi7904311933604650928.tmp
-rw-r--r--. 1 default root       0 Apr 25 05:17 ????????_??????.tmp
```
Setting LANG and filename is shown correctly:
```
bash-4.2$ export LANG="en_US.UTF-8"
bash-4.2$ ls -lrt /tmp/*.tmp
-rw-r--r--. 1 default root   90382 Apr 25 05:13 jffi7904311933604650928.tmp
-rw-r--r--. 1 default root       0 Apr 25 05:17 åæøå_ÆØÅ.tmp
```

Jenkins is started with file.encoding UTF8
```
bash-4.2$ ps -ef |grep java
[...]

PR #280, opened 2017-04-24
set locale to en_US.UTF-8 by default

bparees:
fixes https://github.com/openshift/jenkins/issues/279

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -17,7 +17,9 @@ MAINTAINER Ben Parees <bparees@redhat.com>
 ENV JENKINS_VERSION=2.46.1 \
     HOME=/var/lib/jenkins \
     JENKINS_HOME=/var/lib/jenkins \
-    JENKINS_UC=https://updates.jenkins-ci.org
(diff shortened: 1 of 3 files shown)


bparees:
@gabemontero ptal

2/Dockerfile
```diff
@@ -17,7 +17,9 @@ MAINTAINER Ben Parees <bparees@redhat.com>
 ENV JENKINS_VERSION=2.46.1 \
     HOME=/var/lib/jenkins \
     JENKINS_HOME=/var/lib/jenkins \
-    JENKINS_UC=https://updates.jenkins-ci.org
+    JENKINS_UC=https://updates.jenkins-ci.org \
+    LANG en_US.UTF-8 \
```
gabemontero: The image's s2i test actually caught this ... you need an `=` between `LANG` and `en_US.UTF-8`, same for `LC_ALL`.

The msg from the test was `The s2i test failed with `Error response from daemon: Syntax error - can't find = in "LANG". Must be of the form: name=value``

bparees:
thanks @gabemontero fixed

PR #289, opened 2017-05-03
Removed ability to grab ImageStream slave templates at startup

oatmealraisin:
~~Added logic to search for ImageStreamTags with an appropriate annotation. Also, files in /etc/config/slave-templates are added to the list of slaves. This allows users to mount ConfigMaps with multiple slave templates into the Pod.~~

We are moving the PodTemplate population functions to the [OpenShift Sync Plugin](https://github.com/openshift/jenkins-sync-plugin).

@openshift/devex ptal?

### DIFF
--- a/1/contrib/jenkins/kube-slave-common.sh
+++ b/1/contrib/jenkins/kube-slave-common.sh
@@ -61,68 +61,10 @@ if has_service_account; then
   export oc_serviceaccount_name="$(expr "$(oc whoami)" : 'system:serviceaccount:[a-z0-9][-a-z0-9]*:\([a-z0-9][-a-z0-9]*\)' || true)"
 fi
 
-# get_imagestream_names returns a list of image streams that match the
-# SLAVE_LABEL
-function get_is_names() {
-  [ -z "$oc_cmd" ] && return
-  $oc_cmd get is -n "${PROJECT_NAME}" -l role=${SLAVE_LABEL} -o template --template "{{range .items}}{{.metadata.name}} {{end}}"
-}
-
-# convert_is_to_slave converts the OpenShift imagestream to a Jenkins Kubernetes
-# Plugin slave configuration.
-function convert_is_to_slave() {
-  [ -z "$oc_cmd" ] && return
-  local name=$1
-  local template_file=$(mktemp)
-  local template="
-  <org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
-    <inheritFrom></inheritFrom>
-    <name>{{.metadata.name}}</name>
-    <instanceCap>5</instanceCap>
-    <idleMinutes>0</idleMinutes>
-    <label>{{if not .metadata.annotations}}${name}{{else}}{{if index .metadata.annotations \"slave-label\"}}{{index .metadata.annotations \"slave-label\"}}{{else}}${name}{{end}}{{end}}</label>
-    <serviceAccount>${oc_serviceaccount_name}</serviceAccount>
-    <nodeSelector></nodeSelector>
-    <volumes/>
-    <containers>
-      <org.csanchez.jenkins.plugins.kubernetes.ContainerTemplate>
-        <name>jnlp</name>
-        <image>{{.status.dockerImageRepository}}</image>
-        <privileged>false</privileged>
-        <alwaysPullImage>false</alwaysPullImage>
-        <workingDir>/tmp</workingDir>
-        <command></command>
-        <args>\${computer.jnlpmac} \${computer.name}</args>
-        <ttyEnabled>false</ttyEnabled>
-        <resourceRequestCpu></resourceRequestCpu>
-        <resourceRequestMemory></resourceRequestMemory>
-        <resourceLimitCpu></resourceLimitCpu>
-        <resourceLimitMemory></resourceLimitMemory>
-        <envVars/>
-      </org.csanchez.jenkins.plugins.kubernetes.ContainerTemplate>
-    </containers>
-    <envVars/>
-    <annotations/>
-    <imagePullSecrets/>
-    <nodeProperties/>
-  </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
-  "
-  echo "${template}" > ${template_file}
-  $oc_cmd get -n "${PROJECT_NAME}" is/${name} -o templatefile --template ${template_file}
-  rm -f ${template_file} &>/dev/null
-}
-
 # generate_kubernetes_config generates a configuration for the kubernetes plugin
 function generate_kubernetes_config() {
     [ -z "$oc_cmd" ] && return
-    local slave_templates=""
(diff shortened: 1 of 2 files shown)


bparees:
> We could change the behavior to give Admin permissions to anyone with 'edit' role in OpenShift.

we definitely don't want to do that.  the ideal solution would be to be able to generate a local jenkins account that could be used for this (ie one that is not auth'd via openshift).  barring that possibility (and i'm pretty sure @gabemontero will tell us it's not possible, at least w/o reworking the openshift login plugin significantly), i'm not sure what we want to do. 

Switching back to the approach of putting this logic in the openshift sync plugin might be the right thing to do since it won't need creds (as it is already inside the jenkins process).

gabemontero:
On Wed, May 3, 2017 at 10:18 PM Ben Parees <notifications@github.com> wrote:

> We could change the behavior to give Admin permissions to anyone with
> 'edit' role in OpenShift.
>
> we definitely don't want to do that. the ideal solution would be to be
> able to generate a local jenkins account that could be used for this (ie
> one that is not auth'd via openshift). barring that possibility (and i'm
> pretty sure @gabemontero <https://github.com/gabemontero> will tell us
> it's not possible, at least w/o reworking the openshift login plugin
> significantly), i'm not sure what we want to do.
>
There was a time we special cased the predefined "admin" ID but you had me
remove that @bparees.  It wouldn't be too hard to resuscitate it but I
think the reasoning to remove was and still is valid.

> Switching back to the approach of putting this logic in the openshift sync
> plugin might be the right thing to do since it won't need creds (as it is
> already inside the jenkins process).
>
I believe one of the plugins is the right place.  And it is probably sync
plugin but maybe @oatmealraisin can talk me through the flow in detail on
Thusday and I can see if any new ideas arise on how we package this from a
java perspective.

bparees:
> There was a time we special cased the predefined "admin" ID but you had me remove that @bparees.

i don't remember all my reasoning but i suspect that most of them probably related specifically to calling it admin, or how we setup the password for it, so i could probably convince myself that generating a unique/random ID and password for each jenkins instance, that could be used for this, might be ok... but mostly I think we're in agreement that we should just "do the right thing" and grow the sync plugin capabilities to cover this.

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -107,11 +116,75 @@ function convert_is_to_slave() {
     <nodeProperties/>
   </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
   "
+  echo $(expand_template "is/${name}" "${template}")
```
stevekuznetsov: Useless `echo`

```sh
expand_template "is/${name}" "${template}"
```

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -107,11 +116,75 @@ function convert_is_to_slave() {
...
+        <envVars/>
+      </org.csanchez.jenkins.plugins.kubernetes.ContainerTemplate>
+    </containers>
+    <envVars/>
+    <annotations/>
+    <imagePullSecrets/>
+    <nodeProperties/>
+  </org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
+  "
+  echo $(expand_template "istag/${name}" "${template}")
```
stevekuznetsov: ditto

oatmealraisin:
Somewhat blocked by [#291](https://github.com/openshift/jenkins/pull/291) right now. @stevekuznetsov Any other things I need to change? I've removed the echoes.

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -68,12 +68,21 @@ function get_is_names() {
   $oc_cmd get is -n "${PROJECT_NAME}" -l role=${SLAVE_LABEL} -o template --template "{{range .items}}{{.metadata.name}} {{end}}"
 }
 
+# get_istag_names returns a list of imagestreamtags that match the SLAVE_LABEL
+# role
+function get_istag_names() {
+  [ -z "$oc_cmd" ] && return
```
stevekuznetsov: This is clearly not your invention but it's unclear why it needs to be in every single function, seems more like a clear precondition for any part of this script working

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -68,12 +68,21 @@ function get_is_names() {
...
+  $oc_cmd get istag \
+     -n "${PROJECT_NAME}" \
+     -o 'go-template' \
+     --template="{{ range \$istag := .items }}{{ if \$istag.tag }}{{ if \$istag.tag.annotations }}{{ if index \$istag.tag.annotations \"role\"}}{{ if eq (index \$istag.tag.annotations \"role\") \"${SLAVE_LABEL}\"}}{{\$istag.metadata.name}}{{end}}{{end}}{{end}}{{end}}{{end}}"
```
stevekuznetsov: This can probably be:

```sh
$oc_cmd get istag \
    --namespace "${PROJECT_NAME}" \
    -o jsonpath="{ .items[?@.tag.annotations.role=='${SLAVE_LABEL}'].metadata.name }"
```

In general JSONPath will make all of these about 100000% clearer IMHO

1/contrib/jenkins/kube-slave-common.sh
```diff
@@ -107,11 +116,75 @@ function convert_is_to_slave() {
...
+function convert_istag_to_slave() {
+  [ -z "$oc_cmd" ] && return
+  local name=$1
+  local template="
+  <org.csanchez.jenkins.plugins.kubernetes.PodTemplate>
+    <inheritFrom></inheritFrom>
+    <name>{{.metadata.name}}</name>
+    <instanceCap>5</instanceCap>
+    <idleMinutes>0</idleMinutes>
+    <label>${name}</label>
```
stevekuznetsov: Also clearly not your invention but having these inside of real `.xml` files for syntax highlighting and easy inspection, with this function using an `.xsl` to template it or at least something like `xmlstarlet` for `xpath` editing if the XSL approach is too heavy-weight.

oatmealraisin:
@stevekuznetsov @bparees @gabemontero Since we're moving this to the sync plugin, I've removed it from the script. ptal?

bparees:
this is ok, but we can't do it until we've updated the sync plugin in this image to a level that handles that this script logic used to do.

gabemontero:
With v0.1.18 of the sync plugin and the latest docker.io/openshift/jenkins-2-centos7:latest  image containing that version, we can now move forward with this removal.

going to [test] and make sure the PR is still hunky-dory

gabemontero:
ah ... this PR wasn't rebased to include the install plugins fix ... i'll check out this PR branch, rebase it to my own branch, and move forward when I'm back from PTO.

PR #298, opened 2017-05-24
Gradle slave

tjololo:
slave-gradle folder is based on the slave-maven folder.
Have not added a record in [2/contrib/jenkins/kube-slave-common.sh](https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh) since I can't say what the image source will be.


slave-gradle/Dockerfile.rhel7
```diff
@@ -0,0 +1,44 @@
...
+    ENV=/usr/local/bin/scl_enable \
+    PROMPT_COMMAND=". /usr/local/bin/scl_enable"
+ENV PATH=$PATH:/opt/gradle-${GRADLE_VERSION}/bin
+
+# Install Maven
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \    
+    yum-config-manager --disable epel >/dev/null || : && \
+    INSTALL_PKGS="java-1.8.0-openjdk-devel" && \
+    yum install -y $INSTALL_PKGS && \
+    curl -LOk https://services.gradle.org/distributions/gradle-${GRADLE_VERSION}-bin.zip && \
```
bparees: the rhel image packages must come from official redhat rhel package sources.

slave-gradle/Dockerfile.rhel7
```diff
@@ -0,0 +1,44 @@
...
+    unzip gradle-${GRADLE_VERSION}-bin.zip -d /opt && \
+    rm -f gradle-${GRADLE_VERSION}-bin.zip && \
+    rpm -e java-1.7.0-openjdk --nodeps || true && \
+    rpm -e java-1.7.0-openjdk-devel --nodeps || true && \
+    rpm -e java-1.7.0-openjdk-headless --nodeps || true && \
```
bparees: these lines were only needed in the maven image because the maven package mistakenly drags these deps in.  you don't need them here.  (unless the gradle package also drags jdk1.7)

bparees:
@jstrachan @rawlingsj any opinion on having a single maven+gradle slave image vs 2 separate slave images?

slave-gradle/Dockerfile.rhel7
tjololo: Then we should probably use the same strategy for the centos image, just install the package from the package repository. What do you think?

slave-gradle/Dockerfile.rhel7
tjololo: Waiting with the change until the decision of it will be one maven/gradle image or two separate images

jstrachan:
I guess a maven + gradle image might make things simpler when folks are using the default OOTB build pod template in the kubernetes plugin. Though we're also big fans of specifying your build pod in your Jenkinsfile - or at least use a [pipeline library to automate that stuff for you](https://github.com/fabric8io/fabric8-pipeline-library/#understanding-how-it-works) then different git repos can iterate on upgrading, say, maven versions independently in a more microservice-y way.

We've been having similar discussions lately around aligning developer/build tools with pipelines and eclipse Che; folks kinda like having a terminal where they have lots of tools available (maven, git, unzip, kubectl, oc, curl etc) and it'd be nice to share the same underlying docker image for building and use in Che terminals. (I'm excluding the s2i part here - just mean the pipeline build pod and Che pod).

The traditional docker/k8s way is to keep images separate and composable to maximise reuse; but for the 'developer tools containers' it simplifies build pods and developer Che terminals if we bake them into a single container really.

So we're pondering having a curated set of 'developer tools' images that layer on centos/RHEL then there's the base images for java/golang/node apps and then we layer on the language specific tools (maven/gradle/npm etc) and then add another common layer of tools (git / unzip / kubectl / oc / skopeo etc) so that there are common tools around all in a single image we can use in Che or in Pipeline build pods from a single image. 

TL;DR composing developer tool images together sounds good to me; makes it easier for pipelines & Che tooling. It seems "tooling" images its fine to put anything you may need inside; its actual base images + users apps that we have to try keep things fairly minimal

slave-gradle/Dockerfile.rhel7
bparees: yup.

bparees:
@jstrachan cool, thanks, I think i'm on the same page.  @tjololo so i think it would make sense to just submit a PR that installs the gradle packages in the existing maven dockerfiles.

tjololo:
Do you want me to create a new PR or just update this one?

bparees:
@tjololo whatever you prefer.  updating this one might be better to preserve the discussion of why we went with a single image.

tjololo:
@bparees Seems like there is no gradle package in rhel or centos repository.

bparees:
> @bparees Seems like there is no gradle package in rhel or centos repository.

you can install it from the internet for centos, like you're already doing, but we won't be able to ship a rhel image until we have a package for it.  @smunilla any assistance you can offer w/ tracking down or creating a rhel gradle package?

slave-maven/test/run
```diff
@@ -8,5 +8,6 @@
 
 docker run ${IMAGE_NAME} oc
 docker run --entrypoint="/bin/bash" ${IMAGE_NAME} "-c" "mvn --version"
+docker run --entrypoint="/bin/bash" ${IMAGE_NAME} "-c" "gradle --version"
```
tjololo: This might break the test against the rhel7 image in your ci tool?

bparees: it'll break until we have a gradle package for rhel, yeah.  it'd be nice if we could address that, but the alternative is to only run this particular test when the imagename doesn't contain rhel, for now.

tjololo:
test/run does not run "gradle --version" if the imagename contains the string rhel

slave-maven/Dockerfile
```diff
@@ -3,27 +3,34 @@ FROM openshift/jenkins-slave-base-centos7
 MAINTAINER Ben Parees <bparees@redhat.com>
 
 ENV MAVEN_VERSION=3.3 \
+    GRADLE_VERSION=3.5 \
     BASH_ENV=/usr/local/bin/scl_enable \
     ENV=/usr/local/bin/scl_enable \
     PROMPT_COMMAND=". /usr/local/bin/scl_enable"
+ENV PATH=$PATH:/opt/gradle-${GRADLE_VERSION}/bin
```
bparees: do this as part of the existing ENV command so it doesn't add a new layer.

bparees:
[test]

bparees:
our CI system appears to be slightly broken at the moment, if you rebase your PR on the latest master it should resolve the test failure (normally that would not be necessary).

tjololo:
rebased on the latest master and removed the extra ENV layer

bparees:
[test]

slave-maven/test/run
```diff
@@ -5,8 +5,13 @@
 # IMAGE_NAME specifies the name of the candidate image used for testing.
 # The image has to be available before this script is executed.
 #
+# TODO: When the rhel image get gradle remove the if
+#
 
 docker run ${IMAGE_NAME} oc
 docker run --entrypoint="/bin/bash" ${IMAGE_NAME} "-c" "mvn --version"
+if [[ ${IMAGE_NAME} != *"rhel"* ]]; then
+    docker run --entrypoint="/bin/bash" ${IMAGE_NAME} "-c" "gradle --version"
```
bparees: sorry for the delay on this, got caught up in other things.  the test is failing because of the nature of our entrypoint and some setup that is required.  this should pass:
`docker run --entrypoint="/bin/sh" ${IMAGE_NAME} "-c" "/usr/local/bin/run-jnlp-client;gradle --version"`

tjololo: @bparees no problem have been trying to find why the test didn't work. Now it should work fine

bparees:
[test]

Ben Parees | OpenShift

bparees:
woo.  sorry for the long road on this one, @tjololo but we appreciate the contribution!  After I merge it, the slave image should be updated on dockerhub within an hour or so.

PR #306, opened 2017-06-05
based on older master

bparees:


bparees:
[test]

bparees:
@dobbymoodge @stevekuznetsov so as i suspected, we are not testing these PRs on top of master.  

compare my other PR which passed tests here:
https://github.com/openshift/jenkins/pull/305

both PRs are making the same no-op change, but this PR branch is based on an older version of master that had an issue w/ the jenkins plugin installer script which causes the test to fail.

stevekuznetsov:
From the test in this PR, thus is how we're grabbing the commit to test:

```
+ git fetch --tags --progress origin pull/306/head:pull/306
From https://github.com/openshift/jenkins
 * [new ref]         refs/pull/306/head -> pull/306
++ git log -1 --pretty=%H pull/306
+ commit_to_test=afbc94e8618f83e727f020ff5b5b808d81db3a0f
```

stevekuznetsov:
We then ran the tests like so:

```
+ vagrant test-origin-image --image jenkins --source openshift/jenkins --ref afbc94e8618f83e727f020ff5b5b808d81db3a0f
```

bparees:
@stevekuznetsov well it's got the right commit, which isn't surprising from what i've seen, but it doesn't seem to have rebased/merged that commit on top of the most recent master correctly.

stevekuznetsov:
Or are you saying we're not merging the PR into master? If so, yeah, probably not. After trying to make that work and realizing `vagrant test-origin-image` didn't use the normal `vagrant sync-$foo; vagrant test-$foo` flow I just made the job work as it did before Rosie was involved.

dobbymoodge:
@bparees @stevekuznetsov That look like the wrong commit to me, if I do `git co $commit_to_test ; git diff origin/master` I see more than just a no-op.

bparees:
@dobbymoodge the commit is right, but the commit was not rebased on top of the latest master like it's supposed to be.

bparees:
(or git merge isn't doing what we think it is when we merge the PR branch into the master branch)

dobbymoodge:
@bparees Where does that `$commit_to_test` code come from, anyway? Don't you want to test the merge commit, not the one prior to it? That's why it's pulling the "old" master.

dobbymoodge:
e.g. in this diagram, A is $commit_to_test, it comprises D and E
master comprises D, E, F, and G
H is A + current master, which is what should be tested
$commit_to_test is H-1
H-1 resolves to commit A, which has the new code but not the intervening commits F, G and H
```
       --------A  topic
      /         \
 D---E---F---G---H master
```

bparees:
> @bparees Where does that $commit_to_test code come from, anyway? 

don't know. i'd expect this job to behave exactly like the origin job behaves, and the origin job always did a proper rebase onto current master.

> Don't you want to test the merge commit, not the one prior to it? That's why it's pulling the "old" master.

yes.

stevekuznetsov:
> i'd expect this job to behave exactly like the origin job behaves

Why? This was not how the job worked before Rosie, and barring some serious re-wiring of the Vagrant plugin, it will not work that way. I know we had mentioned that moving to Rosie _could_ make PRs start to be merged before tests, but upon more investigation I found it was going to take too much work and opted to not change the job.

stevekuznetsov:
Also @dobbymoodge that diagram is totally not how we're grabbing the commit SHA to pass to the test. There's no merge. We're just looking at the head of the branch at the source of the PR.

bparees:
> Why? This was not how the job worked before Rosie, and barring some serious re-wiring of the Vagrant plugin, it will not work that way.

I guess i misunderstood what you were doing when you created the rosie job and updated the vagrant plugin.

> I know we had mentioned that moving to Rosie could make PRs start to be merged before tests, but upon more investigation I found it was going to take too much work and opted to not change the job.

ok, i missed or forgot that.

> Also @dobbymoodge that diagram is totally not how we're grabbing the commit SHA to pass to the test. There's no merge. We're just looking at the head of the branch at the source of the PR.

then what is all this?

```
Local merging pull request #306 for repo 'jenkins'
+ pushd jenkins
+ git checkout master
Already on 'master'
+ git checkout -b tpr_broken_test_bparees
Switched to a new branch 'tpr_broken_test_bparees'
+ git pull git@github.com:bparees/jenkins.git broken_test
From github.com:bparees/jenkins
 * branch            broken_test -> FETCH_HEAD
+ git pull git@github.com:bparees/jenkins.git broken_test --tags
From github.com:bparees/jenkins
 * branch            broken_test -> FETCH_HEAD
+ git checkout master
Switched to branch 'master'
+ git merge tpr_broken_test_bparees
+ git submodule update --recursive
+ popd
```

stevekuznetsov:
IIRC the previous behavior was that the Jenkins master would check out the PR branch and the HEAD SHA would be passed to `vagrant test-origin-image`

bparees:
@stevekuznetsov no argument about the previous behavior (as provided by the github PR tester plugin), i just thought the automated rebase/merge into master before testing behavior was part of what we were getting when this was switched to be rosie based.

stevekuznetsov:
> I guess i misunderstood what you were doing when you created the rosie job and updated the vagrant plugin.

Yeah, I started down that road but realized it would require a rewrite of that test code and bailed. I think I worked through this with Gabe, sorry that I did not adequately communicate this to you too.

> then what is all this?

That code is doing the merge, locally, on the Jenkins master. That code is synced to the remote VM but the SHA you pass to the plugin overrides it anyway. I should clean up the job to not do the local merge + sync since it's unused.

stevekuznetsov:
>  i just thought the automated rebase/merge into master before testing behavior was part of what we were getting when this was switched to be rosie based.

Yeah, sorry for the misunderstanding. If the `vagrant test-origin-image` code followed the `local-checkout`, `sync-$foo`, and `test-$foo` pattern you would indeed get this. If you'd like to change the test, I've done the sync half so that would get you where you want to be.

bparees:
> That code is doing the merge, locally, on the Jenkins master. That code is synced to the remote VM but the SHA you pass to the plugin overrides it anyway. I should clean up the job to not do the local merge + sync since it's unused.

could we just change the SHA we pass to be the merge commit?

stevekuznetsov:
> could we just change the SHA we pass to be the merge commit?

The `vagrant test-origin-image` code looks at the GitHub remote for the SHA, I think. The merge commit only exists locally.

dobbymoodge:
> Also @dobbymoodge that diagram is totally not how we're grabbing the commit SHA to pass to the test. There's no merge. We're just looking at the head of the branch at the source of the PR.

So we don't need the `local_merge_pull_request` line at all, but we do need to properly merge the PR against current master on the test instance, huh?

bparees:
> The vagrant test-origin-image code looks at the GitHub remote for the SHA, I think. The merge commit only exists locally.

oh. the way the origin stuff used to work was to merge locally and then sync the local repo over to the ec2 instance.  assumed something similar was happening here.  sounds like we're doing the merge(for no reason at this point) but not the sync.

dobbymoodge:
@bparees @stevekuznetsov It looks like `test-origin-image` doesn't merge the PR, it just checks out the ref provided in `$commit_to_test`: https://github.com/openshift/vagrant-openshift/blob/master/lib/vagrant-openshift/command/test_origin_image.rb#L127

I seriously doubt this is the intended behavior...

stevekuznetsov:
We are doing the sync, I just can't say for sure Thad the code in `vagrant test-origin-image` will check it out correctly. Feel free to try.

bparees:
note to self:
`git fetch origin refs/pull/306/merge:my_new_branch; git checkout my_new_branch`

does what we need.

stevekuznetsov:
I did some testing with `/merge` and it did not always do what we needed. Can't remember the specifics.

bparees:
if all my rosie work has paid off, this [test] should pass now.

stevekuznetsov:
Nice

bparees:
woot.  thanks for your help @stevekuznetsov

PR #319, opened 2017-06-21
Make the images FROM: openshift/origin:latest

stevekuznetsov:
We need to determine the size differences between the images before and
after this change to ensure that they are not egregiously large, but it
makes more sense to base this image off of the latest Origin image in
order to incorporate the appropriate version of the `oc` client binary.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @bparees


1/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:centos7
+FROM openshift/origin:latest
```
bparees: we're deprecating the v1 image this release, i'd rather not touch it at this point.

2/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:centos7
+FROM openshift/origin:latest
```
bparees: i don't want to make this change w/o also changing our rhel image.

2/Dockerfile.rhel7
```diff
@@ -37,7 +37,7 @@ EXPOSE 8080 50000
...
...ery-detached jenkins-plugin-junit jenkins-plugin-kubernetes jenkins-plugin-mailer jenkins-plugin-mapdb-api jenkins-plugin-matrix-project jenkins-plugin-matrix-auth jenkins-plugin-mercurial jenkins-plugin-momentjs jenkins-plugin-multiple-scms jenkins-plugin-openshift-pipeline jenkins-plugin-openshift-login jenkins-plugin-openshift-sync jenkins-plugin-pipeline-build-step jenkins-plugin-pipeline-graph-analysis jenkins-plugin-pipeline-input-step jenkins-plugin-pipeline-rest-api jenkins-plugin-pipeline-stage-step jenkins-plugin-pipeline-stage-view jenkins-plugin-pipeline-utility-steps jenkins-plugin-plain-credentials jenkins-plugin-scm-api jenkins-plugin-script-security jenkins-plugin-ssh-credentials jenkins-plugin-structs jenkins-plugin-subversion jenkins-plugin-workflow-aggregator jenkins-plugin-workflow-api jenkins-plugin-workflow-basic-steps jenkins-plugin-workflow-cps jenkins-plugin-workflow-cps-global-lib jenkins-plugin-workflow-durable-task-step jenkins-plugin-workflow-job jenkins-plugin-workflow-multibranch jenkins-plugin-workflow-remote-loader jenkins-plugin-workflow-scm-step jenkins-plugin-workflow-step-api jenkins-plugin-workflow-step-api jenkins-plugin-workflow-support" && \
```
stevekuznetsov: This snipped out `atomic-openshift-clients`

bparees:
pretty sure this will fail since openshift3/ose 

1) won't be available in the context where we build the jenkins image for testing
2) won't have the subscription hacks we do to the rhel7 image that *is* available

currently the CI job for this repo

1) builds a new rhel7 image w/ some hacks to make it able to grab the necessary RPMs
2) builds based on that image (this is NOT the image that gets published to customers, it's for internal testing purposes only.  the real image gets built on osbs, where openshift3/ose should be available).

so this is going to take a little more thinking, and possibly changing our CI job.

stevekuznetsov:
> so this is going to take a little more thinking, and possibly changing our CI job.

I'm still not entirely certain it's the best move, but I do want some more fluid/configurable dependency relationship so I can build the Jenkins image without having to change your manifest.

bparees:
yeah i'm not explicitly against it, just want to make sure we line up all the ducks.

2/Dockerfile
bparees: also @smarterclayton do you have any concerns w/ our jenkins image being based on openshift/origin?  (the benefit is it implicitly includes the latest oc client so we don't have to explicitly install it in our image and keep bumping the version)

bparees: @smarterclayton bump ^^

smarterclayton:
No I do not have an issue.

bparees:
i'm finishing this up over here:
https://github.com/openshift/jenkins/pull/356

thanks for getting the ball rolling, @stevekuznetsov

PR #338, opened 2017-07-26
Need to quote JAVA_GC_ARGS

scoheb:


bparees:
did you see an issue w/ this?

scoheb:
Not sure you saw my commit message but:

>> otherwise, we get a "/usr/libexec/s2i/run: line 63: [: too many arguments" error

I wanted to override the _-XX:MaxMetaspaceSize=100m_ parameter...and to do that I needed to add the **JAVA_GC_ARGS** . When I added it to the DeploymentConfig, this error appeared. Does not pose a problem for me now since my JAVA_GC_ARGS value made it to the launched process.

bparees:
hm.  the potential error makes sense, though i'm unable to reproduce it in a simple bash script.  Anyway please squash your commits and i'll merge.

bparees:
[merge]

Issue #335, opened 2017-07-20
Allow more fine-grained control for adding/updating plugins using s2i method when using Jenkins PV


scoheb:
Scenario: Provide more control to update plugins on already configured Jenkins instance
Given a Jenkins instance created using the s2i method
And the s2i config contains a plugins.txt file
And the instance is configured using a PV
And the instance has been initially deployed to create JENKINS_HOME on PV
When the image is updated to include a update to plugins.txt to add/update a plugin
And the s2i builder is configured to use OVERRIDE_EXISTING_PLUGINS
Then the images' plugins are used instead of the already existing plugin set in JENKINS_HOME in PV

bparees:
@gabemontero we should look at everything we do/do not copy when the PV is already "configured" and make each type of content (plugins, config, anything else?) individually configurable as "overwrite existing" or not.

gabemontero:
@bparees shall we create a trello card for this as well, or do you want to handle outside of sprint planning?

bparees:
@gabemontero i don't think we need a card for this, it should be pretty trivial.

gabemontero:
As of this comment, an inventory of various config related activites:

1) Dockerfile / docker build
    - oc binary is copied to /usr/local/bin
    - base-plugins.txt and password-encoder.jar are copied to /opt/openshift
    - config.xml.tpl, credentials.xml.tpl, the openshift sample job config, disabling of SSHD via its xml config file,  and the admin user config (for when oauth is disabled) are copied to /opt/openshift/configuration
    - fix-permissions, install-plugins.sh, jenkins-common.sh, and kube-slave-common.sh are copied to /usr/local/bin
    - the s2i assemble and run scripts are copied to  /usr/libexec/s2i
    - the default plugins are installed and various permissions are set so we can switch between 32 and 64 bit java

2) When the s2i run script is launched during jenkins startup, and the jenkins instance is considered "not yet configured" (i.e. JENKINS_HOME/configured or /var/lib/jenkins/configured does not exist), the following occurs:

   - if /opt/openshift/configuration/config.xml does not yet exist, it grabs the config.xml.tpl (copied over by the docker file) and generates a config.xml file, updating the k8s plugin config based on what is mounted in the pod, and if you have overridden the images used for the slave, and the jnlp service, via env vars

[...]

PR #346, opened 2017-08-03
allow override of plugin list, jenkins config, after initial containe…

gabemontero:
…r start

Fixes https://github.com/openshift/jenkins/issues/335

@bparees @scoheb

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -9,6 +9,65 @@
 source /usr/local/bin/jenkins-common.sh
 source /usr/local/bin/kube-slave-common.sh
 
+function update_admin_password() {
+    sed -i "s,<passwordHash>.*</passwordHash>,<passwordHash>$new_password_hash</passwordHash>,g" "${JENKINS_HOME}/users/admin/config.xml"
+    echo $new_password_hash > ${JENKINS_HOME}/password
+}
+
+function create_jenkins_config_xml() {
+    # copy the default configuration from the image into the jenkins config path (which should be a volume for persistence).
+    if [ ! -f "${image_config_path}" ]; then
+	# If it contains a template (tpl) file, we can do additional manipulations to customize
+	# the configuration.
+	if [ -f "${image_config_path}.tpl" ]; then
+	    export KUBERNETES_CONFIG=$(generate_kubernetes_config)
+	    echo "Generating kubernetes-plugin configuration (${image_config_path}.tpl) ..."
+	    envsubst < "${image_config_path}.tpl" > "${image_config_path}"
+	fi
+    fi
+}
+
+function create_jenkins_credentials_xml() {
+    if [ ! -f "${image_config_dir}/credentials.xml" ]; then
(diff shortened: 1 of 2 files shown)


gabemontero:
[test] passed locally ... let's see how ec2/ci.openshift handle it

gabemontero:
quick update ... refactoring this PR along the discussion in the issue ... will push an update once I overcome some bash parsing pain that I'm currently enduring

gabemontero:
disdain for bash reaffirmed ... updates pushed

gabemontero:
the git diff is less than ideal, even with `?w=1` ... but the change moved common blocks of existing code to common routines, and added an else block to the if !configured then block

bparees:
mostly lgtm but i do wonder if we need to have per-config file granularity on this.  It seems like maybe "overwrite pv config" and "overwrite pv plugins" might be sufficient for the requested use case and easier for users.. wdyt?

gabemontero:
yeah, that was my original thought, and went per config file based on my interpretation of your comment https://github.com/openshift/jenkins/issues/335#issuecomment-316701484

I'm fine either way, and if I misinterpreted your comment and went overboard wrt granularity, I can certainly course correct.

bparees:
> yeah, that was my original thought, and went per config file based on my interpretation of your comment #335 (comment)
> I'm fine either way, and if I misinterpreted your comment and went overboard wrt granularity, I can certainly course correct.

yeah let's just go w/ plugins + config as the two things you can choose to keep from the PV, or replace from the image.  (with the default still being to keep what's on the PV)

gabemontero:
OK, I've left some of the new functions just cause I think it makes the script's main path more readable, but I've removed the specific credentials/sshd/admin user overrides, leaving just main jenkins config and plugins.

Also made a small update to the readme to document the new env vars.

bparees:
lgtm. @scoheb are you happy w/ this implementation/capability?

scoheb:
so for my use case... I just need to set 

> OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS

Looks good to me!

Thanks guys!

gabemontero:
cool / thanks @scoheb !

I'll merge after the last test complete @bparees - thanks !

gabemontero:
[merge]

bparees:
@gabemontero let's make sure the new env vars make it into the docs (repo
readme at a minimum but probably the official docs need an explanation of
how config is managed too)


Ben Parees | OpenShift

bparees:
Oops, you covered the readme already. Nm.

Ben Parees | OpenShift

gabemontero:
mirror issue building the centos image during merge ... trying again

[merge]

gabemontero:
ahhhh ... flake https://github.com/openshift/origin/issues/14898

[merge]

gabemontero:
ok ... flake https://github.com/openshift/jenkins/issues/348 in the jenkins repo to pacify rosie

[merge]

PR #353, opened 2017-08-11
deprecate/stop testing the v1 image

bparees:


bparees:
@gabemontero ptal
[test]

bparees:
> lgtm ... in addition to the push image job you changed, there is a test job on aos-cd-jobs to change as well, right?

the test job just runs make test, this will cover it.
https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_jenkins_images.yml#L19-L20

bparees:
[merge]

gabemontero:
On Fri, Aug 11, 2017 at 11:26 AM, Ben Parees <notifications@github.com>
wrote:

> lgtm ... in addition to the push image job you changed, there is a test
> job on aos-cd-jobs to change as well, right?
>
> the test job just runs make test, this will cover it.
> https://github.com/openshift/aos-cd-jobs/blob/master/sjb/
> config/test_cases/test_branch_jenkins_images.yml#L19-L20
>

ah cool thx for info

PR #356, opened 2017-08-12
use origin as the base image so we don't have to match the oc binary

bparees:
finalization of https://github.com/openshift/jenkins/pull/319


bparees:
@gabemontero @stevekuznetsov ptal.  https://github.com/openshift/aos-cd-jobs/pull/562 laid the groundwork for this.
[test]

bparees:
as feared, the brew registry isn't accessible, so either we need to build on top of the last published image (which will always be way behind from where we want to be), or we need to build the origin/ose image locally first.

unless @stevekuznetsov can tell me a location where we are publishing the last build master ose image that can be accessed.

slave-base/Dockerfile.rhel7
```diff
@@ -17,7 +17,7 @@ ENV HOME=/home/jenkins
 USER root
 # Install headless Java
 RUN yum-config-manager --disable epel >/dev/null || : && \
-    INSTALL_PKGS="atomic-openshift-clients bc gettext git java-1.8.0-openjdk-headless java-1.8.0-openjdk-headless.i686 lsof rsync tar unzip which zip" && \
+    INSTALL_PKGS="bc gettext git java-1.8.0-openjdk-headless java-1.8.0-openjdk-headless.i686 lsof rsync tar unzip which zip" && \
```
gabemontero: what was the motivation for installing java in the base vs. maven?

stevekuznetsov:
We historically have not had and contemporarily continue to not have a well-formed and complete CI pipeline for OCP/OSE on RHEL.

slave-base/Dockerfile.rhel7
bparees: That's not a change in this pr, but we need java in the base to run the jenkins slave client. 

You may have seen in the maven image that i'm removing logic that previously removed(not added) the 1.7 jdk. That logic was needed before because the maven rpms incorrectly pulled in 1.7 when we were already installing 1.8 explicitly. That rpm dependency has been fixed, so we don't need that cleanup logic anymore.

bparees:
Do we push the images to the ci.dev registry? That would be a place to get
the base image from potentially too.

Ben Parees | OpenShift

stevekuznetsov:
Would need to ask @jupierce if we push `openshift3/ose-` images to that registry

bparees:
ok I switched the job to build the local "rhel" image based on registry.access.redhat.com/openshift3/ose for now.  That means the ci.dev image will have an older oc client binary, but i don't think anyone is using those images these days anyway since we've gotten better at spinning the brew images frequently.

[test]

bparees:
[test]

jupierce:
@bparees @stevekuznetsov We don't currently push to ci.dev. The build process pushes all images to registry.reg-aws.openshift.com:443/online  (which unfortunately requires authentication to pull).

bparees:
[test]

bparees:
@smunilla @jupierce fyi, this is going to change the base image for the jenkins images to be the openshift/origin (openshift3/ose) images so we don't have to explicitly install the oc client.  We need to make sure the dockerfiles in dist-git reflect the FROM change once this merges.

bparees:
[merge]

Issue #361, opened 2017-08-21
error: mkdir /var/lib/origin/openshift.local.config: permission denied


slinger-scm:
Hello guys!
I'm having this issue while trying to execute my pipeline, any idea I saw that has a push to dockerhub 4 days ago in the image that my pipeline use as slave.. https://hub.docker.com/r/openshift/jenkins-slave-nodejs-centos7/tags/

```
OpenShift Build tasks-pipeline13
[Pipeline] node
Still waiting to schedule task
All nodes of label ‘nodejs’ are offline
Running on nodejs-69e36c499aac1 in /tmp/workspace
[Pipeline] {
[Pipeline] sh

[cicd] Running shell script

oc project cicd
error: mkdir /var/lib/origin/openshift.local.config: permission denied
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: script returned exit code 1
Finished: FAILURE
```

bparees:
can you share the pod logs from the slave pod that got spun up (nodejs-69e36c499aac1)?

also if you can share your jenkinsfile/pipeline definition that could help.

maiconbaumx:
Hello @bparees! That post wast initiated from my second account, the @slinger-scm.
This issue was present in the latest jenkins-slave-nodejs-centos7 image from OpenShift's Dockerhub repository. After several tries, I decided to build a new Jenkins Slave image from this Dockerfile and use for my pipeline. I don't have these logs anymore, but this error appears in the logs.. After build a new image and giving the /var/lib/jenkins directory permission for write  of user 1001, everything works fine... My JenkinsFile:



node('nodejs') {
   def sourceCode = "xx"
   def webhook = "xx"
   def application = "xx"
   def urlDev = "xx"
   def urlProd = "xx"
   def secret = "xx"
   def projectDEV = "xx"
   def projectPROD = "xx"
   def webhookDEV = "xx"
   def webhookPROD = "xx"

   stage 'Clean'
   sh "oc project $projectDEV"
   sh "oc delete bc,dc,svc,route -l app=${application} -n $projectDEV"

   stage 'Build'
   pauseAlert(true,3)
   git branch: "master", credentialsId: 'xx', url: 'xx'
   def v = version()
[...]

PR #363, opened 2017-08-22
loosen /var/lib/origin permissions for oc

bparees:
fixes https://github.com/openshift/jenkins/issues/361

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -55,6 +55,7 @@ RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
     chmod 775 /usr/bin && \
     chmod 775 /usr/lib/jvm-exports && \
     chmod 775 /usr/share/man/man1 && \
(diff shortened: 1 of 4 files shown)


bparees:
@gabemontero @csrwng ptal.  i'm a little concerned this seems to have just broken, any idea if i'm missing something here?

bparees:
[test]

bparees:
(also surprised it didn't break any of our jenkins openshift client plugin testing?)

csrwng:
lgtm

bparees:
well that makes me feel better that it's probably not a regression.
[merge]

gabemontero:
re: the curiosity on previous breakage ...we don't do much if any project creation within our pipelines that I'm aware of ... I thought by default the jenkins SA by default does not have permission to do that anyway ... or I at least seem to remember it being that way at some point

bparees:
[merge]

bparees:
[merge]

PR #365, opened 2017-08-24
set entrypoint to null in rhel image

bparees:


bparees:
@gabemontero this will fix the rhel image issue you saw once it makes its way to dist-git.
[merge]

bparees:
@smunilla @jupierce we need to make sure this gets into distgit and the jenkins images get respun, what's on brew right now is totally broken (not sure why QE hasn't noticed?)

jupierce:
@adammhaile We will have a couple reconciliations for 3.7 jenkins dockerfiles.

PR #378, opened 2017-09-13
include hidden files when copying and moving in s2i/run

silbernm:
I've added "shopt -s dotglob" in s2i/run as well, such that e.g. ${JENKINS_INSTALL_DIR}/configuration/.gitconfig file will be copied as well.

Furthermore, I moved the "shopt -s dotglob" command in s2i/assemble to the top, since it changes the global state of the bash process and should not be hidden in the code. As a consequence, hidden files from ${JENKINS_INSTALL_DIR}/plugins will also be copied, but I can't think of any harm there.


bparees:
lgtm
[merge]

PR #379, opened 2017-09-13
template all *.tpl files in the Jenkins configuration

silbernm:
The environment substitution process in s2i/run is quite handy. At the moment, it only works for the config.xml and credentials.xml file, but in my case it would be great if it would also fill in environment variables in other configuration files, e.g.
* hudson.plugins.sonar.SonarGlobalConfiguration.xml
* proxy.xml
* config.xml of seed jobs

I've extended s2i/run to substitute all *.tpl files in the tree under /opt/openshift/configuration, during the initial configuration run of Jenkins. I took care not to modify the special cases for the config.xml and credentials.xml files.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -41,6 +41,22 @@ function create_jenkins_credentials_xml() {
   fi
 }
 
+function create_jenkins_config_from_templates() {
+    find ${image_config_dir} -type f -name "*.tpl" -print0 | while IFS= read -r -d '' template_path; do
+        local target_path=${template_path%.tpl}
+        if [[ ! -f "${target_path}" ]]; then
+            if [[ "${target_path}" == "${image_config_path}" ]]; then
+                create_jenkins_config_xml
+            elif [[ "${target_path}" == "${image_config_dir}/credentials.xml" ]]; then
+                create_jenkins_credentials_xml
+            else
+                # Allow usage of environment variables in templated files, e.g. ${DOLLAR}MY_VAR is replaced by $MY_VAR
+                DOLLAR='$' envsubst < "${template_path}" > "${target_path}"
+            fi
+        fi
+    done
+}
+
 function install_plugins() {
   # If the INSTALL_PLUGINS variable is populated, then attempt to install
   # those plugins before copying them over to JENKINS_HOME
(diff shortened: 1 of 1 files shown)


gabemontero:
yeah we discussed getting more generic with the substitution when we first introduced it but decided against it ... don't recall all the particulars, but we may have simply been erring on the side of caution.

@bparees is that more or less your recollection?

starting a test run ... see how it does
[test]

bparees:
@gabemontero i don't even recall that much but yeah i don't have an explicit issue with this.

2/contrib/s2i/run
```diff
@@ -41,6 +41,22 @@ function create_jenkins_credentials_xml() {
   fi
 }
 
+function create_jenkins_config_from_templates() {
+    find ${image_config_dir} -type f -name "*.tpl" -print0 | while IFS= read -r -d '' template_path; do
+        local target_path=${template_path%.tpl}
+        if [[ ! -f "${target_path}" ]]; then
+            if [[ "${target_path}" == "${image_config_path}" ]]; then
+                create_jenkins_config_xml
+            elif [[ "${target_path}" == "${image_config_dir}/credentials.xml" ]]; then
```
bparees: this should be looking for credentials.xml.tpl right?

silbernm: no, target_path already should have ".tpl" stripped

silbernm:
@bparees @gabemontero a further option would be to invoke an optional hook script in s2i/run. then the consumers of your image could do all custom replacements in that script.

bparees:
> @bparees @gabemontero a further option would be to invoke an optional hook script in s2i/run. then the consumers of your image could do all custom replacements in that script.

at that point i think i'd just expect someone to fork+replace our run script with their own during the s2i build process.

2/contrib/s2i/run
bparees: ah, right.

bparees:
[merge]

silbernm:
> at that point i think i'd just expect someone to fork+replace our run script with their own during the s2i build process.

true, i didn't even consider this or else i would probably not have opened this pull request. let's hope other people will find it useful as well.

thank you both for your help with the pull request and thanks for feedback!

PR #381, opened 2017-09-18
move centos image to jenkins LTS 2.73.1

bparees:


bparees:
@coreydaley please ensure the rhel image is updated to this level as part of your active card (we'd said to update to 2.60 previously, but we might as well move to 2.73 at this point).

bparees:
@gabemontero fyi

[test]

Issue #390, opened 2017-09-29
Using plugins.txt does not work at Windows if file contains \r\n line ending


ErikLundJensen:
First, the file lock can't be created as the path contains \r:
Creating initial locks...
Locking maven:2.17
Locking
Locking blueocean:1.2.4
Locking
mkdir: cannot create directory '/opt/openshift/plugins/\r.lock': File exists
Failed to install plugins.
Build failed

Secondly, when removing \r from plugins.txt we stil get an error in line 201:

url="$JENKINS_UC_DOWNLOAD/plugins/$plugin/$version/${plugin}.hpi"

It works when ${plugin} is be replaced by $plugin in line 201

However, a solution could be to handle new-line stripping in all config files.

PR #395, opened 2017-10-10
clean up dos file carriage returns

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/390

@openshift/devex fyi

thx for the `bash` assist @bparees 

validated with both dos and unix files

`dos2unix` was not available in the openshift/origin image

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -338,6 +338,11 @@ main() {
 
     mkdir -p "$REF_DIR" || exit 1
 
+    for file in $@; do
+	# clean up any dos file injected carriage returns
+	sed -i 's/\r$//' $file
+    done
+
     # Create lockfile manually before first run to make sure any explicit version set is used.
     echo "Creating initial locks..."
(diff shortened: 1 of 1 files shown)


2/contrib/jenkins/install-plugins.sh
```diff
@@ -338,6 +338,11 @@ main() {
 
     mkdir -p "$REF_DIR" || exit 1
 
+    for file in $@; do
+	echo "Cleaning up file ${file}"
```
bparees: can remove this.

gabemontero:
it does work with empty lines in the file.... not sure what you mean by trailing empty lines

2/contrib/jenkins/install-plugins.sh
gabemontero: echo converted to comment

bparees:
>  not sure what you mean by trailing empty lines

just empty lines at the end of the file vs in the middle.  no reason i'd expect that to be different.

bparees:
lgtm

gabemontero:
[merge]

PR #396, opened 2017-10-11
Fix OpenJDK default settings so as to use less memory.

jerboaa:
- Set CONTAINER_INITIAL_PERCENT to something small by
  default, rather than 100%. Setting it to 100% results in `-Xms==-Xmx`
  which is bad as it tells the GC that it's fine to use at least the max amount.
  Thus it'll never attempt to return pages to the OS which it would otherwise do when
  `-Xms` is set to something small. That's memory other containers could use.
- Use `-Dsun.zip.disableMemoryMapping=true` as this saves some percent
  on memory. The added (performance) cost for class loading is negligible.
- Set `-XX:MinHeapFreeRatio` and `-XX:MaxHeapFreeRatio` to more aggressive values. This together with a low -Xms will make the GC do the right thing.
- Set `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` unconditionally
  as this tells the JVM that physical memory is determined by the cgroup limit (not just the heap). There are some off-heap structures which might get sized wrongly without that option.

This saves about 29% of used VmRSS over the old config for my testing. `388MB` on average for the old settings. `274MB` on average for the new settings. See:
jenkins_plain_old_vs_new

The old settings produced this java command in the container:

    $ java -XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m -Xms256m -Xmx256m -Duser.home=/var/lib/jenkins -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war

The new settings produce this java command in the container:

     $ java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m -Xms17m -Xmx256m -Duser.home=/var/lib/jenkins -Dsun.zip.disableMemoryMapping=true -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war

fixes https://github.com/openshift/jenkins/issues/411

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -138,8 +138,9 @@ if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
 fi
 
 if [ -z "$JAVA_GC_OPTS" ]; then
-  # note - MaxPermSize no longer valid with v8 of the jdk ... used to have -XX:MaxPermSize=100m
-  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
+    # We no longer set MaxMetaspaceSize because the JVM should expand metaspace until it reaches the container limit.
+    # See http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/4dd24f4ca140/src/share/vm/memory/metaspace.cpp#l1470
+    JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
 fi
 
 if [ ! -z "${USE_JAVA_DIAGNOSTICS}" ]; then
@@ -150,8 +151,8 @@ if [ ! -z "${CONTAINER_CORE_LIMIT}" ]; then
     JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMT} -XX:CICompilerCount=2"
 fi
 
-if [ -z "${JVM_ARGS}" ]; then
-    JVM_ARGS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+if [ -z "${JAVA_OPTS}" ]; then
+    JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
 fi
 
 # Since OpenShift runs this Docker image under random user ID, we have to assign
@@ -227,7 +228,7 @@ fi
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JVM_ARGS $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
+   exec java $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
 fi
 
 # As argument is not jenkins, assume user want to run his own process, for sample a `bash` shell to explore this image
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -131,17 +131,15 @@ if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
 
     JAVA_MAX_HEAP_PARAM="-Xmx${CONTAINER_HEAP_MAX}m"
     if [ -z $CONTAINER_INITIAL_PERCENT ]; then
-      # jboss default was 100% or ms==mx
-      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_HEAP_MAX}m"
-    else
-      CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
-      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
+      CONTAINER_INITIAL_PERCENT=0.07
     fi
+    CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+    JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
 fi
 
 if [ -z "$JAVA_GC_OPTS" ]; then
   # note - MaxPermSize no longer valid with v8 of the jdk ... used to have -XX:MaxPermSize=100m
-  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
+  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
 fi
 
 if [ ! -z "${USE_JAVA_DIAGNOSTICS}" ]; then
@@ -225,7 +223,7 @@ fi
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
+   exec java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dsun.zip.disableMemoryMapping=true -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
 fi
 
 # As argument is not jenkins, assume user want to run his own process, for sample a `bash` shell to explore this image
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -150,6 +150,10 @@ if [ ! -z "${CONTAINER_CORE_LIMIT}" ]; then
     JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMT} -XX:CICompilerCount=2"
 fi
 
+if [ -z "${JVM_ARGS}" ]; then
+    JVM_ARGS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+fi
+
 # Since OpenShift runs this Docker image under random user ID, we have to assign
 # the 'jenkins' user name to this UID.
 generate_passwd_file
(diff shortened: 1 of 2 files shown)


jerboaa:
@bparees @gabemontero Please have a look.

2/contrib/s2i/run
```diff
@@ -225,7 +223,7 @@ fi
 
 # if `docker run` first argument start with `--` the user is passing jenkins launcher arguments
 if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
-   exec java $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
+   exec java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dsun.zip.disableMemoryMapping=true -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
```
gabemontero: Yeah we had heard that the experimental options re: cgroups were coming, but they were still not available at that time

Cool deal if it is now - thanks.

gabemontero:
[test]

gabemontero:
@bparees / @jupierce / @smunilla - separate from updating the plugin versions (love that new job :-) )a new jenkins image for rhel will not be created unless we explicitly request it now, right?

@bparees - I'm assuming we want this to be post 3.7, so it can soak through a full release cycle.

gabemontero:
And of course we want the PR test to confirm jenkins with the JDK we currently install comes up OK with these new options.

gabemontero:
@jerboaa - both here and in you associated fabric8 PR, I was unable to infer what kind of traffic or activity was being induced against Jenkins when you were gathering your measurements.

If I just missed it, apologies (and could you give me an explicit pointer?).

Otherwise, could you elaborate?

Among other things, some capacity planning type concerns have emerged from our customer activity.  Things like "I'll run 50 pipeline jobs concurrently, so how big should my heap be?"

Of course, a lot of that *depends on the details of the job* :-)

But we are thinking of at least providing some sort of base example.  Perhaps the particulars of your test runs could be input into that.

thanks

jupierce:
@gabemontero Yes, a new Jenkins version is still manual, but @smunilla is going to be working on a job for it as well.

gabemontero:
On Wed, Oct 11, 2017 at 10:46 AM, Justin Pierce <notifications@github.com>
wrote:

> @gabemontero <https://github.com/gabemontero> Yes, a new Jenkins version
> is still manual, but @smunilla <https://github.com/smunilla> is going to
> be working on a job for it as well.
>

Thanks for the confirmation, and cool deal on the upcoming job

jerboaa:
> @jerboaa - both here and in you associated fabric8 PR, I was unable to infer what kind of traffic or activity was being induced against Jenkins when you were gathering your measurements.
>If I just missed it, apologies (and could you give me an explicit pointer?).
>Otherwise, could you elaborate?
>Among other things, some capacity planning type concerns have emerged from our customer activity. >Things like "I'll run 50 pipeline jobs concurrently, so how big should my heap be?"
>Of course, a lot of that depends on the details of the job :-)
>But we are thinking of at least providing some sort of base example. Perhaps the particulars of your >test runs could be input into that.

The testing for this and for fabric8 was different. For fabric8 I've used 3 jenkins pipeline builds ins succession of the vertx booster example.

For this PR it was manually adding a maven job which built this 3 times in succession:
http://icedtea.classpath.org/mirror/git/thermostat-ng-web-gateway/
Nothing fancy, watched console output of the first one. It's not super-scientific, but it's something. I've tried to be consistent for both runs and results seem repeatable.

I haven't done any experiments as to how the memory should increase per additional concurrent pipeline job. I suppose this can be empirically tested.

2/contrib/s2i/run
bparees: while we're messing with this i think we should ensure that all the JVM args being passed in can be overwritten by the user in case this causes problems.  e.g.:

```
if [ -z JVM_ARGS ]; then
  JVM_ARGS =  -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM -Duser.home=${HOME} $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -Dsun.zip.disableMemoryMapping=true -Dfile.encoding=UTF8
fi

exec java $JVM_ARGS -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS $JENKINS_ACCESSLOG "$@"
```

also not sure what UseCGroupMemoryLimitForHeap accomplishes here since we're explicitly setting the heap size anyway.

jerboaa: OK. I'll make the changes. @bparees There are some off-heap structures of the JVM which get sized correctly with this option. -Xmx is heap only. The JVM is not just heap. I've tried to explain this in the initial PR comment.

jerboaa:
> https://github.com/openshift/jenkins/blob/master/slave-base/contrib/bin/run-jnlp-client#L76-L115 needs the same treatment while we're doing this.

Right. FWIW, slave-maven should get some options for running java with good options too. Can this be done in a separate patch or this one as well?

bparees:
> Right. FWIW, slave-maven should get some options for running java with good options too. Can this be done in a separate patch or this one as well?

slave-maven inherits from slave-base, they should be using the same java invocation to launch the jenkins slave process, so you should only need to change slave-base.

i'd rather do it all in this PR so we have one PR that updates all the JVM settings at once, given the changes should be very similar.

2/contrib/s2i/run
bparees: Sorry, did not read the PR comment closely :(

2/contrib/s2i/run
```diff
@@ -152,6 +150,10 @@ if [ ! -z "${CONTAINER_CORE_LIMIT}" ]; then
     JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMT} -XX:CICompilerCount=2"
 fi
 
+if [ -z "${JVM_ARGS}" ]; then
+    JVM_ARGS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+fi
```
bparees: Looks like you could reuse the JAVA_OPTS env var we already support for this (today it looks like JAVA_OPTS doesn't get set by us, only by the user, so you could make this the default value for it now).

2/contrib/s2i/run
```diff
@@ -131,17 +131,15 @@ if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
...
+      CONTAINER_INITIAL_PERCENT=0.07
     fi
+    CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+    JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
 fi
 
 if [ -z "$JAVA_GC_OPTS" ]; then
   # note - MaxPermSize no longer valid with v8 of the jdk ... used to have -XX:MaxPermSize=100m
-  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
+  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
```
bparees: is the metaspace one of the things that's runed by the cgroup limits setting?  can we get rid of that argument here?  it's been a source of pain as noted by @gabemontero and others.

bparees:
> @bparees - I'm assuming we want this to be post 3.7, so it can soak through a full release cycle.

@gabemontero yeah.

gabemontero:
On Thu, Oct 12, 2017 at 9:11 AM, Ben Parees <notifications@github.com>
wrote:

> *@bparees* requested changes on this pull request.
>
> In 2/contrib/s2i/run
> <https://github.com/openshift/jenkins/pull/396#discussion_r144283697>:
>
> > @@ -152,6 +150,10 @@ if [ ! -z "${CONTAINER_CORE_LIMIT}" ]; then
>      JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMT} -XX:CICompilerCount=2"
>  fi
>
> +if [ -z "${JVM_ARGS}" ]; then
> +    JVM_ARGS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
> +fi
>
> Looks like you could reuse the JAVA_OPTS env var we already support for
> this (today it looks like JAVA_OPTS doesn't get set by us, only by the
> user, so you could make this the default value for it now).
>
> In 2/contrib/s2i/run
> <https://github.com/openshift/jenkins/pull/396#discussion_r144283888>:
>
> >  fi
>
>  if [ -z "$JAVA_GC_OPTS" ]; then
>    # note - MaxPermSize no longer valid with v8 of the jdk ... used to have -XX:MaxPermSize=100m
> -  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
> +  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m"
>
> is the metaspace one of the things that's runed by the cgroup limits
> setting? can we get rid of that argument here? it's been a source of pain
> as noted by @gabemontero <https://github.com/gabemontero> and others.
>
Yeah @jerboaa suggested just that in
https://github.com/fabric8-services/fabric8-tenant-jenkins/pull/30#issuecomment-336090860

I think it is a simply a recommendation we inherited somewhere along the
way ... perhaps it has outlived its usefulness.

2/contrib/s2i/run
jerboaa: I think it makes sense to rely on the cgroup setting to size metaspace as appropriate and let the JVM do appropriate expansions if needed. By explicitly setting the upper bound on the metaspace it's more likely to run into the limit.

+1 on removing it.

2/contrib/s2i/run
jerboaa: Right. But what if existing users have other JAVA_OPTS? If they don't have those options part of their JAVA_OPTS now they would end up with a container java that's not looking at cgroup settings. This new variable would ensure people would get this always. Thoughts?

jerboaa:
> slave-maven inherits from slave-base, they should be using the same java invocation to launch the jenkins slave process, so you should only need to change slave-base.

>i'd rather do it all in this PR so we have one PR that updates all the JVM settings at once, given the changes should be very similar.

I'm not sure. It is my understanding that the maven slave would need to have some `MAVEN_OPTS` set so that forked java processes get the right options.

bparees:
> I'm not sure. It is my understanding that the maven slave would need to have some MAVEN_OPTS set so that forked java processes get the right options.

i'm +1 on doing maven_opts changes in a separate PR, i just meant the slave-base changes should be in this one (and it looks like you've done that).

2/contrib/s2i/run
bparees: yeah i considered that.  my feeling is most people probably don't have JAVA_OPTS today and I really don't want to end up supporting more env variables in this image than we already do (there are already too many and they're not nicely organized, unfortunately).

2/contrib/s2i/run
gabemontero: So we are agreed, nuke it as part of this PR please @jerboaa 

When doing so, if you don't mind, add a comment with something like `we no longer set metaspace becausethe JVM should expand metaspace until it reaches the container limit. See http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/4dd24f4ca140/src/share/vm/memory/metaspace.cpp#l1470 `

bparees:
have we actually confirmed that the JDK we're installing in our image has the `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` options?  (not sure if it'll throw an error if it's an unknown option?)

gabemontero:
It complains if you supply an unknown option.

jerboaa:
> have we actually confirmed that the JDK we're installing in our image has the -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap options? (not sure if it'll throw an error if it's an unknown option?)

Builds `java-1.8.0-openjdk-1.8.0.131-3.b12` and up have that option. Current builds are `1.8.0-144`.

jerboaa:
OK, updated as requested. I've done another run. This time I get ~35% less memory than the old config. `388MB` old, `249MB` new. The truth is probably around 30%.

Old java command:

    $ java -XX:+UseParallelGC -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MaxMetaspaceSize=100m -Xms256m -Xmx256m -Duser.home=/var/lib/jenkins -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war

New java command:

     $ java -XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Xms17m -Xmx256m -Duser.home=/var/lib/jenkins -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true -Dfile.encoding=UTF8 -jar /usr/lib/jenkins/jenkins.war

Graph looked like this:
jenkins_plain_old_vs_new_final

Good to go now?

bparees:
thanks @jerboaa, lgtm.

@gabemontero i think we want to merge this post-3.7.  I'm thinking about creating the 3.7 branch now so we can go ahead and merge this into master, unless you anticipate significant additional changes for 3.7 that will be annoying to backport to two branches?

gabemontero:
On Mon, Oct 16, 2017 at 5:51 AM, Ben Parees <notifications@github.com>
wrote:

> thanks @jerboaa <https://github.com/jerboaa>, lgtm.
>
> @gabemontero <https://github.com/gabemontero> i think we want to merge
> this post-3.7. I'm thinking about creating the 3.7 branch now so we can go
> ahead and merge this into master, unless you anticipate significant
> additional changes for 3.7 that will be annoying to backport to two
> branches?
>
Agreed on post 3.7, and no major changes anticipated.

thanks @jerboaa !

gabemontero:
fyi ... @csrwng just hit a metaspace OOM tuning issue trying to run Jenkins with 64bit and > 4GB container with some CD team work ... I'm having him set JAVA_CG_OPTS to remove our setting of MaxMetaSpace ... so some real use validation of our decision here may be forthcoming soon :-)

csrwng:
For my test, removing the MaxMetaSpace has gotten rid of the Metaspace OOM errors and now Jenkins is functional with the 64-bit JVM (it was not responsive before)

bparees:
[merge]

bparees:
[merge]

jim-minter:
@jerboaa for this workload do you think there is there any benefit in setting -Xms vs leaving it unset?  I'm wondering if it's possible to remove that code in the Jenkins scripts?

jerboaa:
@jim-minter The default of `-Xms` or `-XX:InititialHeapSize` is `1/64` or `1/InitialRAMFraction`. So if you leave it out you'll get roughly `1/64 = 1.5%` for `-Xms`. This patch had it at 7%. I would think it should be fine to remove it.

jim-minter:
Thanks @jerboaa.  Another question: you mention that without `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` "some off-heap structures might get sized wrongly" - please could you provide me an example?

jerboaa:
@jim-minter Examples are Metaspace and card tables of the GC.
`-XX:+UseCGroupMemoryLimitForHeap` tells the JVM to base all of its sizings on the cgroup memory limit. For example if the physical memory is 32 GB and the cgroup limit is 500MB, then the `-XX:+UseCGroupMemoryLimitForHeap` switch will tell the JVM that it's sufficient to size the GC card tables for 500MB of memory, ***not*** 32 GB. Note that you can simulate this behaviour outside of docker via setting` -XX:MaxRAM=n` where n is the container limit.

jim-minter:
Many thanks for the additional detail @jerboaa !

PR #398, opened 2017-10-12
update centos to LTS v2.73.2

bparees:


bparees:
@gabemontero fyi

[merge]

PR #404, opened 2017-10-23
Improve support for Jenkins monitoring plugin

scoheb:
The Jenkins monitoring plugin stores its data in /var/lib/jenkins/monitoring/<hostName>.
Since the pod name changes everytime there is a deployment, any trending data is lost over
re-deployments. We force the application name to allow for historical data collection.


bparees:
[merge]

PR #417, opened 2017-11-10
[WIP] Export JAVA_OPTS and MAVEN_OPTS

jim-minter:
If not in the environment already, export JAVA_OPTS and MAVEN_OPTS with options
that ensure that the JVM respects the cgroup memory size for its heap
calculations.

### DIFF
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -110,6 +110,12 @@ if [[ $# -gt 1 ]]; then
 
   if [ -z "${JAVA_OPTS}" ]; then
       JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+      export JAVA_OPTS
+  fi
+
+  if [ -z "${MAVEN_OPTS}" ]; then
+      MAVEN_OPTS="${JAVA_OPTS}"
+      export MAVEN_OPTS
   fi
 
   echo Running java $JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS $JAVA_OPTS -cp $JAR hudson.remoting.jnlp.Main -headless $PARAMS "$@"
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -153,6 +153,12 @@ fi
 
 if [ -z "${JAVA_OPTS}" ]; then
     JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+    export JAVA_OPTS
+fi
+
+if [ -z "${MAVEN_OPTS}" ]; then
+    MAVEN_OPTS="${JAVA_OPTS}"
+    export MAVEN_OPTS
 fi
 
 # Since OpenShift runs this Docker image under random user ID, we have to assign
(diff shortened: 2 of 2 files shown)


jim-minter:
@bparees @jerboaa @gabemontero going to test this, but would also appreciate your opinions on this.  The intention is to export a small sane JAVA_OPTS such that if the Jenkins slave ends up running Java executables which respect the env var, the heap size calculations will "just work".  Naturally since maven doesn't respect JAVA_OPTS, we do the same for MAVEN_OPTS.

The aim is that anyone who runs gradle (respects JAVA_OPTS) or maven on the slave won't run into trouble with it overallocating and being killed by the OOM killer.

Thoughts?

gabemontero:
@bparees @jerboaa @jim-minter I'm wondering if we should just prepend `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true` to `JAVA_OPTS` and `MAVEN_OPTS` even if they have some existing value associated with them.

If the user has specified some other -XX option that is somehow contradictory to them, they will still override since the last one wins.  But if the user is specifying other parameters unrelated, they still get the benefits these parameters provide.

Thoughts?

That said, I definitely believe @jim-minter is on the right track here.

bparees:
>  I'm wondering if we should just prepend -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true to JAVA_OPTS and MAVEN_OPTS even if they have some existing value associated with them.

As long as -Xmx overrides +UseCGgroupMemoryLimitForHeap it seems like a reasonable suggestion to me, but we need to confirm that behavior, otherwise we leave users w/ no way to opt out of the cgroup heap memory limit.

@jim-minter assuming your testing proves out, it seems like a reasonable step to me.

jerboaa:
> @bparees @jerboaa @jim-minter I'm wondering if we should just prepend -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true to JAVA_OPTS and MAVEN_OPTS even if they have some existing value associated with them.

This makes a lot of sense to me. At least the `UseCGroupMemoryLimitForHeap` option as this tells the JVM that it shouldn't use the physical memory of the host, but rather the cgroup limit for sizing its internal structures, choosing a sane default heap (if not set otherwise), etc. Not sure if some people might want to have `-Dsun.zip.disableMemoryMapping=false`.

jerboaa:
> As long as -Xmx overrides +UseCGgroupMemoryLimitForHeap it seems like a reasonable suggestion to me, but we need to confirm that behavior, otherwise we leave users w/ no way to opt out of the cgroup heap memory limit.

Setting `UseCGgroupMemoryLimitForHeap` tells the JVM that it should fail with OOM once the container limit has been reached. If both options, `+UseCGgroupMemoryLimitForHeap` ***and*** `-Xmx` are set, then instead of choosing the default heap size via ergonomics based on the container limit it will use `-Xmx`.

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -110,6 +110,12 @@ if [[ $# -gt 1 ]]; then
 
   if [ -z "${JAVA_OPTS}" ]; then
       JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+      export JAVA_OPTS
+  fi
+
+  if [ -z "${MAVEN_OPTS}" ]; then
```
jerboaa: Shouldn't this be set in the maven slave? One other point worth raising with this is that sometimes maven builds might fork regular JVMs during tests, those JVMs usually don't get the right options unless set explicitly via surefire or other means. One possible work-around is to set `_JAVA_OPTIONS` an env which the [JVM will pick up](http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/8015f8f73cde/src/share/vm/runtime/arguments.cpp#l2641). A drawback of this is that not all JVM flags can be set with this (e.g. turning on NativeMemoryTracking). Another is that you don't see it on the ps tree, but only via the "Picked up _JAVA_OPTIONS ..." messages.

gabemontero:
I've confirmed @jerboaa statement regarding use of both `+UseCGgroupMemoryLimitForHeap` and `-Xmx`, with `-Xmx` winning.

See https://github.com/openshift/jenkins/issues/418

Also, based on the customer interactions, seems like we should be worrying about `GRADLE_OPTS` as well, no?

slave-base/contrib/bin/run-jnlp-client
gabemontero: Yeah I'm not sure how bullet proof we'll be able to get wrt auto tuning all the maven / gradle permutations.  A simple starting point in our scripts, along with some sort of generic warning in the doc about needing to understand the process model of your usage, etc. so that you can extend images and modify as needed seems more in our sweet spot.

Thoughts @bparees @jim-minter @jerboaa ?

bparees: I have mixed feelings about whether it should be in the maven slave or here, but i think i'm inclined to agree with @jerboaa that it should be in the maven slave.

and yes, whatever we do here needs to be accompanied by better doc/warnings about "when running maven/gradle in a slave, here are settings you need to consider.... and here's what/how we attempt to default them"

jerboaa:
> I've confirmed @jerboaa statement regarding use of both +UseCGgroupMemoryLimitForHeap and -Xmx, with -Xmx winning.

The term "winning" is wrong. Without `-Xmx` and with `+UseCGgroupMemoryLimitForHeap` the default JVM ergonomics will be in place. It's as if you'd run `java` on a system with `$CONTAINER_LIMIT` as physical memory. In that case the JVM will pick some value for `MaxHeapSize` (usually 25% of that). Defaults can be overriden with various flags, -Xmx is one such hook to override the default.

jim-minter:
Closing for https://github.com/openshift/jenkins/pull/442

PR #420, opened 2017-11-15
Allow user to override more configuration than config.xml in the volume

arnaud-deprez:
Hi,

Currently the user can only override config.xml in the volume.
This fix/improvement allows him to override whatever configuration he needs to.

Regards,

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -175,7 +175,7 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
     create_jenkins_config_from_templates
 
     echo "Copying Jenkins configuration to ${JENKINS_HOME} ..."
-    cp -r /opt/openshift/configuration/* ${JENKINS_HOME}
+    cp -r $image_config_dir/* ${JENKINS_HOME}
 
     install_plugins
 
@@ -186,12 +186,22 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
     touch ${JENKINS_HOME}/configured
 else  
   if [ ! -z "${OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG}" ]; then
-    echo "Overriding jenkins config.xml stored in ${JENKINS_HOME}/config.xml"
-    rm -f ${JENKINS_HOME}/config.xml
-	
-    create_jenkins_config_xml
+    echo "Overriding jenkins configuration stored in ${JENKINS_HOME}"
 
-    cp -r ${image_config_path} ${JENKINS_HOME}
+    # remove any files/directories in $JENKINS_HOME that are specified in $image_config_dir, 
+    # otherwise the two will get merged when cp is run below
+    for f in ${image_config_dir}/*; do
+      fileName=${f#${image_config_dir}'/'}
+      if [ -e "${JENKINS_HOME}/${fileName}" ]; then
+        echo "---> Removing ${JENKINS_HOME}/${fileName}..."
+        rm -rf "${JENKINS_HOME}/${fileName}"
+      fi
+    done
+
+    create_jenkins_config_from_templates
+
+    echo "Copying Jenkins configuration to ${JENKINS_HOME}..."
+    cp -r $image_config_dir/* ${JENKINS_HOME}
   fi
     
   if [ ! -z "${OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS}" ]; then


2/contrib/s2i/run
```diff
@@ -186,12 +186,21 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
...
+    echo "Overriding jenkins configuration stored in ${JENKINS_HOME}"
+
+    # When overriding config, we should override every config file except plugins which might be a special case (?)
+    for f in ${image_config_dir}/*; do
+      fileName=${f#${image_config_dir}'/'}
+      if [ -e "${JENKINS_HOME}/${fileName}" ]; then
+        echo "---> Removing ${JENKINS_HOME}/${fileName}..."
+        rm -rf "${JENKINS_HOME}/${fileName}"
+      fi
+    done
```
bparees: is this actually necessary?  seems like the cp is just going to overwrite anyway.

arnaud-deprez: I hesitated to put this.
The cp will overrite existing files but it will not remove existing content in the directory.
Let say for example if we want to rename or delete a job from our git repository, this job will actually remains as cp command will not remove it.

bparees: hm.  but this means if your new config contains even a single job, all existing jobs will be deleted (because the jobs dir will be removed).  Which I guess is what you asked for by asking for the override behavior, but it's a bit scary.

arnaud-deprez: Indeed! 
That’s why I hesitated but I decided to do it that way for these reasons: 
1. Regarding the doc and the override aspect, this seems logical to me.
1. If you look the assemble script, it does the same thing for the _Openshift example_, it deletes the job if other jobs are provided (but indeed it’s only in the image layer here)
1. IMO, once you start configuring jenkins in your image, you should do it for every single configuration to manage your infrastructure as code.

To summarize my opinion, the persistent volume makes sense for the workspace, to keep some build history or for stash/caching but not to store the configuration. It also emphases the immutability of the docker image.

At the end, I’m open to any suggestion, so if you have a better one, please shoot, I would gladly adapt it. 

My goal is to allow user to manage the jenkins configuration in the image, so as long we can do that, I would be happy! :-)

bparees: yeah i think your justification is reasonable.

bparees:
@csrwng @jim-minter @gabemontero thoughts on this?

2/contrib/s2i/run
```diff
@@ -186,12 +186,21 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
     touch ${JENKINS_HOME}/configured
 else  
   if [ ! -z "${OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG}" ]; then
-    echo "Overriding jenkins config.xml stored in ${JENKINS_HOME}/config.xml"
-    rm -f ${JENKINS_HOME}/config.xml
-	
-    create_jenkins_config_xml
+    echo "Overriding jenkins configuration stored in ${JENKINS_HOME}"
+
+    # When overriding config, we should override every config file except plugins which might be a special case (?)
```
jim-minter: # remove any files/directories in $JENKINS_HOME that are specified in $image_config_dir, otherwise the two will get merged when cp is run below

2/contrib/s2i/run
```diff
@@ -186,12 +186,21 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
...
+    for f in ${image_config_dir}/*; do
+      fileName=${f#${image_config_dir}'/'}
+      if [ -e "${JENKINS_HOME}/${fileName}" ]; then
+        echo "---> Removing ${JENKINS_HOME}/${fileName}..."
+        rm -rf "${JENKINS_HOME}/${fileName}"
+      fi
+    done
+
+    create_jenkins_config_from_templates
```
jim-minter: `# re-running install_plugins when OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG is set is not currently implemented`

jim-minter:
I'm in favour.  I think the comments could be clearer and I've suggested alternatives.  Thanks @arnaud-deprez !

gabemontero:
A quick historical perspective - allowing the override of more than config.xml was originally entertained but then dismissed, where the general notion was to be conservative and wait for user requests / use cases.

That said, I'm certainly not opposed to all this.  Just relaying there was an explicit choice made in the past.

arnaud-deprez:
Hi,

@jim-minter ok for your comments I will add it! I also wondering why we don't override plugins here too. The answer was obvious, because there is a second environment variable, but why is that ? Does it make sense ?
Is it what you mean when you say:

> #re-running install_plugins when OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG is set is not currently implemented

Because it's actually implemented just bellow only if you set OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS too.

@gabemontero ok, good to know.

arnaud-deprez:
Justo to add few info, I can certainly: 

- [ ] Move the code from _OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS block_ to _ OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG block_, so basically the overriding becomes the same as the normal installation process.
- [ ] Adapt the doc accordingly.

What do you think ?

bparees:
overriding the plugins should remain separate from overriding the rest of the config.

I don't see what in your OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG block is preventing the plugins from being overridden?

arnaud-deprez:
It does not, just have to set 2 variables instead of 1. But ok let's keep it split.
I'll add the comments tomorrow as my computer is already closed and then we can merge.
Have a nice week-end !

arnaud-deprez:
Can we integrate this now ? Thank you :-)

2/contrib/s2i/run
```diff
@@ -186,12 +186,23 @@ if [ ! -e ${JENKINS_HOME}/configured ]; then
...
+    # otherwise the two will get merged when cp is run below
+    for f in ${image_config_dir}/*; do
+      fileName=${f#${image_config_dir}'/'}
+      if [ -e "${JENKINS_HOME}/${fileName}" ]; then
+        echo "---> Removing ${JENKINS_HOME}/${fileName}..."
+        rm -rf "${JENKINS_HOME}/${fileName}"
+      fi
+    done
+
+    # re-running install_plugins when OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG is set is not currently implemented
```
bparees: this comment is very strange.  this would never be implemented, it's why `OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS` exists.

arnaud-deprez:
@bparees indeed. I forgot to remove it. Done

bparees:
[merge]

Issue #421, opened 2017-11-27
Prevent leakage of zombie processes in jenkins master 


gabemontero:
See https://github.com/openshift/release/issues/432 for details, including research around the core problem and how it was fixed in similar docker based images

bparees:
@jim-minter if you find room for this in your other jenkins tech debt work this sprint, please considering knocking this out too.

I think @smarterclayton was also on board with tini.

gabemontero:
Just started poking around as well ... assuming we only want to install the 64bit version, hopefully running a 32 bit JVM is orthogonal.  Something to include in unit testing.

gabemontero:
Forgot ... `make test` in this repo runs 32 bit mode

PR #422, opened 2017-11-27
employ dumb-init as init process, handle docker zombies

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/421
Fixes https://github.com/openshift/release/issues/432

did not find any yum repos for either centos or rhel

use of entrypoint/cmd combo was per tini doc recommendations ... does that interfere with our s2i extension path ?

putting all of it in CMD directive did not work

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -33,7 +33,8 @@ EXPOSE 8080 50000
 RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.89.2-1.1" && \
+    curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
+    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip dumb-init java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.89.2-1.1" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all  && \
@@ -88,5 +89,5 @@ RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
 VOLUME ["/var/lib/jenkins"]
 
 USER 1001
-ENTRYPOINT []
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["/usr/libexec/s2i/run"]
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -41,7 +41,7 @@ RUN yum-config-manager --disable epel >/dev/null || : && \
     # establishing a symbolic link for that guy as well, so that existing plugins in JENKINS_HOME/plugins pointing to /usr/lib64/jenkins
     # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
     ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
-    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
+    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip dumb-init  java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all  && \
@@ -106,5 +106,5 @@ RUN rm /opt/openshift/base-plugins.txt && \
 VOLUME ["/var/lib/jenkins"]
 
 USER 1001
-ENTRYPOINT []
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["/usr/libexec/s2i/run"]


2/Dockerfile.rhel7
```diff
@@ -41,6 +41,7 @@ RUN yum-config-manager --disable epel >/dev/null || : && \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all  && \
+    curl -fL https://github.com/krallin/tini/releases/download/v0.16.1/tini-static -o /bin/tini && chmod +x /usr/bin/tini && \
```
bparees: can't do this in rhel.  is it packaged?

bparees:
> use of entrypoint/cmd combo was per tini doc recommendations ... does that interfere with our s2i extension path ?

i don't think so since s2i only updates the CMD for the image, as i recall.

but it sounds like using sub-reaping might be a less invasive approach:

bparees:
it seems like we could also leave the ENTRYPOINT alone, change our run script to:

```
#!/bin/sh
exec tini jenkins-startup.sh
```

(tini will become pid 1 due to the exec)

and then "jenkins-startup.sh" would contain what the run script currently contains.

However that approach makes it slightly harder for people using it as an s2i builder to replace the run script.  (they'd have to replace jenkins-startup.sh instead).

2/Dockerfile.rhel7
gabemontero: i could not find a repo to access via yum/dnf, which is why I went this way, but just revisited the release page and it does in fact have an RPM to download

i'll switch to that

gabemontero:
[test]

gabemontero:
pushed the download rpm update ...still pondering the options around how to leverage `tini` from the dockerfile

gabemontero:
@bparees I was unable to install dumb-init via yum with the existing repos for centos/rhel

this is a bit of an indirect inference, but if I read the tea leaves at https://apps.fedoraproject.org/packages/dumb-init correctly, I think dumb-init is "too new", and has not matriculated from fedora down to centos/rhel.

what do you think?

bparees:
@gabemontero it looks like someone is in the process of packaging it for openshift, actually:
https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=626011
(see the rhaos-3.8-rhel-7-candidate tag)

i'd track down Yaakov Selkowitz  and ask what the intent is there in terms of distribution.  Probably loop @jupierce in too.

jupierce:
@gabemontero Looks like dumb-init (1.1.3-12.el7) is in our current 3.8 puddle and shipped in 3.7 errata. Should be ok to use.

gabemontero:
thanks for the updates @jupierce @bparees 

a couple of follow up questions (which will reveal how green I still am wrt packaging and the various distros ;-) ):

1) we should be able to install those on our centos image, right?

2) it is a question of configuring the right repos in order to access the new rpm, right?

3) perhaps our rhel7 / centos7 ami's will be equiped in such a way, right?  For example, I have a rhel7 once up on aws currently with the following listing:

```
-rw-r--r--.  1 root root  283 Dec  3 00:49 centos-paas-sig-openshift-origin13-rpms.repo
-rw-r--r--.  1 root root  283 Dec  3 00:49 centos-paas-sig-openshift-origin14-rpms.repo
-rw-r--r--.  1 root root  283 Dec  3 00:49 centos-paas-sig-openshift-origin15-rpms.repo
-rw-r--r--.  1 root root  283 Dec  3 00:49 centos-paas-sig-openshift-origin36-rpms.repo
-rw-r--r--.  1 root root  283 Dec  3 00:49 centos-paas-sig-openshift-origin37-rpms.repo
-rw-r--r--.  1 root root  392 Oct 12 20:29 charlie.repo
-rw-r--r--.  1 root root  379 Dec  3 00:49 openshift-rhel7-dependencies.repo
-rw-r--r--.  1 root root 2033 Sep 29 17:50 oso-rhui-rhel-server-extras.repo
-rw-r--r--.  1 root root  907 Sep 29 17:50 oso-rhui-rhel-server-releases-optional.repo
-rw-r--r--.  1 root root  823 Sep 29 17:50 oso-rhui-rhel-server-releases.repo
-rw-r--r--.  1 root root  816 Sep 29 17:51 oso-rhui-rhel-server-rhscl.repo
-rw-r--r--.  1 root root  358 Jul 11 16:07 redhat.repo
-rw-r--r--.  1 root root  607 Dec  6 17:06 redhat-rhui-client-config.repo
-rw-r--r--.  1 root root 8679 Dec  6 17:06 redhat-rhui.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.1-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.2-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.3-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.4-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.5-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.6-rpms.repo
-rw-r--r--.  1 root root  807 Dec  3 00:49 rhel-7-server-ose-3.7-rpms.repo
-rw-r--r--.  1 root root   80 Dec  6 17:06 rhui-load-balancers.conf
```

4) If I want to build on my local fedora laptop then, do I need to import some set of the repo files I pasted in 3) to my laptop's /etc/yum.repos.d dir ?

thanks

bparees:
1) no, those are subscription packages.  for centos you'll have to install from elsewhere (from the internet is acceptable though not ideal.)

2) for rhel the repos should already be configured but if not, yes, it would require updates to our CI job.
see: https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/common/test_cases/rhel_base_images.yml#L28

3) same answer

4) yeah but i doubt you'll have access to them from your laptop.

gabemontero:
ok switch to dumb-init pushed (tested locally only for centos ...if rhel fails I have a rhel ami instance ready for local iteration)

grdryn:
> no, those are subscription packages. for centos you'll have to install from elsewhere (from the internet is acceptable though not ideal.)

@gabemontero @bparees I'm not sure if you'd consider it more or less ideal/trustworthy than from the wider internet, but there's a copr repo for dumb-init for el7 here, that you could use for CentOS7 (maybe you're already doing that, I still see only the tini proposal in the changes in this PR): https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/

From what I understand, that's maintained by the same person as the Fedora package that you linked to earlier.

bakito:
I've sucecssfully used the statically-linked binary from https://github.com/Yelp/dumb-init in rhel.

gabemontero:
@bparees @jupierce , quick refresh:  i can install dumb-init for the openshift jenkins rhel image as @jupierce predicted last month. But I have to curl the dumb-init binary for centos currently.  The current repos available to our openshift jenkins centos image build do not have it.

next, @grdryn notes there is a repo (https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/) that might be usable for rpm installing on centos.  If it were acceptable to enable that repo for our centos (is it?), I need to update the repo files in aos-cd-jobs to include that, right?  If so, I'll refresh on what is there and figure out what to update.  Or is there some other path I need to investigate / follow? ... maybe I have to update some of the yum-config-manager invocations in the centos dockerfile?

thanks

bparees:
> next, @grdryn notes there is a repo (https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/) that might be usable for rpm installing on centos. If it were acceptable to enable that repo for our centos (is it?

it is

> I need to update the repo files in aos-cd-jobs to include that, right? If so, I'll refresh on what is there and figure out what to update. Or is there some other path I need to investigate / follow? ... maybe I have to update some of the yum-config-manager invocations in the centos dockerfile?

yes you should be able to just add the repo as part o the dockerfile, like we do for the jenkins rpm repo:
https://github.com/openshift/jenkins/blob/master/2/Dockerfile#L33-L34

gabemontero:
great, thanks @bparees 

unfortunately, I'm getting 503's right now from copr.fedorainfracloud.org ... I'll circle back to this in the near term (later today or tomorrow hopefully) and see if it is up.  If it is up and this is more of a one time hiccup, we'll go with that.  Othewise, though not ideal per @bparees earlier response, we'll go with the curl download of the binary for centos.

gabemontero:
ok doing a yum install on centos as well ... thanks again for the pointers @grdryn and @bparees 

also, my local regression (beyond this repo's PR tests) of launching pipelines that fork processes, and seeing all that look clean when the pipelines complete, looks good.

assuming the tests pass, please revisit as time permits @bparees for concluding if this is the way we want to go.... fyi, there is a customer induced bugzilla on this as well

bparees:
lgtm but let's make sure QE does a solid regression after this goes in.

2/Dockerfile
```diff
@@ -33,6 +33,8 @@ EXPOSE 8080 50000
 RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
+    curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
+    yum install -y dumb-init && \
```
bparees: you can move dumb-init into INSTALL_PKGS

the yum install centos-release-scl-rh is actually installing a repository(I think), which is why it has to be done separately.

gabemontero:
comment addressed ... and yeah, we should be able to drive/monitor QE's testing via the bugzilla process ... I've cc:ed you back on the bugzilla (wasn't sure if you got notified on all image comp bugzillas)

gabemontero:
[merge]

gabemontero:
admin note: ... this was recently reported (after we initially started on this) via customer case being tracked by https://bugzilla.redhat.com/show_bug.cgi?id=1528548

2/Dockerfile
```diff
@@ -33,7 +33,8 @@ EXPOSE 8080 50000
 RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/jenkins.repo && \
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.89.2-1.1" && \
+    curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
```
sferich888: Why are we leveraging EPEL packages for this? 

Why are we not leveraging the Build in RHEL 7 image systemd: https://access.redhat.com/solutions/1118173 
Why are we not using a RHEL supported package?

bparees: you're looking at the centos image's dockerfile, not the rhel image.

Ladicek:
I wonder if you considered doing the same for the Jenkins agent images (resp. the base Jenkins agent image, as the other would inherit it I think)?

bparees:
> I wonder if you considered doing the same for the Jenkins agent images (resp. the base Jenkins agent image, as the other would inherit it I think)?

we don't have any plans to do so..since the agents are not long lived, there's really no concern about accumulating zombie processes.  They'll go away as soon as the agent completes.

PR #425, opened 2017-11-30
rework rhel plugin install from ln to cp; move action from dockerfile…

gabemontero:
… to run script

first step for https://bugzilla.redhat.com/show_bug.cgi?id=1519034

@openshift/sig-developer-experience fyi / ptal

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -37,6 +37,10 @@ EXPOSE 8080 50000
 
 RUN yum-config-manager --disable epel >/dev/null || : && \
     yum-config-manager --enable rhel-7-server-ose-onlineint-rpms || : && \
+    # for backward compatibility with pre-3.6 installs leveraging a PV, where rpm installs went to /usr/lib64/jenkins, we are
+    # establishing a symbolic link for that guy as well, so that existing plugins in JENKINS_HOME/plugins pointing to /usr/lib64/jenkins
+    # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
+    ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
     INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
@@ -51,9 +55,6 @@ ADD ./contrib/s2i /usr/libexec/s2i
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
-    # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
-    # HPI file created by rpm to /opt/openshift/plugins folder.
-    for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
     chmod 775 /etc/passwd && \
     chmod -R 775 /etc/alternatives && \
     chmod -R 775 /var/lib/alternatives && \
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -78,10 +78,15 @@ function install_plugins() {
     /usr/local/bin/install-plugins.sh "${plugins_file}"
   fi
   
-  if [ "$(ls -A /opt/openshift/plugins 2>/dev/null)" ]; then
+  if [ "$(ls /opt/openshift/plugins/* 2>/dev/null)" ]; then
     mkdir -p ${JENKINS_HOME}/plugins
-    echo "Copying $(ls /opt/openshift/plugins | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
-    cp -r /opt/openshift/plugins/* ${JENKINS_HOME}/plugins/
+    echo "Copying $(ls /opt/openshift/plugins/* | wc -l) files to ${JENKINS_HOME} ..."
+    # in case of rhel, if there are plugins in /opt/openshift/plugins, that means the user is overriding
+    # the image's base set of plugins and taking ownership of the plugin's version going forward; and
+    # so we break the link to /usr/lib/jenkins via use of --remove-destination; in the case of centos, it
+    # is not necessary (no links are in play) but has no ill effect so we maintain a common path for
+    # both offerings
+    for FILENAME in /opt/openshift/plugins/* ; do cp --remove-destination $FILENAME ${JENKINS_HOME}/plugins; done
     rm -rf /opt/openshift/plugins
   fi
 }
@@ -169,6 +174,51 @@ new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
 # finish the move of the default logs dir, /var/log/jenkins, to the volume mount
 mkdir ${JENKINS_HOME}/logs
 ln -sf ${JENKINS_HOME}/logs /var/log/jenkins
+
+# clean up any plugins in JENKINS_HOME/plugins which we previously linked, but
+# have since deleted from the image
+for FILENAME in ${JENKINS_HOME}/plugins/*.jpi; do
+    # test if it is a sym link, otherwise users have overriden, so leave alone
+    # test command silent, don't need dev/null
+    test -h $FILENAME
+    if [ $? -eq 0 ]; then
+	stat -L $FILENAME >& /dev/null
+	if [ $? -eq 0 ]; then
+	    continue
+	fi
+	echo "Unlinking plugin ${FILENAME} since it has been removed"
+	unlink $FILENAME
+    fi
+done
+
+# for rhel, some maintenance wrt copying plugins is done on every startup;  this logic
+# covers the following scenarios:
+# 1) this is the initial start of the image and the JENKINS_HOME/configured file is not
+# yet present
+# 2) this is a release upgrade, and new plugins need to be added
+#
+# we employ links instead of copies, and check for the existance in JENKINS_HOME/plugins,
+# so that:
+# 1) we don't copy over plugins the user has overridden since this image first was started
+# (either via this image's extension mechanisms or the jenkins plugin mgr console)
+# 2) if versions of an existing plugin change as a result of a release upgrade, those changes
+# are picked up
+#
+# all of this is moot on centos, since our centos image does not store any plugins in /usr/lib/jenkins
+if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
+    # in case very first time through, make sure plugins dir is there
+    mkdir -p ${JENKINS_HOME}/plugins
+   
+   for FILENAME in /usr/lib/jenkins/*.hpi ; do
+        basefilename=`basename $FILENAME .hpi`
+        plugin_image_volume_location="${JENKINS_HOME}/plugins/${basefilename}.*"
+        if [ "$(ls ${plugin_image_volume_location} 2>/dev/null)" ]; then
+          continue
+        fi
+	echo "Linking ${FILENAME} RPM installed Jenkins plugins to ${JENKINS_HOME} ..."  
+        ln -s  $FILENAME ${JENKINS_HOME}/plugins/${basefilename}.jpi
+   done
+fi
 
 if [ ! -e ${JENKINS_HOME}/configured ]; then
     # This container hasn't been configured yet
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -114,7 +114,8 @@ copy_reference_file() {
             action=${action:-"INSTALLED"}
             log=true
             mkdir -p "$JENKINS_HOME/${dir:23}"
-            cp -r "${f}" "$JENKINS_HOME/${rel}";
+	    # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
+            cp --remove-destination -r "${f}" "$JENKINS_HOME/${rel}";
             # pin plugins on initial copy
             touch "$JENKINS_HOME/${rel}.pinned"
             echo "$image_version" > "$JENKINS_HOME/${version_marker}"
@@ -128,7 +129,8 @@ copy_reference_file() {
             action="INSTALLED"
             log=true
             mkdir -p "$JENKINS_HOME/${dir:23}"
-            cp -r "${f}" "$JENKINS_HOME/${rel}";
+	    # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
+            cp --remove-destination -r "${f}" "$JENKINS_HOME/${rel}";
         else
             action="SKIPPED"
         fi


gabemontero:
[test]

2/contrib/s2i/run
```diff
@@ -60,6 +60,18 @@ function create_jenkins_config_from_templates() {
 }
 
 function install_plugins() {
+  mkdir -p ${JENKINS_HOME}/plugins
+    
+  if [ "$(ls -A /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
+      # found rhel rpm installed plugins in the 3.7 or later location
+      echo "Copying $(ls /usr/lib/jenkins/*.hpi | wc -l) RHEL RPM installed Jenkins plugins to ${JENKINS_HOME} ..."
```
bparees: remove the references to RHEL since this script is shared w/ the centos image.

2/contrib/s2i/run
```diff
@@ -60,6 +60,18 @@ function create_jenkins_config_from_templates() {
...
+      for FILENAME in /usr/lib/jenkins/*hpi ; do cp $FILENAME ${JENKINS_HOME}/plugins/`basename $FILENAME .hpi`.jpi; done
+  elif [ "$(ls -A /usr/lib64/jenkins/*.hpi 2>/dev/null)" ]; then
```
bparees: i don't think this should be an elif, we should just do both.

gabemontero: ok, if both though, I'm giving the newer flavor precedence (i.e. executed afterward to override whatever was done previously)

2/contrib/s2i/run
gabemontero: this path though will only execute on rhel ... there will be no `/usr/lib/jenkins/*.hpi` files in the centos image

bparees: unless someone has extended the image and installed their own plugins from rpms on a centos image, but is still using our run script....

2/contrib/s2i/run
bparees: what are we considering the "new" flavor?  lib64?

gabemontero: old == lib64, new == lib

2/contrib/s2i/run
gabemontero: ok i'm converted removing "RHEL"

2/contrib/s2i/run
bparees: ok. I guess i don't particularly care either way.

gabemontero:
Hmmm ... test passed, but I'm not seeing messages I was expecting .... investigating

gabemontero:
OK it looks like we don't always dump the container logs in the test; think I'm going to add that, even if we decide it should be temporary, just to confirm some of the changes in RHEL

gabemontero:
I am wondering if we should be considering flavors like https://stackoverflow.com/questions/8377312/how-to-convert-symlink-to-regular-file in the new use of `cp`

gabemontero:
ok with https://github.com/openshift/jenkins/pull/425/commits/3b17bd5f339eda20fce326d1d0b211f5ea003182 in place I was able to confirm that the new copy path in the `run` script worked as expected

I'm going to circle back and try/test the "link from the prior release exists, and now we are trying to copy" scenario.

gabemontero:
I think I've overthought / got turned around wrt https://github.com/openshift/jenkins/pull/425#issuecomment-348302460 

That said, I'm still trying to reconcile some details wrt the pod logs provided from the customer and the `run` script, now that I've revisited it in detail.  Going to do some more local testing, as well as probably ping @bparees and @jim-minter to cross reference a couple of details.

gabemontero:
OK, yeah, this change as is not sufficient.
1) I do need `--remove-destination`, otherwise I get `cp: not writing through dangling symlink ...` ; you get that with that sequence the follows from installing pre 3.7, then moving up to 3.7, on a PV
  - ln -s filea fileb
  - cp fileb filec  (filec is on the PV)
  - remove filea (i.e. the /usr/lib64 file ... it goes away with the new image)
  - cp newfilea (i.e. the /usr/lib file) filec (filec on the PV is still linked to filea)
2) `install_plugins` will not get installed on the PV based upgrade, cause `${JENKINS_HOME}/configured` is present ... that is why I did not see any of the expected `echo` statements around copying files that would have indicated as such
3) the error occurs because the `cp -r /opt/openshift/plugins/* ${JENKINS_HOME}/plugins/` call from install_plugins from the first startup preserves the link to `/usr/lib64` (vs. copying the file)

So we either 
a) do the `cp --remove-destination` outside the `${JENKINS_HOME}/configured` check ... make sure we don't break various scenarios where the customer can override the plugin list
b) follow @jupierce 's suggestion of simply linking /usr/lib64/jenkins -> /usr/lib/jenkins in the Dockerfile

I'm beginning to lean toward b), but want to walk through the permutations to be sure ...

gabemontero:
`ln -s /usr/lib/jenkins /usr/lib64/jenkins` would seem to work ...a local approximation:

```
gmontero /tmp/ln-test $ mkdir usr usr64 opt var 
gmontero /tmp/ln-test $ cp ~/Downloads/openshift-client-102.hpi usr64
gmontero /tmp/ln-test $ ln -s /tmp/ln-test/usr64/openshift-client-102.hpi /tmp/ln-test/var/plugin.jpi
gmontero /tmp/ln-test $ ls -la var
total 0
drwxrwxr-x. 2 gmontero gmontero  60 Nov 30 16:18 .
drwxrwxr-x. 6 gmontero gmontero 120 Nov 30 16:18 ..
lrwxrwxrwx. 1 gmontero gmontero  43 Nov 30 16:18 plugin.jpi -> /tmp/ln-test/usr64/openshift-client-102.hpi
gmontero /tmp/ln-test $ cp ~/Downloads/openshift-client-102.hpi usr
gmontero /tmp/ln-test $ rm -rf usr64
gmontero /tmp/ln-test $ ln -s /tmp/ln-test/usr /tmp/ln-test/usr64
gmontero /tmp/ln-test $ ls -la 
total 0
drwxrwxr-x.  5 gmontero gmontero 120 Nov 30 16:20 .
drwxrwxrwt. 24 root     root     500 Nov 30 15:37 ..
drwxrwxr-x.  2 gmontero gmontero  40 Nov 30 16:18 opt
drwxrwxr-x.  2 gmontero gmontero  60 Nov 30 16:19 usr
lrwxrwxrwx.  1 gmontero gmontero  16 Nov 30 16:20 usr64 -> /tmp/ln-test/usr
drwxrwxr-x.  2 gmontero gmontero  60 Nov 30 16:18 var
gmontero /tmp/ln-test $ ls -la var
total 0
drwxrwxr-x. 2 gmontero gmontero  60 Nov 30 16:18 .
drwxrwxr-x. 5 gmontero gmontero 120 Nov 30 16:20 ..
lrwxrwxrwx. 1 gmontero gmontero  43 Nov 30 16:18 plugin.jpi -> /tmp/ln-test/usr64/openshift-client-102.hpi
gmontero /tmp/ln-test $ jar tvf var/plugin.jpi 
     0 Tue Nov 07 12:31:36 EST 2017 META-INF/
   881 Tue Nov 07 12:31:34 EST 2017 META-INF/MANIFEST.MF
     0 Tue Nov 07 12:31:34 EST 2017 WEB-INF/
     0 Tue Nov 07 12:31:34 EST 2017 WEB-INF/lib/
     0 Tue Nov 07 12:31:34 EST 2017 BaseStep/
   660 Tue Nov 07 12:31:34 EST 2017 WEB-INF/licenses.xml
110935 Tue Nov 07 12:31:34 EST 2017 WEB-INF/lib/openshift-client.jar
   132 Tue Nov 07 12:31:34 EST 2017 BaseStep/help-credentials.html
    61 Tue Nov 07 12:31:34 EST 2017 BaseStep/help-logLevel.html
    97 Tue Nov 07 12:31:34 EST 2017 BaseStep/help-advancedArguments.html
   121 Tue Nov 07 12:31:34 EST 2017 BaseStep/help-project.html
    64 Tue Nov 07 12:31:34 EST 2017 BaseStep/help-cluster.html
     0 Tue Nov 07 12:31:36 EST 2017 META-INF/maven/
     0 Tue Nov 07 12:31:36 EST 2017 META-INF/maven/com.openshift.jenkins.plugins/
     0 Tue Nov 07 12:31:36 EST 2017 META-INF/maven/com.openshift.jenkins.plugins/openshift-client/
  4985 Tue Nov 07 12:31:16 EST 2017 META-INF/maven/com.openshift.jenkins.plugins/openshift-client/pom.xml
   130 Tue Nov 07 12:31:34 EST 2017 META-INF/maven/com.openshift.jenkins.plugins/openshift-client/pom.properties
gmontero /tmp/ln-test $ 
```

bparees:
i'm fine w/ symlinking /usr/lib64/jenkins to /usr/lib/jenkins.

gabemontero:
had typo in docker file ... pushing new commit momentarily

gabemontero:
per previous sign off from @bparees merging

[merge]

2/Dockerfile.rhel7
```diff
@@ -54,6 +54,10 @@ RUN rm /opt/openshift/base-plugins.txt && \
     # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
     # HPI file created by rpm to /opt/openshift/plugins folder.
     for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
+    # for backward compatability with pre-3.6 installs leveraging a PV, where rpm installs went to /usr/lib64/jenkins, we are
+    # establishing a symbolic link for that guy as well, so that existing plugins in JENKINS_HOME/plugins pointing to /usr/lib64/jenkins
+    # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
+    ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
```
bparees: you should create this symlink *before* installing the rpms, so that any rpm that installs to /usr/lib64/jenkins lands them in the symlinked dir.

2/Dockerfile.rhel7
```diff
@@ -54,6 +54,10 @@ RUN rm /opt/openshift/base-plugins.txt && \
     # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
     # HPI file created by rpm to /opt/openshift/plugins folder.
     for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
+    # for backward compatability with pre-3.6 installs leveraging a PV, where rpm installs went to /usr/lib64/jenkins, we are
```
bparees: compatibility

bparees:
I'm reminded that another reason to prefer copying things is if we ship a 3.8 image that doesn't even install a particular plugin that we previously shipped in 3.7, the PV will still have a broken symlink to that plugin.

bparees:
so i think we should do both.  create the symlink, but also going forward, copy the plugins to the target dir instead of symlinking them, for fresh installs.

bparees:
(aborted the merge)

gabemontero:
doing both, fixed order, pushed

2/contrib/s2i/run
```diff
@@ -79,7 +87,6 @@ function install_plugins() {
   fi
   
   if [ "$(ls -A /opt/openshift/plugins 2>/dev/null)" ]; then
-    mkdir -p ${JENKINS_HOME}/plugins
     echo "Copying $(ls /opt/openshift/plugins | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
     cp -r /opt/openshift/plugins/* ${JENKINS_HOME}/plugins/
```
bparees: doesn't this also need a --remove-destination?

2/Dockerfile.rhel7
```diff
@@ -51,6 +51,10 @@ ADD ./contrib/s2i /usr/libexec/s2i
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
+    # for backward compatability with pre-3.6 installs leveraging a PV, where rpm installs went to /usr/lib64/jenkins, we are
+    # establishing a symbolic link for that guy as well, so that existing plugins in JENKINS_HOME/plugins pointing to /usr/lib64/jenkins
+    # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
+    ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
     # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
     # HPI file created by rpm to /opt/openshift/plugins folder.
     for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
```
bparees: we don't need to create these symlinks in the image anymore, right?

the /var/lib/jenkins symlinks on a PV will be pointing to /usr/lib*/jenkins and the new initialization logic will copy from /usr/lib*, right?  So nothing should need the /opt/openshift/plugins content.

gabemontero: too much multi tasking ... totally mucked up the change ... updating shortly

2/contrib/s2i/run
gabemontero: this is the vanilla first install path ... JENKINS_HOM/plugins is empty

gabemontero: and in reality, this is centos specific now, and independent of all the rhel symlink history

gabemontero:
ok - dockerfile update, no longer symlink specific files, pushed

2/contrib/s2i/run
bparees: this function is also called when OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS is set.

bparees: so it would be called in both a rhel and centos image when that variable is set and install_plugins.sh put some plugin files in /opt/openshift/plugins

gabemontero: so if they override an rpm installed plugin (which could have symlink) via this mechanism, they could need --remove-destination; otherwise, they cp will just overwrite

which we don't prevent, so sure (maybe we should prevent ... pretend I didn't say that)

this would be an existing bug then, which nobody has reported yet ... but sure, I'll switch to the other cp mechanism

gabemontero: though the suffix should be jpi since install_plugins.sh is used, so slightly different

bparees: yeah it's an existing bug.

gabemontero:
new cp pushed

gabemontero:
[test]

2/contrib/s2i/run
```diff
@@ -79,9 +87,9 @@ function install_plugins() {
   fi
   
   if [ "$(ls -A /opt/openshift/plugins 2>/dev/null)" ]; then
-    mkdir -p ${JENKINS_HOME}/plugins
     echo "Copying $(ls /opt/openshift/plugins | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
-    cp -r /opt/openshift/plugins/* ${JENKINS_HOME}/plugins/
+    # since install_plugins populates this, the files already end in jpi
+    for FILENAME in /opt/openshift/plugins/* ; do cp --remove-destination $FILENAME ${JENKINS_HOME}/plugins; done
```
bparees: couldn't still use cp -r?

2/Dockerfile.rhel7
```diff
@@ -51,9 +51,10 @@ ADD ./contrib/s2i /usr/libexec/s2i
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
-    # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
-    # HPI file created by rpm to /opt/openshift/plugins folder.
-    for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
+    # for backward compatibility with pre-3.6 installs leveraging a PV, where rpm installs went to /usr/lib64/jenkins, we are
+    # establishing a symbolic link for that guy as well, so that existing plugins in JENKINS_HOME/plugins pointing to /usr/lib64/jenkins
+    # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
+    ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
```
bparees: I still think this needs to happen before we yum install anything, to avoid the scenario where a package creates this dir(/usr/lib64/jenkins) and then we can't create the symlink.

2/contrib/s2i/run
gabemontero: it occurred to me that there could be non-jpi files here (leftover lock files); decided to be more precise, thinking any overhead from a few more `cp` calls is neglible

2/Dockerfile.rhel7
gabemontero: ok i misinterpreted the earlier comment then (thinking it only pertained to this run block)

moving above

gabemontero:
move on ln pushed

bparees:
this lgtm. @jim-minter can you please give it a once over as well?

2/contrib/s2i/run
```diff
@@ -60,6 +60,14 @@ function create_jenkins_config_from_templates() {
 }
 
 function install_plugins() {
+  mkdir -p ${JENKINS_HOME}/plugins
+  
+  # with the /usr/lib64/jenkins -> /usr/lib/jenkins symlink, don't need to check both directories
+  if [ "$(ls -A /usr/lib/jenkins 2>/dev/null)" ]; then
```
jim-minter: nits: the ls -A selects a different set of files to the $(ls in the next line, which is different again to the /usr/lib/jenkins/*hpi below that. Copying *hpi doesn't guarantee that basename ... .hpi will work.  I'd be tempted to:
1) remove the if statement
2) echo "Copying $(ls /usr/lib/jenkins/*.hpi | wc -l) ...
3) for FILENAME in /usr/lib/jenkins/*.hpi ...

2/contrib/s2i/run
jim-minter: Same comments as made above apply here.  Given the move from cp -r, did you mean to do do FILENAME in .../*.jpi here?

2/contrib/s2i/run
```diff
@@ -60,6 +60,14 @@ function create_jenkins_config_from_templates() {
...
+      echo "Copying $(ls /usr/lib/jenkins | wc -l) RPM installed Jenkins plugins to ${JENKINS_HOME} ..."
+      for FILENAME in /usr/lib/jenkins/*hpi ; do cp --remove-destination $FILENAME ${JENKINS_HOME}/plugins/`basename $FILENAME .hpi`.jpi; done
```
jim-minter: I don't feel like I understand why --remove-destination is used here, especially as we're doing the lib/lib64 symlink now.  A code comment would be handy.

2/contrib/s2i/run
gabemontero: yeah, about to contradict myself here ... i moved to single file cp to avoid unavoided things, but now realize that at the time avoided the prefix in case we switched from jpi to hpi

it should be jpi, based on how the download works

I'll make it specific - thanks for the catch

2/contrib/s2i/run
gabemontero: it stemmed from my testing, and when files in JENKINS_HOME/plugins are links to begin with i.e. we have a PV and this is an upgrade scenario - the new sym link we are establishing has no bearing - if you copy a file to a new destination, where that destination is a link to another file, you get `cp: not writing through dangling symlink ...` ... see https://github.com/openshift/jenkins/pull/425#issuecomment-348321117 and https://github.com/openshift/jenkins/pull/425#issuecomment-348325086

I'll add a comment to try and illustrate / clarify

2/contrib/s2i/run
gabemontero: In your 3), the crux of what you are saying is to do `*.hpi` instead of `*hpi`, right?

gabemontero: also, you've got me wondering now if we should ditch the `ls -A` check around `/opt/openshift/plugins` 

I guess the general question is will the for loop blow up if the file expr supplied to the `in` clause does not work.

Will try an experiment locally and report back.

jim-minter: sorry - I guess the if check is needed, but it might make more sense if it were if [ "$(ls /usr/lib/jenkins/*.hpi ...

jim-minter: `*.hpi` instead of `*hpi`, yes.

gabemontero: the answer is that it blows up:

```
gmontero ~ $ ls -la /usr/lib/jenkins
ls: cannot access '/usr/lib/jenkins': No such file or directory
gmontero ~ $ for FILENAME in /usr/lib/jenkins/*.hpi ; do ls -la $FILENAME; done
ls: cannot access '/usr/lib/jenkins/*.hpi': No such file or directory
gmontero ~ $ which stat
/usr/bin/stat
gmontero ~ $ for FILENAME in /usr/lib/jenkins/*.hpi ; do cp --remove-destination $FILENAME ${JENKINS_HOME}/plugins/`basename $FILENAME .hpi`.jpi; done
cp: cannot stat '/usr/lib/jenkins/*.hpi': No such file or directory
gmontero ~ $ 
```

It is not a given that `/usr/lib/jenkins` or `/opt/openshift/plugins` exists.

Rather than using `ls` in the if, how about the use of what is described in https://stackoverflow.com/questions/59838/check-if-a-directory-exists-in-a-shell-script

gabemontero: actually based on running jenkins img tests locally, going to switch to https://stackoverflow.com/questions/638975/how-do-i-tell-if-a-regular-file-does-not-exist-in-bash and employ `-f` on `/usr/lib/jenkins/*.hpi`

gabemontero: check that  ... `-d` results still ok, though error message which is ignored, `-f` with wildcards not so much

back to exploring ls with * but no -A

gabemontero:
update for @jim-minter comments pushed

jim-minter:
all lgtm, thanks @gabemontero

bparees:
lgtm, merge at will (not sure if you're doing any additional offline testing... at a minimum you should make sure you can "update" a 3.6 jenkins to this image, when using a PV since that was the point of the exercise.

gabemontero:
thanks @bparees @jim-minter 

unfortunately, the test failure may be legitimate ... s2i test failed; my presumption to not copy files other than `.jpi` from `/opt/openshift/plugins` has proven incorrent (now, I thought I ran the full test suite locally, but seems something got lost in the multi-task shuffle).

New update coming, where we can revisit yet again my love for hash :-(

And +1 on manual upgrade test @bparees

gabemontero:
my feelings on the multitude of ways to extended the jenkins image are not the best right now either ;-)

gabemontero:
ahh ... i did not see the s2i failure because I did not have an s2i binary, but I only caught "SUCCESS" at the end, as the `make test ...` still passes

```
which: no s2i in (/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/home/gmontero/go/bin:/usr/local/go/bin:/usr/local/ec2/ec2-api-tools-1.7.4.0/bin:/home/gmontero/go/src/github.com/openshift/origin/_output/local/bin/linux/amd64:/home/gmontero/go/src/github.com/openshift/source-to-image/_output/local/bin/linux/amd64:/home/gmontero/etcd-v2.1.1-linux-amd64:/home/gmontero/apache-maven-3.3.3/bin:/home/gmontero/protoc/bin:/home/gmontero/.local/bin:/home/gmontero/bin:/home/gmontero/go/bin:/usr/local/go/bin:/usr/local/ec2/ec2-api-tools-1.7.4.0/bin:/home/gmontero/go/src/github.com/openshift/origin/_output/local/bin/linux/amd64:/home/gmontero/go/src/github.com/openshift/source-to-image/_output/local/bin/linux/amd64:/home/gmontero/etcd-v2.1.1-linux-amd64:/home/gmontero/apache-maven-3.3.3/bin:/home/gmontero/protoc/bin)
WARNING: Please install the s2i binary to run the s2i tests.
SUCCESS!
Stopping and removing container 0811d79e99ed8946d6ea29e6b6b69342ded79624f35c155b15c66b8f6138274c...
0811d79e99ed8946d6ea29e6b6b69342ded79624f35c155b15c66b8f6138274c
0811d79e99ed8946d6ea29e6b6b69342ded79624f35c155b15c66b8f6138274c
Done.
```

bparees:
> my feelings on the multitude of ways to extended the jenkins image are not the best right now either ;-)

yeah.  Now perhaps you (and more so @csrwng) understand why I'm really risk averse whenever we want to add one more codepath hack for use case foo.

gabemontero:
adjustment for s2i scenario pushed (in separate commit to facilitate review)

will squash if everyone is happy

moving on to constructing PV upgrade scenario

2/contrib/s2i/run
```diff
@@ -131,16 +146,17 @@ if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
 
     JAVA_MAX_HEAP_PARAM="-Xmx${CONTAINER_HEAP_MAX}m"
     if [ -z $CONTAINER_INITIAL_PERCENT ]; then
-      CONTAINER_INITIAL_PERCENT=0.07
+      # jboss default was 100% or ms==mx
+      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_HEAP_MAX}m"
+    else
+      CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
```
bparees: what's all this?

gabemontero:
On Fri, Dec 1, 2017 at 2:56 PM Ben Parees <notifications@github.com> wrote:

>
> In 2/contrib/s2i/run
> <https://github.com/openshift/jenkins/pull/425#discussion_r154435035>:
>
> > @@ -131,16 +146,17 @@ if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
>
>      JAVA_MAX_HEAP_PARAM="-Xmx${CONTAINER_HEAP_MAX}m"
>      if [ -z $CONTAINER_INITIAL_PERCENT ]; then
> -      CONTAINER_INITIAL_PERCENT=0.07
> +      # jboss default was 100% or ms==mx
> +      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_HEAP_MAX}m"
> +    else
> +      CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
> +      JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
>
> what's all this?
>
made the change on the wrong version - redoing

gabemontero:
ok wrong branch snafu fixed ... s2i commit has the change to copy all files (including those with the pinned suffix)

(as the comment notes, other elements there around count, general check, looks at .jpi still)

2/contrib/s2i/run
```diff
@@ -78,10 +90,13 @@ function install_plugins() {
     /usr/local/bin/install-plugins.sh "${plugins_file}"
   fi
   
-  if [ "$(ls -A /opt/openshift/plugins 2>/dev/null)" ]; then
-    mkdir -p ${JENKINS_HOME}/plugins
-    echo "Copying $(ls /opt/openshift/plugins | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
-    cp -r /opt/openshift/plugins/* ${JENKINS_HOME}/plugins/
+  if [ "$(ls /opt/openshift/plugins/*.jpi 2>/dev/null)" ]; then
```
jim-minter: Sorry for the additional comment - now below is not *.jpi, should that also be the case here and in the echo line?

gabemontero: I made an explicit choice to not do that.  Only go into the block if there are *.jpi, and only take the total number of files the *.jpi ls returns for the echo citing the number of plugins, but when we actually copy, in addition to *.jpi, copy any associated pinned files.

The comment below starting with "NOTE" tried to convey that ... was the text there not sufficient?

jim-minter: Got it.  I'll stop poking - sorry :)

2/contrib/s2i/run
```diff
@@ -78,10 +90,13 @@ function install_plugins() {
...
+    echo "Copying $(ls /opt/openshift/plugins/*.jpi | wc -l) Jenkins plugins to ${JENKINS_HOME} ..."
+    # since install_plugins populates this, the files already end in jpi, don't need to swap suffixes
```
bparees: well, it's not just install_plugins that populates it:
https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/assemble#L3
https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/assemble#L25

so if someone does an s2i build using this as the builder image, and they have a .hpi file in their source plugins dir, it's going to get lost here.

bparees: seems safer to just continue to check if plugins is empty or not.

bparees: what was the motivation to narrow the check?

2/contrib/s2i/run
gabemontero: no worries :-)

2/contrib/s2i/run
gabemontero: simply trying to prevent unintended side effects from unintended copying of files

given the assemble discovery, I'm retracting ... other than incorporating the `-A` comment from @jim-minter I'll just copy everything

if something pops up from copying something we shouldn't we'll handle then

gabemontero: adjusting the echo as well

bparees:
lgtm pending your successful migration test.

gabemontero:
I squashed the commits and pushed; will update the 3.7 branch version momentarily.

I'm hoping to get to the ec2 / rhel based migration scenario testing either later this evening or Sunday.

gabemontero:
ok docker.io/gmontero/jenkins-2-rhel7-test:latest is ready ... going to pull it down locally and give it a go

gabemontero:
the migration test failed ... though we talked about install_plugins not getting called once the configured file is present, and I'm pretty sure we had a prior incarnation with the copies outside in install_plugins, this current incarnation does not.

so what happens is the existing plugins in /var/lib/jenkins/plugins point to the latest plugins, but any new dependencies for existing plugins where their versions changed between 3.6 and 3.7 are not copied.  existing plugins at the later versions then complain about dependency plugins missing

*ALWAYS* copying the /usr/lib/jenkins plugins (i.e. outside of install_plugins and the is configured file present block) *MAY* be ok if it is done prior to that section, so that any overrides following are then copied ... would like to ponder more at a more decent hour.  But what about extension images that change existing plugins, but then call our run script as the entry point.

Though a bit more complicated, I'm leaning toward perhaps *ALWAYS* copying plugins from /usr/lib/jenkins that *DO NOT* exist in /var/lib/jenkins/plugins.

gabemontero:
The "ALWAYS copying plugins from /usr/lib/jenkins that DO NOT exist in /var/lib/jenkins/plugins." passed the migration test.

Will push up latest commit (non-squashed) in a couple

gabemontero:
change to make actual upgrade scenario work pushed

2/contrib/s2i/run
```diff
@@ -181,6 +181,19 @@ new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
...
+   for FILENAME in /usr/lib/jenkins/*.hpi ; do
+        basefilename=`basename $FILENAME .hpi`
+        otherfile="${JENKINS_HOME}/plugins/${basefilename}.*"
+        if [ "$(ls ${otherfile} 2>/dev/null)" ]; then
+          continue
+        fi
+	echo "Copying ${FILENAME} RPM installed Jenkins plugins to ${JENKINS_HOME} ..."  
+        cp --remove-destination $FILENAME ${JENKINS_HOME}/plugins/${basefilename}.jpi
+   done
+fi
```
bparees: I wonder if we should just always copy, period.  Otherwise you risk not updating a dependent plugin to the necessary version if the version changed in the new image.

bparees: The obvious downside would be users have no way to remove a plugin, or update it to a newer version than what's in the image.  (the former doesn't bother me, but the latter certainly does)

gabemontero: Yes the later is what I tried to convey in https://github.com/openshift/jenkins/pull/425#issuecomment-348742162

The update of an existing dependency to a new version will work as things are currently constructed because we have links at the moment from /var/lib/jenkins/plugins to /usr/lib/jenkins.

But on subsequent upgrades that will break down, as we've started copying.  The next time a dependencies version needs to increase, it would no longer be correct.  So linking instead of copying helps in this case.  The "master", "original" version is in /usr/lib/jenkins.

So, yes, implies moving back to links from /var/lib/jenkins/plugins to /usr/lib/jenkins ...

Now, if the user wants to override, they go through the various paths to install-plugins.sh, then they break the links, but they have chose to manage things on their own and that point.

And all this just made me realize that we need to update install-plugins.sh and its use of `cp` to include the use of `--remove-destination` to probably override a link .... 

... sign, this is never going to end, is it.

bparees: with symlinks, presumably if they update a plugin version (Via the console), that will update the .jpi file in /usr/lib (where the symlink points).  Since that's not a PV location, the next time jenkins restarted, they'd lose the update anyway.  (this is a problem in the existing implementation).

Basically I think the only safe thing to do is copy the plugins, and only do it the first time a new version of the image is run against a given /var/lib/jenkins dir.  Or do a timestamp check and don't overwrite files that are newer than what's in the image.

bparees: I guess timestamps won't work since:

image1 contains f1.a at t1
image2 is shipped with f1.c at t2
user updates f1.a to f1.b at t3
user picks up image2, but f1.c won't overwrite f1.b because it's not "newer"

gabemontero: No, updates via the console will update the .jpi file in /var/lib/jenkins/plugins ... it will break the link to /usr/lib.  The jenkins home that jenkins works off of is set to the PV.

If we could somehow determine that we are in an upgrade scenario (I'm guessing your timestamp exploration was an attempt at that), then yeah, we could process things differently.

gabemontero: Confirmed via manual test that it breaks the link when upgrading a plugin via the jenkins console

gabemontero: A back up file exists in fact that still has the link ... so if you were to revert, things would work as expected.

bparees: ok so where does this leave us? Preferring to create links in the PV, but only if there's no non-link file already in the PV?

gabemontero: yes, i believe that is what we should do .... I'll get started on the rework

gabemontero:
link based overhaul pushed (single commit)

embarking on migration testing while the repo test starts up here - will report back here on migration test status

gabemontero:
while the test passesd for me locally for centos, I'm already seeing an issue for the rhel side on my ec2 instance:

```
Linking /usr/lib/jenkins/ace-editor.hpi RPM installed Jenkins plugins to /var/lib/jenkins ...
ln: failed to create symbolic link ‘/var/lib/jenkins/plugins/ace-editor.jpi’: No such file or directory
```

not sure it makes sense to me off the cuff ... will debug there while this PR test continues

gabemontero:
I see ... when starting up the image via docker the first time the volume exists, but it is empty... need to create the plugins dir.

gabemontero:
OK, rhel image start up looks good with the latest commit.

The migration scenario "worked" in that the image started up.  The /usr/lib64/jenkins -> to /usr/lib/jenkins link is there.  All the plugins, including the new ones introduced by 3.7 are in /var/lib/jenkins/plugins, but there are not sym links.  They are actual files.  i.e. the sym link was lost in all the machinations.

Given the new "keep the links" mantra, investigating.

gabemontero:
Must have did something wrong before.

With the same image, I brought up on a PV from scratch, are the links from /var/lib/jenkins/plugins/*.jpi to /usr/lib/jenkins/*.hpi are there.

I then added some debug, but now in both the "from scratch" and "migration" scenarios, the links are there.

Going to remove the debug just to make sure it did not have some unintended side effect.

gabemontero:
Yep ran a few times both ways (migration and from scratch) with a rhel image from this commit only ... links look good.

@bparees - if we are good with the implementation specifics, this is merge-able.

bparees:
So let me attempt to summarize the behavior:

 * plugins baked into the image(in /usr/lib*/jenkins) are, at startup, created as symlinks in /var/lib/jenkins/plugins unless the file already exists in /var/lib/jenkins/plugins

 * content installed by the user(via the jenkins console or the various image plugin install mechanisms) is copied into /var/lib/jenkins/plugins (overwriting any symlink that might be there already, and blocking the future overwrite of that plugin by something from the image)

 * /usr/lib/jenkins and /usr/lib64/jenkins are aliases so any existing symlinks in /var/lib/jenkins that point to /usr/lib*/jenkins will not be broken when using this new image.


In a future image if we update a plugin, nothing needs to happen (the symlink will point to the new version.  

If we add a new plugin to the image, it will be symlinked into /var/lib/jenkins during startup (because we run this logic on *every* startup, not just the first one).  (unless the user has already installed that plugin at some level in which case we leave it alone).


Does that sound right?

bparees:
The one problem I have is broken symlinks.  If we remove a plugin in the future, the symlink will still exist in /var/lib/jenkins, and it sounds like that will prevent jenkins from starting up?  Can we scan /var/lib/jenkins for broken symlinks and remove them?

gabemontero:
On Mon, Dec 4, 2017 at 4:14 PM Ben Parees <notifications@github.com> wrote:

> So let me attempt to summarize the behavior:
>
>    -
>
>    plugins baked into the image(in /usr/lib*/jenkins) are, at startup,
>    created as symlinks in /var/lib/jenkins/plugins unless the file already
>    exists in /var/lib/jenkins/plugins
>    -
>
>    content installed by the user(via the jenkins console or the various
>    image plugin install mechanisms) is copied into /var/lib/jenkins/plugins
>    (overwriting any symlink that might be there already, and blocking the
>    future overwrite of that plugin by something from the image)
>    -
>
>    /usr/lib/jenkins and /usr/lib64/jenkins are aliases so any existing
>    symlinks in /var/lib/jenkins that point to /usr/lib*/jenkins will not be
>    broken when using this new image.
>
> In a future image if we update a plugin, nothing needs to happen (the
> symlink will point to the new version.
>
> If we add a new plugin to the image, it will be symlinked into
> /var/lib/jenkins during startup (because we run this logic on *every*
> startup, not just the first one). (unless the user has already installed
> that plugin at some level in which case we leave it alone).
>
> Does that sound right?
>

Yes it does.

gabemontero:
On Mon, Dec 4, 2017 at 4:15 PM Ben Parees <notifications@github.com> wrote:

> The one problem I have is broken symlinks. If we remove a plugin in the
> future, the symlink will still exist in /var/lib/jenkins, and it sounds
> like that will prevent jenkins from starting up?
>
We could add an unlink in the run script when we remove from plugin text
file.

> Can we scan /var/lib/jenkins for broken symlinks and remove them?
>
will ask google when i get back to my desk

bparees:
> We could add an unlink in the run script when we remove from plugin text file.

yeah i'd rather just do it generically for broken symlinks.  

here's an easy way to test it, though cat'ing large binary files to /dev/null may prove non-performant, so you'll want to check that before doing it for N .jpi files.

```
$ ls -l a b broken
-rw-rw-r--. 1 bparees bparees 0 Dec  4 16:23 a
lrwxrwxrwx. 1 bparees bparees 1 Dec  4 16:24 b -> a
lrwxrwxrwx. 1 bparees bparees 1 Dec  4 16:23 broken -> c

$ cat b >& /dev/null ; echo $?
0

$ cat broken >& /dev/null ; echo $?
1
```

gabemontero:
On Mon, Dec 4, 2017 at 4:25 PM, Ben Parees <notifications@github.com> wrote:

> We could add an unlink in the run script when we remove from plugin text
> file.
>
> yeah i'd rather just do it generically for broken symlinks.
>
> here's an easy way to test it, though cat'ing large binary files to
> /dev/null may prove non-performant, so you'll want to check that before
> doing it for N .jpi files.
>
> $ ls -l a b broken
> -rw-rw-r--. 1 bparees bparees 0 Dec  4 16:23 a
> lrwxrwxrwx. 1 bparees bparees 1 Dec  4 16:24 b -> a
> lrwxrwxrwx. 1 bparees bparees 1 Dec  4 16:23 broken -> c
>
> $ cat b >& /dev/null ; echo $?
> 0
>
> $ cat broken >& /dev/null ; echo $?
> 1
>
>
`stat -L b` will do it wrt existence ... `test -h` will let us know if it
is a symlink ... I'll cook up combination of those two

gabemontero:
unlink if link points to non-existent pushed

gabemontero:
manual migration test passed

bparees:
lgtm, would like one more second review from @jim-minter though since we've established this is not as trivial as we thought :)

2/contrib/s2i/run
```diff
@@ -170,6 +175,51 @@ new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
...
+# have since deleted from the image
+for FILENAME in ${JENKINS_HOME}/plugins/*.jpi; do
+    # test if it is a sym link, otherwise users have overriden, so leave alone
+    # test command silent, don't need dev/null
+    test -h $FILENAME
+    if [ $? -eq 0 ]; then
+	stat -L $FILENAME >& /dev/null
+	if [ $? -eq 0 ]; then
+	    continue
+	fi
```
jim-minter: (fwiw I think all the lines above condense down to `if [ ! -e "$FILENAME" ]; then`.  Am not requiring the change be made.)

2/contrib/s2i/run
```diff
@@ -170,6 +175,51 @@ new_password_hash=`obfuscate_password ${JENKINS_PASSWORD:-password} $old_salt`
...
+    fi
+done
+
+# for rhel, some maintenance wrt copying plugins is done on every startup;  this logic
+# covers the following scenarios:
+# 1) this is the initial start of the image and the JENKINS_HOME/configured file is not
+# yet present
+# 2) this is a release upgrade, and new plugins need to be added
+#
+# we employ links instead of copies, and check for the existance in JENKINS_HOME/plugins,
```
jim-minter: (existence)

jim-minter:
lgtm too - happy for this to merge as is.

gabemontero:
Thanks @jim-minter @bparees

gabemontero:
@bparees and I discussed updating the README to explain the particulars of all this ... going to do that in a separate PR.

gabemontero:
[merge]

PR #429, opened 2017-12-01
Upgrade Node.js slave to Node.js 8

grdryn:
This is mainly just an observation that the current Node.js 4 that's
referenced in these Dockerfiles is quite old at this point. While it's
still maintained, it's in the 'Maintenance' phase of its LTS cycle,
and will be EOL in April 2018:

https://github.com/nodejs/Release#lts-schedule

I was originally going to propose Node.js 6 here, but given that
Node.js 8 is the most recent active LTS, this proposes to go straight
to that.

I've successfully built the CentOS image from this commit locally, but
I haven't done any extensive testing on it -- I wanted to first see if
this is a change that you actually want.

I tried building from the Dockerfile.rhel7 image inside a CDK
minishift VM, but it couldn't find the rh-nodejs8* packages, even
though there's a RHEL7 Dockerfile that appears to use them at [1], and
I see a nodejs-8-rhel7 image on the Red Hat Container Catalog[2].

[1] https://github.com/sclorg/s2i-nodejs-container/blob/master/8/Dockerfile.rhel7
[2] https://access.redhat.com/containers/#/registry.access.redhat.com/rhscl/nodejs-8-rhel7

### DIFF
--- a/slave-nodejs/Dockerfile
+++ b/slave-nodejs/Dockerfile
@@ -2,7 +2,7 @@ FROM openshift/jenkins-slave-base-centos7
 
 MAINTAINER Ben Parees <bparees@redhat.com>
 
-ENV NODEJS_VERSION=4.4 \
+ENV NODEJS_VERSION=8 \
     NPM_CONFIG_PREFIX=$HOME/.npm-global \
     PATH=$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:$PATH \
     BASH_ENV=/usr/local/bin/scl_enable \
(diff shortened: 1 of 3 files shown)


bparees:
we need to be careful making changes here.  anyone using this slave image will be pushed to nodejs8 which could break their workflow.  I think this needs to be introduced as a new slave image instead.

grdryn:
@bparees Yes, I meant to suggest that as another option, e.g. a new directory here called `slave-nodejs8` (and maybe even a `slave-nodejs6` one?). Would you like me to change this PR to reflect that?

bparees:
Let's hold off on that for now, we are in the middle of some other issues and depending on the outcome of that, it may drive our slave image strategy going forward.

tjololo:
Any updates on this? We need a jenkins nodejs slave at a newer version than v4 (preferably v8).
For now we have just forked the repository and built our own, but would have been nice to have a updated image provided.

gabemontero:
There is a rework of our images coming that among other things will lead to the update of nodejs for the slave image.  It will also
- lead to the deprecation of the existing images 
- support multi arch scenarios
- possibly change who owns these images ... i.e. they *may* not come out of this repo anymore, but that is not decided yet

The work is currently being tracked by https://trello.com/c/YQm7osws/1493-5-jenkins-slave-multi-arch-enablement-jenkinsintegration ... you can join trello board and track its progress if you like.  It is proposed for the openshift 3.10 time frame, but not committed yet ... most likely that means 2nd half of 2018.

Most likely we will not be updating the current slave images in the interim as a result of all this.

PR #431, opened 2017-12-01
use a tag for slave image config and drop .net slave image config

bparees:
in conjunction with https://github.com/openshift/origin/pull/17562 to use a version specific slave image tag in the slave pod template definition, instead of latest, and https://github.com/openshift/aos-cd-jobs/pull/897 to start pushing version-specific tags for the jenkins master+slave images.

@jim-minter @smarterclayton ptal.

@tmds fyi we're removing the pre-configured .net slave image configuration after realizing that backwards compatibility is virtually impossible to guarantee since the .net image includes the oc client binary, which must be aligned with the cluster version the slave image is used on.  So we don't want to deliver a jenkins master image that is preconfigured to use the .NET image when we can't guarantee that the .NET image it's preconfigured to use, is compatible with that cluster level.

So we prefer to force users to select an appropriate .net image tag on their own.  (Plus in general we're realizing including pre-configured slaves in the master, while convenient for users, was probably a mistake from a maintenance perspective)

For our own slave images, we are going to start publishing them with origin/ocp version tags (e.g. v3.9) so we can enforce the following chain:

A v3.9 cluster will install a v3.9 jenkins master and instruct it (via the JENKINS_SLAVE_IMAGE_TAG env var being added here https://github.com/openshift/origin/pull/17562) to use the v3.9 slave image tags.


bparees:
[test]

bparees:
@gabemontero @jim-minter needs some manual testing but ptal.

@smarterclayton please re-confirm you're ok w/ us migrating existing user slave config in this way.  Note that I had no choice but to hardcode the default version value in order to ensure keep moving it up each time a new image gets deployed on a cluster.  In fact now I don't even think we should have JENKINS_SLAVE_IMAGE_TAG as an env var, because it's just going to get set to the version the user deployed against, and then never move up when their image version/cluster version moves up.

2/contrib/s2i/run
```diff
@@ -6,6 +6,11 @@
 #
 # It also sets the admin password to ${JENKINS_PASSWORD}.
 #
+
+# This value needs to be updated to reflect the current release version
+# to ensure we properly migrate old slave configurations to the new tag.
+JENKINS_SLAVE_IMAGE_TAG=${JENKINS_SLAVE_IMAGE_TAG:-v3.9}
```
jim-minter: can this be wired through directly from the release.version file?  E.g. ADD it via the Dockerfiles, then do `JENKINS_SLAVE_IMAGE_TAG=${JENKINS_SLAVE_IMAGE_TAG:-$(cat path/to/release.version)}`?

2/contrib/s2i/run
```diff
@@ -86,6 +91,17 @@ function install_plugins() {
...
+  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-maven-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-maven-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-nodejs-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-nodejs-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+
+  # replace registry.access.redhat.com/openshift3/jekins-slave-xxxx-rhel7[:optionaltag] with registry.access.redhat.com/openshift3/jenkins-slave-xxxx-rhel7:VersionTag
+  sed -i "s#<image>registry.access.redhat.com/openshift3/openshift3/jenkins-slave-maven-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  sed -i "s#<image>registry.access.redhat.com/openshift3/openshift3/jenkins-slave-nodejs-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
```
jim-minter: `openshift3/openshift3` here and on the line above looks like a mistake

2/contrib/s2i/run
bparees: I was looking at doing this but the short answer is "not easily" and i don't want to make things too confusing for the dist-git repo that's going to have to build this for rhel.

2/contrib/s2i/run
bparees: ooof. thank you.

bparees: fixed

jim-minter: I'm wondering whether the hard-coding of "registry.access.redhat.com/" is going to inhibit QE testing; whether it would make sense to make it a parameter.

2/contrib/s2i/run
jim-minter: If it's not wired through, would it be advisable to have some test to ensure that the two are in sync?  Is that any nicer?

2/contrib/s2i/run
```diff
@@ -86,6 +91,17 @@ function install_plugins() {
   fi
 }
 
+function migrate_slave_configuration() {
+  # replace [docker.io/]openshift/jekins-slave-xxxx-centos7[:optionaltag] with docker.io/openshift/jenkins-slave-xxxx-centos7:VersionTag
+  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-maven-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-maven-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-nodejs-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-nodejs-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+
+  # replace registry.access.redhat.com/openshift3/jekins-slave-xxxx-rhel7[:optionaltag] with registry.access.redhat.com/openshift3/jenkins-slave-xxxx-rhel7:VersionTag
+  sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
```
jim-minter: I'm wondering whether the hard-coding of "registry.access.redhat.com/" is going to inhibit QE testing; whether it would make sense to make it a parameter.

bparees: it shouldn't.  they can provide/set the config.xml to any value they want and then allow migration to run and check that the config.xml is properly updated.

i'd really like to minimize the moving pieces here if we can.

2/contrib/s2i/run
bparees: for now i've worked around the symlink problem by making the "real" file live in the "2" directory and the root one be the symlink.

If we ever create a "3" directory we'll have to revisit how this is being done.

bparees: (you can't add a symlink from within a dockerfile if the symlink points outside the docker build context dir)

jim-minter:
lgtm

bparees:
[merge]

PR #438, opened 2017-12-07
Make plugin download timeouts configurable

goober:
The default timeout of 5 seconds for downloading plugins are to narrow for us.
This PR increases the curl connection timeout from 5 to 20 seconds. This follows the original [install-plugins-sh](https://github.com/jenkinsci/docker/commit/21b0f37a8b2d4dfa359cfa87eb849531ff5a9290) from the [official jenkins docker image](https://github.com/jenkinsci/docker)

At the same time make it possible to configure the timeouts for plugin downloads, which also reflects the script from the original jenkins docker image. See [install-plugins.sh#L78](https://github.com/jenkinsci/docker/blob/master/install-plugins.sh#L78)


bparees:
thanks @goober 
[merge]

gabemontero:
+1 thanks @goober

PR #440, opened 2017-12-11
perform update center syncrhonization during startup instead of first login

gabemontero:

I've observed within the last few weeks and inordinate delay the very first time you login to a jenkins instance instantiated from our image, often resulting in a 504 / gateway timeout

In conjunction, I observer a lengthy CPU burn with the jenkins process.  And it occurred across different permutations (i.e. both with oauth on and off, no openshift pod, etc.)

Thread dumps during the CPU burn revealed that the administrative alerts were being accessed on the login threads, and the jenkins core update / security alert monitors were taking forever, downloading/parsing their json data.  

UPDATE March 5 2018:
Instead of the monitor disablement below, we were able to move the processing to jenkins startup via an init groovy scrip
END UPDATE

I suspect a recent jenkins LTS upgrade coupled with the pipeline plugin list growing in complexity has lead to this.  In typical, centralized, long running jenkins deployments, I suspect this isn't a huge deal.

The only viable workaround I could conjure was to disable those two monitors on startup.

Of course the user can re-enable as they desire.

Per discussion with @bparees, the core update monitor is more or less orthogonal to our scenario since any jenkins core update would be best proscribed via an image regen.

For the security alert (basically those warnings when they publish an plugin update for a discovered security hole), there are other means of getting pushed notified via mailing lists.  But perhaps this is more arguable.

### DIFF
--- a/2/contrib/openshift/configuration/init.groovy.d/init.groovy
+++ b/2/contrib/openshift/configuration/init.groovy.d/init.groovy
@@ -1,0 +1,3 @@
+import jenkins.model.Jenkins
+Jenkins.getInstance().getPluginManager().doCheckUpdatesServer()
+Jenkins.getInstance().getUpdateCenter().getCoreSource().getData()
--- a/pkg/jenkins/jenkins.go
+++ b/pkg/jenkins/jenkins.go
@@ -80,7 +80,7 @@ func (j *Jenkins) Start(image string, env []string) error {
 }
 
 func (j *Jenkins) wait() error {
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 8*time.Minute)
 	defer cancel()
 
 	for {


bparees:
Can you open this w/ upstream jenkins
1) in hopes that they can fix/alleviate it?
2) in case we're doing something wrong?

bparees:
I mean open an issue w/ upstream.

2/contrib/s2i/run
```diff
@@ -56,8 +58,10 @@ function create_jenkins_config_from_templates() {
             elif [[ "${target_path}" == "${image_config_dir}/credentials.xml" ]]; then
                 create_jenkins_credentials_xml
             else
-                # Allow usage of environment variables in templated files, e.g. ${DOLLAR}MY_VAR is replaced by $MY_VAR
-                DOLLAR='$' envsubst < "${template_path}" > "${target_path}"
+                # 1) Allow usage of environment variables in templated files, e.g. ${DOLLAR}MY_VAR is replaced by $MY_VAR;
+		# 2) have to fake out envsubst to wrt internal java class reference hudson.model.UpdateCenter$CoreUpdateMonitor
+		# where we are disabling jenkins core update monitor because of CPU churn on initial login
+                CoreUpdateMonitor="\$CoreUpdateMonitor" DOLLAR='$' envsubst < "${template_path}" > "${target_path}"
```
bparees: this block isn't applicable to config.xml.tpl anyway.

2/contrib/openshift/configuration/config.xml.tpl
```diff
@@ -1,6 +1,11 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <hudson>
-  <disabledAdministrativeMonitors/>
+  <!-- disable core update monitor because of cpu churn during initial login;-->
+  <!-- update of jenkins core for our image should be controlled via image extension -->
+  <disabledAdministrativeMonitors>
+    <string>hudson.model.UpdateCenter$CoreUpdateMonitor</string>
```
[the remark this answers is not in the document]
gabemontero: I thought I tried this and it failed, but maybe I had a typo or something.  I'll give it another go.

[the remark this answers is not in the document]
bparees: you'll have to edit s2i/run since s2i/run isn't setting the DOLLAR variable before running envsubst, for the config.xml.tpl processing.

2/contrib/s2i/run
gabemontero: Yeah I saw that in testing, but was uncertain exactly what path this did in fact cover.  Is it for injecting templates *OTHER* than the config.xml and credentials.xml?

Anyway, if there was anyway this could still impact config.xml, it would be covered, and shouldn't impact the other cases.

But sure, based on the clarification, I can remove if it makes sense.

gabemontero:
Yeah, on the upstream jenkins issue, I've been thinking about this, but haven't settled on how to do this in a way that will be productive.

They'll want a repro specification of some form they can use.

Minimally, that would entail our current set of plugins and jenkins version.  Telling them to use our image might cause them to disqualify out of hand without objectively giving it a go.  

Bottom line, I will need to spend some more time to make sure if that was sufficient, or if not, what else would be needed.

2/contrib/s2i/run
bparees: >  Is it for injecting templates OTHER than the config.xml and credentials.xml?

yes.  and i'm not sure there's really a great reason there being special switches for each file, but yes.

> But sure, based on the clarification, I can remove if it makes sense.

overall i'd rather move to a common ${DOLLAR} based approach for all *tpl substitution. assuming you can sort out why it wasn't working for you earlier.

gabemontero:
the latest jenkins image 2.89.1 has a change long regarding the url the update center uses ... it is conceivable (though I've found no precise evidence of this yet) that could have some bearing on the delay I've seen

will report back once @bparees PR for that has merged and I've had a chance to try with it

gabemontero:
Through some continued observation, I have some clarifications wrt:

1) the delay appear only with 
    a) the initial start of jenkins
    b) when a plugin has been changed (of something I can do repeatedly when debugging or fixing something wrt a plugin)
2) the implication of course being that once "plugin state" has been determined and constant, we don't go down the plugin analysis path that leads to the CPU burn
3) I'm pretty sure I've seen some variability in the delay base on the responsiveness of the update center; this seems plausible in that communicating with the update center when dealing with plugin versions is a reasonable assumption ... a side project of mine is to learn some of the linux network capturing tools sufficiently so that I can capture proof of this

All said, I'm more inclined to say what we are doing here is "not wrong", and this delay does not stem specifically from how we've set things up, but I'm not ready to claim "due diligence investigation" is complete.

gabemontero:
Per @jim-minter 's analysis re: java GC, I ran my jenkins pods with a mem limit of 2Gi and did *NOT* encounter the problem today, even through my multiple restarts and plugin reinstalls as part of diagnosing / fixing some problems.

CPU was much less .. all that, though I need to learn from @jim-minter the best way to monitor per thread cpu on linux, I think I was able to look at it this time (I previously was only using `top` and only seeing the entire jenkins jvm usage), and it seemed better.

Next step - see about the gc arg changes @jim-minter proposed when running at 512

bparees:
@gabemontero can you reproduce this just running jenkins directly on your machine (not in a container) with the same heap/gc arguments?

If so, that might be a good way to produce a reproducer to cloudbees.

My current thinking on this is:

0) see if you see the same behavior locally (not in a container)
1) see if you see the same behavior in online (you can deploy your own jenkins image in online to test it)
2) if so, i'm inclined to proceed with merging this PR as a temporary workaround while pursuing the problem upstream w/ cloudbees.
3) i'd like some clarification on exactly what update warnings get disabled with this (just plugins? the jenkins version itself?  other things?)  Is it warnings about any plugin with a newer version or just plugins with security issues?  The answer to that may influence my thinking on (2).

gabemontero:
On Fri, Mar 2, 2018 at 1:19 PM, Ben Parees <notifications@github.com> wrote:

> @gabemontero <https://github.com/gabemontero> can you reproduce this just
> running jenkins directly on your machine (not in a container) with the same
> heap/gc arguments?
>
> If so, that might be a good way to produce a reproducer to cloudbees.
>
I think at best all we'll prove there is a) not a problem with running our
image in a pod, and b) that jenkins can no longer run in a 512 Mi JVM with
X number of plugins,
and they are going are going to say "yes, we recommend a minimum heap of
Y", and not consider it a bug.

I think there is some value in a), so will pursue running jenkins directly
on your machine (not in a container) with the same heap/gc arguments today.

> My current thinking on this is:
>
>    1. see if you see the same behavior locally (not in a container)
>    2. see if you see the same behavior in online (you can deploy your own
>    jenkins image in online to test it)
>    3. if so, i'm inclined to proceed with merging this PR as a temporary
>    workaround while pursuing the problem upstream w/ cloudbees.
>    4. i'd like some clarification on exactly what update warnings get
>    disabled with this (just plugins? the jenkins version itself? other
>    things?) Is it warnings about any plugin with a newer version or just
>    plugins with security issues? The answer to that may influence my thinking
>    on (2).
>
> Not sure I'll be able to answer all those different nuances without
detailed investigation of their code.  But I'll see what I can uncover
through internet searching
and perusal of their docs.

Also, unless you have already done it, I'm going to revisit running with
1Gi for the memory limit.

Pending what I discover in the above, I'm on the fence between disabling
the update stuff vs. bumping the heap (at least to 1Gi if that is
sufficient), or some combination.

bparees:
> Also, unless you have already done it, I'm going to revisit running with 1Gi for the memory limit.

i have not.

>  I'm on the fence between disabling the update stuff vs. bumping the heap (at least to 1Gi if that is sufficient), or some combination.

starter online only gives users 1gig of quota, total.  So going to 1gig heap for jenkins isn't really viable w/o revisiting those limits.  @sspeiche would need to be involved.  But certainly if we're going to do anything quickly, disabling the checks seems like the only option.

gabemontero:
I'm having trouble duplicating our image in a pod env via running jenkins outside of a container.

What I attempted was : 
a) launch our template per usual, but do not log in
b) rsh in, tar up /var/lib/jenkins and copy to my laptop fs
c) also cp the jenkins.war from the pod
d) fetch the precise java invocation via ps, copy/paste that

However, jenkins forces me down the initial setup path, does a bunch of plugin stuff as part of that, and ends up blowing away our plugin list for their proscribed list.

So it isn't a valid comparison.

I think the closest I'll be able to duplicate is do a docker run of our image, where I use our existing env vars to limit the heap accordingly (with a 512Mi pod, we get a 256Mi heap).

thoughts @bparees ?

**Aside, 256Mi is comparatively pretty damn small for a java app with the breath of Jenkins**

Went back and reviewed what disabling means.  In the current form, it prevents both the jenkins core check and the plugins checks with their update centers (that is the interaction with the update center I mentioned previously).  Hence all those security advisories for either jenkins core or any of the plugins will not occur.  Warning for us not turning on say CSRP or some such do still occur.

Perhaps it is that flavor of answer you were worried about.

gabemontero:
On the security advisory alerting, cloudbees does have web pages with the details, emails alerts, etc. so customer could find out about them from other means.  It is just a convenience thing.

gabemontero:
OK, I tried out the following docker run to mimic what we do with our image.  Using 

`docker run -e "JAVA_MAX_HEAP_PARAM=-Xmx256m" -m 512M --memory-swap 512M -p 8080:8080 -it docker.io/openshift/jenkins-2-centos7:latest`

Where we establish the same
a) default max heap for the jvm of 256M
b) the max host mem or 512M, and the lack of swap

I see quite the same GC and CPU activity as we see with booting up our image in openshift with the default template.  From full GC cycles, seconds spent spiking my laptops CPU, all the same.

However, the 504 GW timeout behavior is not the same.  Not seeing this.

So it dawned on me - is it the openshift route that is cutting things off.

I found this:  https://docs.openshift.com/container-platform/3.5/install_config/configuring_routing.html

And did this:  `oc annotate route jenkins --overwrite haproxy.router.openshift.io/timeout=5m`

I still see the bigtime CPU burn and GC for many seconds, where the login attempt sits for a bit, but after a good number of repro attempts, no 504 / GW timeout.

I think we have another solution option.  And some validation of previous findings.  And I at least am convinced now that running 2.89.4 jenkins with our set of plugins with 256M heap / 512M host with no swap outside of a container is necessary.

thoughts @bparees ?

gabemontero:
Also, I forgot to mention in the original description, I tried to look into initiating the PluginManager/UpdateCenter activity during the first log in attempt in one of our plugins as part of post startup, so that the processing was already in place when a user first tries to log in.

Unfortunately, the PluginManager/UpdateCenter methods related to this are protected with annotations like `@Restricted(DoNotUse.class) // WebOnly`and `@Restricted(NoExternalUse.class)`

bparees:
>I think the closest I'll be able to duplicate is do a docker run of our image, where I use our existing env vars to limit the heap accordingly (with a 512Mi pod, we get a 256Mi heap).
thoughts @bparees ?

No particular thoughts.  It would be nice to be able to hand cloudbees a "clean" reproducer, but perhaps giving them the docker invocation w/ our image is sufficient (I fear they may simply make claims our image's configuration/java args are the problem).  I think it's at least worth opening the issue...  if i had to guess, i'd speculate they are trying to do all the plugin checking in parallel and perhaps if they could be convinced to instead do it in batches they could avoid blowing up the heap.


> Aside, 256Mi is comparatively pretty damn small for a java app with the breath of Jenkins

I kind of hate this suggestion, but we could consider only disabling the checking for containers with 512meg limits.  Or even making the disablement configurable via env variable which we'd set by default in our template.  (The main problem being that this choice is made the first time you bring up jenkins, after that the choice is baked into the PV and we're not going to start rewriting the config on the PV)


> Went back and reviewed what disabling means. In the current form, it prevents both the jenkins core check and the plugins checks with their update centers (that is the interaction with the update center I mentioned previously). Hence all those security advisories for either jenkins core or any of the plugins will not occur. Warning for us not turning on say CSRP or some such do still occur.
Perhaps it is that flavor of answer you were worried about.

yeah i'm not sure what i was hoping for/fearing, but it definitely sounds like a lot of stuff to disable by default.

> So it dawned on me - is it the openshift route that is cutting things off.

yes I assumed this was the case, sorry we didn't discuss it :)

> I still see the bigtime CPU burn and GC for many seconds, where the login attempt sits for a bit, but after a good number of repro attempts, no 504 / GW timeout.
I think we have another solution option. And some validation of previous findings. And I at least am convinced now that running 2.89.4 jenkins with our set of plugins with 256M heap / 512M host with no swap outside of a container is necessary.

So with the longer GW timeout, I assume that the first time a user goes to login to jenkins they are going to see their browser spin for a long time(several minutes?) before the login proceeds?  That is probably better than a 504 but still likely to be a bad user experience.

> Also, I forgot to mention in the original description, I tried to look into initiating the PluginManager/UpdateCenter activity during the first log in attempt in one of our plugins as part of post startup, so that the processing was already in place when a user first tries to log in.

That's too bad.  I don't suppose there's a way to programmatically fake a login/trigger the update from somewhere higher in the stack?

gabemontero:
On Sun, Mar 4, 2018 at 8:27 PM, Ben Parees <notifications@github.com> wrote:

> I think the closest I'll be able to duplicate is do a docker run of our
> image, where I use our existing env vars to limit the heap accordingly
> (with a 512Mi pod, we get a 256Mi heap).
> thoughts @bparees <https://github.com/bparees> ?
>
> No particular thoughts. It would be nice to be able to hand cloudbees a
> "clean" reproducer, but perhaps giving them the docker invocation w/ our
> image is sufficient (I fear they may simply make claims our image's
> configuration/java args are the problem). I think it's at least worth
> opening the issue... if i had to guess, i'd speculate they are trying to do
> all the plugin checking in parallel and perhaps if they could be convinced
> to instead do it in batches they could avoid blowing up the heap.
>

They are not going to accept an issue, or agree this is a problem.  I do
not believe this will be productive.  Look at
https://support.cloudbees.com/hc/en-us/articles/204859670-Java-Heap-settings-best-practice

2G is the minimum they recommend there.  If we tell them we see high GC at
256Mi with 2.89.4, with the number of plugin we have, you can guess their
response.


> Aside, 256Mi is comparatively pretty damn small for a java app with the
> breath of Jenkins
>
> I kind of hate this suggestion, but we could consider only disabling the
> checking for containers with 512meg limits. Or even making the disablement
> configurable via env variable which we'd set by default in our template.
> (The main problem being that this choice is made the first time you bring
> up jenkins, after that the choice is baked into the PV and we're not going
> to start rewriting the config on the PV)
>
First blush not a big fan.

> Went back and reviewed what disabling means. In the current form, it
> prevents both the jenkins core check and the plugins checks with their
> update centers (that is the interaction with the update center I mentioned
> previously). Hence all those security advisories for either jenkins core or
> any of the plugins will not occur. Warning for us not turning on say CSRP
> or some such do still occur.
> Perhaps it is that flavor of answer you were worried about.
>
> yeah i'm not sure what i was hoping for/fearing, but it definitely sounds
> like a lot of stuff to disable by default.
>
> So it dawned on me - is it the openshift route that is cutting things off.
>
> yes I assumed this was the case, sorry we didn't discuss it :)
>
> I still see the bigtime CPU burn and GC for many seconds, where the login
> attempt sits for a bit, but after a good number of repro attempts, no 504 /
> GW timeout.
> I think we have another solution option. And some validation of previous
> findings. And I at least am convinced now that running 2.89.4 jenkins with
> our set of plugins with 256M heap / 512M host with no swap outside of a
> container is necessary.
>
> So with the longer GW timeout, I assume that the first time a user goes to
> login to jenkins they are going to see their browser spin for a long
> time(several minutes?)
>
2 minutes has seemed sufficient with my testing here at home. I'm curious
what it looks like from the office on Monday.

> before the login proceeds? That is probably better than a 504 but still
> likely to be a bad user experience.
>
The least of the evils IMO.  And IMO this is what we should do to meets min
for 3.9

As you've noted, we need to talk to Steve about the implications for
online.  But for any other scenario then, I think we need to start
documenting the need to bump off the current default in the template.

> Also, I forgot to mention in the original description, I tried to look
> into initiating the PluginManager/UpdateCenter activity during the first
> log in attempt in one of our plugins as part of post startup, so that the
> processing was already in place when a user first tries to log in.
>
> That's too bad. I don't suppose there's a way to programmatically fake a
> login/trigger the update from somewhere higher in the stack?
>
I have not found one as of yet.  After a couple of separate attempts, I'm
not optimistic, certainly not by EOB Monday.

bparees:
> 2G is the minimum they recommend there.  If we tell them we see high GC at
256Mi with 2.89.4, with the number of plugin we have, you can guess their
response.

fair enough

> The least of the evils IMO.  And IMO this is what we should do to meets min
for 3.9

yeah, certainly for 3.9 it seems worth doing.

torstenwalter:
> Also, I forgot to mention in the original description, I tried to look into initiating the PluginManager/UpdateCenter activity during the first log in attempt in one of our plugins as part of post startup, so that the processing was already in place when a user first tries to log in.
> 
> Unfortunately, the PluginManager/UpdateCenter methods related to this are protected with annotations like @Restricted(DoNotUse.class) // WebOnlyand @Restricted(NoExternalUse.class)

@gabemontero You can use a init.groovy.d script in order to check for Jenkins updated during startup.
See https://wiki.jenkins.io/display/JENKINS/Post-initialization+script

Below is an example:

```
import jenkins.model.Jenkins
Jenkins.getInstance().getPluginManager().doCheckUpdatesServer()
```

gabemontero:
thanks so much for the suggestion @torstenwalter !!

Just tried it out, and while I see the script getting executed:

```
Mar 05, 2018 8:43:56 PM jenkins.util.groovy.GroovyHookScript execute
INFO: Executing /var/lib/jenkins/init.groovy.d/init.groovy
```

It is not sufficient for this particular issue.  Either more than the `doCheckUpdatesServer()` call is needed, or the annotations I mentioned are resulting in the call getting no-op'ed.  The plugin checks are still occurring and cause massive GC during the first login if our heap is sufficiently low.

I'll see about adding other plugin manager related start up calls and if it makes any difference, but at first blush this seemingly useful idea is not proving fruitful.

gabemontero:
OK, based on revisiting the thread dump analysis when the GC is occurring, I added one more call to the init.groovy.  Based on that:

a) the download/processing of all that update center stuff happens on startup; add about 1 to 1.5 minutes to the startup
b) but login once the pod is up is pretty instantaneous

I'll push the update again in a moment.  Thanks again @torstenwalter for the general idea.

@bparees - the delay on startup is the lesser of the evils I think ... agree?

bparees:
> @bparees - the delay on startup is the lesser of the evils I think ... agree?

yeah, agreed.

gabemontero:
Cool thanks

gabemontero:
[merge]

bparees:
@gabemontero you'll want to update the PR description

gabemontero:
yep PR title and initial description updated 

and yeah the tests have started timing out again, even with the last PR (which had some failures but the last test passed to validate the merge).

Talking a pre-bedtime look ...

gabemontero:
... the new additional delay during startup might be drifting to a bad place ... the timeout for container start appears to be 2 minutes

gabemontero:
[merge]

Issue #423, opened 2017-11-29
Confusing startup messaging


jim-minter:
```
Using 64 bit Java since OPENSHIFT_JENKINS_JVM_ARCH is not set (historic setting)
max heap in MB is 256 and 64 bit was not explicitly set so using 32 bit Java
```

should be clearer.

gabemontero:
I believe @jim-minter sorted this out to his satisfaction with the above PR

of course reopen @jim-minter if I misinterpreted things

PR #442, opened 2017-12-13
Jenkins memory cleanup

jim-minter:


jim-minter:
@bparees this is a starting point - want to get things about right in run-jnlp-client, then maybe squash, then carry the same changes through the Jenkins startup.  Not tested yet either.  ptal for early feedback and discussion?

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
     fi
   fi
 
-  CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
-  DEFAULT_MEMORY_CEILING=$((2**40-1))
-  if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
+  if [[ -z "${JAVA_TOOL_OPTIONS}" ]]; then
+    # these options will automatically be picked up by any child JVM process
```
bparees: any child jvm process or just "any jvm process"?

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
...
-    #if machine has 4GB or less, meaning max heap of 2GB given current default, force use of 32bit to save space unless user
-    #specifically want to force 64bit
-    HEAP_LIMIT_FOR_32BIT=$((2**32-1))
-    HEAP_LIMIT_FOR_32BIT_IN_MB=$((${HEAP_LIMIT_FOR_32BIT}/1024**2))
     CONTAINER_HEAP_MAX=$(echo "${CONTAINER_MEMORY_IN_MB} ${CONTAINER_HEAP_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
-    if [[ -z $OPENSHIFT_JENKINS_JVM_ARCH && "${CONTAINER_HEAP_MAX}" -lt "${HEAP_LIMIT_FOR_32BIT_IN_MB}"  ]]; then
-      echo "max heap in MB is ${CONTAINER_HEAP_MAX} and 64 bit was not explicitly set so using 32 bit Java"
-      alternatives --set java $JVMPath32bit
-      export MALLOC_ARENA_MAX=1
+    if [[ $CONTAINER_HEAP_LIMIT_IN_MB && $CONTAINER_HEAP_MAX -gt $CONTAINER_HEAP_LIMIT_IN_MB ]]; then
```
bparees: can we call this(CONTAINER_HEAP_LIMIT_IN_MB)  "JNLP_MAX_HEAP_UPPER_BOUND_MB"?  yeah it's wordy but nothing else offers the same clarity that:

1) this is for the jnlp jvm (presumably we'll want a different, or no, value for the other JVMs)
2) it's the upper bound for the max heap setting (vs the upper bound for the min heap setting, or just being the max heap limit setting for the jvm)

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
...
+      if [[ -z "$JAVA_INITIAL_HEAP_PARAM" ]]; then
+        JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
+      fi
+    fi
+
+    # if the calculated max heap is < 2GiB and OPENSHIFT_JENKINS_JVM_ARCH is not
+    # set, use 32 bit JVM for space efficiency
+    if [[ -z $OPENSHIFT_JENKINS_JVM_ARCH ]]; then
+      if [[ "${CONTAINER_HEAP_MAX}" -gt 2048 ]]; then
+        echo "max heap in MB is ${CONTAINER_HEAP_MAX}, using 64 bit Java"
```
bparees: While you're cleaning things up, both of these messages(here and the 32 bit one below) should say "OPENSHIFT_JENKINS_JVM_ARCH was not set and max heap in MB is X.  Using Y bit Java"

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
...
-  fi
 
-  if [ -z $JAVA_GC_OPTS ]; then
-      # We no longer set MaxMetaspaceSize because the JVM should expand metaspace until it reaches the container limit.
-      # See http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/4dd24f4ca140/src/share/vm/memory/metaspace.cpp#l1470
-      JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
+  else
+    # no k8s/docker memory limit set
+    if [[ -z $OPENSHIFT_JENKINS_JVM_ARCH ]]; then
+      echo "no cgroup memory limit found, using 64 bit Java"
```
bparees: OPENSHIFT_JENKINS_JVM_ARCH is not set and no cgroup memory limit was found, using 64 bit Java"

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
...
-  if [ ! -z "${CONTAINER_CORE_LIMIT}" ]; then
-      JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMIT} -XX:CICompilerCount=2"
+  if [[ "${CONTAINER_CORE_LIMIT}" ]]; then
+    JAVA_CORE_LIMIT="-XX:ParallelGCThreads=${CONTAINER_CORE_LIMIT} -Djava.util.concurrent.ForkJoinPool.common.parallelism=${CONTAINER_CORE_LIMIT} -XX:CICompilerCount=2"
   fi
 
-  if [ -z "${JAVA_OPTS}" ]; then
-      JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dsun.zip.disableMemoryMapping=true"
+  if [[ -z "${CONTAINER_JAVA_OPTIONS}" ]]; then
+    CONTAINER_JAVA_OPTIONS="$JAVA_GC_OPTS $JAVA_INITIAL_HEAP_PARAM $JAVA_MAX_HEAP_PARAM $JAVA_CORE_LIMIT $JAVA_DIAGNOSTICS"
```
bparees: JNLP_JAVA_OPTIONS.  Same elsewhere as applicable.

gabemontero:
At first blush I like @bparees 's "I wonder if we should pick 32bit java purely based on the container's cgroup limit. (less than a 2gig container? you get 32bit. otherwise you get 64bit and if you want something else you tune it yourself)."

We'll see if that still holds after parsing @jim-minter 's changes in detail and @bparees 's feedback.

slave-base/contrib/bin/run-jnlp-client
jim-minter: will change to "any JVM process".  I mean(t) the spawned JNLP JVM and any subsequent JVMs it may spawn.

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -31,7 +31,7 @@ elif [ "${OPENSHIFT_JENKINS_JVM_ARCH}" == "x86_64"  ]; then
 else
     echo "OPENSHIFT_JENKINS_JVM_ARCH is set to ${OPENSHIFT_JENKINS_JVM_ARCH} so using 32 bit Java"
     alternatives --set java $JVMPath32bit
-    export MALLOC_ARENA_MAX=1
+    export MALLOC_ARENA_MAX=${MALLOC_ARENA_MAX:-1}
```
gabemontero: as an fyi, via commit 8f60691a47512bce66d328d0f8ca72612d3bc001 this stemmed from purely duplicating advice from the JBOSS side of the house.

slave-base/contrib/bin/run-jnlp-client
jim-minter: sure

slave-base/contrib/bin/run-jnlp-client
gabemontero: just for my education, JAVA_OPTS was a first class env var of sorts, specifically looked for by say maven and/or gradle or some such, and hence we are avoiding overlap, right?

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -108,8 +108,10 @@ if [[ $# -gt 1 ]]; then
     JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
   fi
 
-  if [ ! -z "${USE_JAVA_DIAGNOSTICS}" ]; then
-      JAVA_DIAGNOSTICS="-XX:NativeMemoryTracking=summary -XX:+PrintGC -XX:+PrintGCDateStamps -XX:+PrintGCTimeStamps -XX:+UnlockDiagnosticVMOptions"
+  # USE_JAVA_DIAGNOSTICS and JAVA_DIAGNOSTICS are legacy and may be removed in
+  # a future version of this script.
+  if [[ "${USE_JAVA_DIAGNOSTICS}" ]]; then
+    JAVA_DIAGNOSTICS="-XX:NativeMemoryTracking=summary -XX:+PrintGC -XX:+PrintGCDateStamps -XX:+PrintGCTimeStamps -XX:+UnlockDiagnosticVMOptions"
```
gabemontero: aside from the comment, how about we echo out that message if those variables were in fact set.

slave-base/contrib/bin/run-jnlp-client
jim-minter: Will do - am assuming to keep the already established CONTAINER_CORE_LIMIT / CONTAINER_INITIAL_PERCENT etc. as they are.

slave-base/contrib/bin/run-jnlp-client
jim-minter: I can definitely see that it could make a difference to export MALLOC_ARENA_MAX; I'm just trying to ensure that things are reasonably overridable.

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -21,14 +21,12 @@ source /usr/local/bin/generate_container_user
...
-    echo "Using 64 bit Java since OPENSHIFT_JENKINS_JVM_ARCH is not set"
-    alternatives --set java $JVMPath64bit
-elif [ "${OPENSHIFT_JENKINS_JVM_ARCH}" == "x86_64"  ]; then
+# If OPENSHIFT_JENKINS_JVM_ARCH is explicitly set, set the JVM architecture
+# used.  If not explicitly set, the architecture is decided below.
+if [[ "${OPENSHIFT_JENKINS_JVM_ARCH}" == "x86_64" ]]; then
     echo "64 bit Java explicitly set in OPENSHIFT_JENKINS_JVM_ARCH"
     alternatives --set java $JVMPath64bit
-else
+elif [[ "${OPENSHIFT_JENKINS_JVM_ARCH}" ]]; then
```
gabemontero: I haven't noticed the duplicate logging in the master logs (but maybe I've just always missed it?), but does it make sense if the same fix planned in either this PR a separate PR for the master run script?  Or was something different that prevented duplicate logging from occurring there?

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -96,11 +96,12 @@ if [[ $# -gt 1 ]]; then
     fi
 
     JAVA_MAX_HEAP_PARAM="-Xmx${CONTAINER_HEAP_MAX}m"
-    if [ -z $CONTAINER_INITIAL_PERCENT ]; then
-      CONTAINER_INITIAL_PERCENT=0.07
+    if [[ "$CONTAINER_INITIAL_PERCENT" ]]; then
+      CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+      if [[ -z "$JAVA_INITIAL_HEAP_PARAM" ]]; then
+        JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
+      fi
```
gabemontero: so if I caught the gist of the recent exchanges with Severin correctly (not tagging him on purpose to save him some additional chatter) the gist was that the JVM will do a better job at picking the default when used in conjunction with the honor cgroup flags, right?

if so, a comment explaining all that would be good.

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -73,6 +73,11 @@ if [[ $# -gt 1 ]]; then
   CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
   DEFAULT_MEMORY_CEILING=$((2**40-1))
   if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
+    # set this JVM's -Xmx and -Xms if not set already (not propagated to any
+    # child JVMs).  -Xmx can be calculated as a percentage, capped to a maximum,
+    # or specified straight.  -Xms can be calculated as a percentage or
+    # specified straight.  For the JNLP slave by default we specify -Xmx of 50%,
+    # uncapped; -Xms unspecified (JVM default is 1/64 of -Xmx).
```
gabemontero: ah I'm reviewing commit by commit  ... this comment is probably sufficient wrt my ask for a comment elsewhere, unless @jim-minter feels so inclined

jim-minter:
> We're now going to pick the java arch based on the jnlp JVM's heap. Which means someone who tunes for a small JNLP heap will get 32bit java and then their maven jvms will also be run in 32bit java.
>
> I wonder if we should pick 32bit java purely based on the container's cgroup limit. (less than a 2gig container? you get 32bit. otherwise you get 64bit and if you want something else you tune it yourself).

I think that's a good call (IMO an alternative in the JNLP slave would be entirely removing the 32/64-bit JVM selection).

I think that due to a bug the existing checked in behaviour is effectively: 32-bit if (overridden && !overridden to 64-bit) || (!overridden && cgroup limit set && cgroup limit < 8 GiB).  @bparees you want to change that to 2GiB (i.e. 1GiB heap) or 4GiB (i.e. expected 2GiB heap)?  This PR should currently be doing the latter, I'm actually in favour of the former.

jim-minter:
[the above also has the advantage of getting all the OPENSHIFT_JENKINS_JVM_ARCH code in one place, so that it can then go in a separate bash function]

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -65,57 +63,88 @@ if [[ $# -gt 1 ]]; then
...
+      CONTAINER_INITIAL_HEAP=$(echo "${CONTAINER_HEAP_MAX} ${CONTAINER_INITIAL_PERCENT}" | awk '{ printf "%d", $1 * $2 }')
+      if [[ -z "$JAVA_INITIAL_HEAP_PARAM" ]]; then
+        JAVA_INITIAL_HEAP_PARAM="-Xms${CONTAINER_INITIAL_HEAP}m"
+      fi
+    fi
+
+    # if the calculated max heap is < 2GiB and OPENSHIFT_JENKINS_JVM_ARCH is not
+    # set, use 32 bit JVM for space efficiency
+    if [[ -z $OPENSHIFT_JENKINS_JVM_ARCH ]]; then
+      if [[ "${CONTAINER_HEAP_MAX}" -gt 2048 ]]; then
```
gabemontero: I'm a bit curious as to how the old logic broke down, but it has been so long, I don't precisely remember the various test scenarios I did in fact run (though I certainly ran something).

Any break down here or off line would be cool.

slave-base/contrib/bin/run-jnlp-client
jim-minter: will do

slave-base/contrib/bin/run-jnlp-client
jim-minter: The JVM default is Xms=1/64 of Xmx, the 64 is configurable using a JVM flag.  Severin made a change to stop us setting Xms==Xmx and set it to 7% instead.  In fact I think he could have gone further, I think there's no harm in using the JVM default here.

slave-base/contrib/bin/run-jnlp-client
jim-minter: https://github.com/openshift/jenkins/issues/423 ;-)
I'll keep a single PR for now but close that issue off when it goes in.

slave-base/contrib/bin/run-jnlp-client
jim-minter:
```
    HEAP_LIMIT_FOR_32BIT=$((2**32-1))
    HEAP_LIMIT_FOR_32BIT_IN_MB=$((${HEAP_LIMIT_FOR_32BIT}/1024**2))
```

means that HEAP_LIMIT_FOR_32BIT_IN_MB comes out at 4095 - this was then tested against the heap size, not the container size.

bparees:
>  @bparees you want to change that to 2GiB (i.e. 1GiB heap) or 4GiB (i.e. expected 2GiB heap)? This PR should currently be doing the latter, I'm actually in favour of the former.

Why do you prefer the former?  Just because you don't want to default people into 32bit jvms except in very small containers?

slave-base/contrib/bin/run-jnlp-client
gabemontero: thanks for that ...now, maybe it wasn't clear ... I meant a comment in the code

jim-minter:
Yes.

bparees:
@jim-minter i guess that's fine with me.

jim-minter:
> just for my education, JAVA_OPTS was a first class env var of sorts, specifically looked for by say maven and/or gradle or some such, and hence we are avoiding overlap, right?

@gabemontero Yes. Gradle, among others, but not Maven.

jim-minter:
repushed, should include all changes.

bparees:
this looks like a fine start to me, no further comments on this iteration.

gabemontero:
same here - let's squash and merge !

I'll start a test run in the interim.

[test]

2/contrib/s2i/run
```diff
@@ -96,9 +96,9 @@ function install_plugins() {
 }
 
 function migrate_slave_configuration() {
-  # replace [docker.io/]openshift/jekins-slave-xxxx-centos7[:optionaltag] with docker.io/openshift/jenkins-slave-xxxx-centos7:VersionTag
-  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-maven-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-maven-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
-  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-nodejs-centos7\(:.*\)\{0,1\}</image>#<image>docker.io/openshift/jenkins-slave-nodejs-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  # replace [docker.io/]openshift/jekins-slave-xxxx-centos7[:optionaltag] with [docker.io/]openshift/jenkins-slave-xxxx-centos7:VersionTag
+  sed -i "s#<image>\(docker.io/\)\{0,1\}openshift/jenkins-slave-maven-centos7\(:.*\)\{0,1\}</image>#<image>\1openshift/jenkins-slave-maven-centos7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
```
jim-minter: @bparees ptal - needed this for testing - do you want it in a separate PR?

jim-minter:
[test]

jim-minter:
[test]

jim-minter:
I believe this is ready to be merged.

gabemontero:
[merge]

jim-minter:
Thanks @gabemontero.  Do I need to do anything to get new images built and pushed to docker hub?

gabemontero:
On Tue, Dec 19, 2017 at 3:54 PM Jim Minter <notifications@github.com> wrote:

> Thanks @gabemontero <https://github.com/gabemontero>. Do I need to do
> anything to get new images built and pushed to docker hub?
>
NP - and nope the push_jenkins_image on ci.openshift will pick up this
merge and push to docker hub automatically

There will also be a rhel image pushed to ci.openshit

To get things in the pipeline for registry.redhat we worked through the CI
team but Justin just created a new job for us.

We can wait until after the holidays for that though, right?

jim-minter:
@gabemontero You bet.  Thanks and happy holidays!

PR #449, opened 2017-12-19
add version tags to Dockerfiles

### LINKED ISSUE #446: Don't use alpha clients as oc client for CentOS images
Currently the parent image for CentOS `Dockerfile` used is
`
FROM openshift/origin
`
Which is built with latest alpha version of oc client. We can't rely on that, but instead client built into this image should follow origin releases. Current stable version is 3.7.

Version of `oc` client with latest Jenkins image:

```
$ oc version
oc v3.9.0-alpha.0+1c24d18-233
kubernetes v1.8.1+0d5291c
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://172.30.0.1:443
openshift v3.7.9
kubernetes v1.7.6+a08f5eeb62
```

tanskann:
Adds version tags to parent docker images so that at least `oc` will be stable instead of latest alpha version. 

Fixes #446

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin
+FROM openshift/origin:v3.7.0
 
 # Jenkins image for OpenShift
 #
--- a/slave-base/Dockerfile
+++ b/slave-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin
+FROM openshift/origin:v3.7.0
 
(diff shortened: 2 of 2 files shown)


bparees:
we aren't going to lock this down:
1) it's a burden to update the values
2) it prevents us from testing the latest oc clients within the jenkins images we are building.

bparees:
if you make this pull against the openshift-3.7 branch instead i'll merge it there.  that will get you a 3.7 tagged jenkins image w/ the origin 3.7 client.

PR #450, opened 2017-12-19
Update credentials-binding plugin

jcpowermac:
When using `withCredentials` with v1.12 of the credentials-binding
plugin causes the console output to become corrupted with extra
asterisks making output difficult to read.  This is also described in [JENKINS-41760](https://issues.jenkins-ci.org/browse/JENKINS-41760)

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -23,6 +23,8 @@ mercurial:1.54
 matrix-project:1.7.1
 multiple-scms:0.6
 ssh-credentials:1.13
+credentials-binding:1.13
+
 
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

on the surface I'm good with this, but I want to see the workflow/pipeline plugin flow that and if it was bringing in the older version.  Ultimately, as it progresses and picks up later versions of credential-bindings, we'll want to remove this bumping of it and pick it up from there.

I'll examine our test output and report back.  Based on the results, I may have @jcpowermac add a comment so we remember what to track / look for.

thanks

gabemontero:
OK, pipeline-model-definition (declaritive pipeline) pulls it in directly, as well as indirectly via its dependency on docker-workflow/docker-common.  But those only bump it to 1.12.

I'll go ahead a merge this but create a follow up PR to add the appropriate comments so we don't forget
to remove it long term.

gabemontero:
[merge]

PR #452, opened 2017-12-20
Create LICENSE

Donis-:
Hi, we found this project useful and would like to use it our organisation, but unfortunately it's missing licensing terms. With this request i'm proposing to introduce Apache Licence 2.0 terms for openshift/jenkins.


bparees:
I guess if it's good enough for origin (https://github.com/openshift/origin/blob/master/LICENSE)

Issue #451, opened 2017-12-20
s2i build of jenkins does not know the installed plugins since they are only linked in the run script


bakito:
I'm customizing the openshift jenkins by using the s2i build as described here: https://docs.openshift.com/container-platform/3.6/using_images/other_images/jenkins.html#jenkins-as-s2i-builder 

Since this commit (
https://github.com/openshift/jenkins/commit/f7b3afb30f027d29c6bbf46f56a345e7f87e0a8c#diff-e6dcb991982488881a38bc0d22e481a0 line 208ff) the installation of plugins during s2i build does not work correctly, because at assembly time the pre-installed plugins are not yet visible for the install-plugin.sh script.

Depending on what plugins get installed, some dependent plugins with a much lower version get installed, because the do not exist for the script.

When the container is started, the links from the base installation are created, which leads to an inconsistent and unusable jeknins instance.

To assure the s2i is working correctly, the links should also/already be made from the assembly script.

gabemontero:
On Wed, Dec 20, 2017 at 7:15 AM, Marc Brugger <notifications@github.com>
wrote:

> I'm customizing the openshift jenkins by using the s2i build as described
> here: https://docs.openshift.com/container-platform/3.6/using_
> images/other_images/jenkins.html#jenkins-as-s2i-builder
>
> Since this commit (
> f7b3afb#diff-e6dcb991982488881a38bc0d22e481a0
> <https://github.com/openshift/jenkins/commit/f7b3afb30f027d29c6bbf46f56a345e7f87e0a8c#diff-e6dcb991982488881a38bc0d22e481a0>)
> the installation of plugins during s2i build does not work correctly,
> because at assembly time the pre-installed plugins are not yet visible for
> the install-plugin.sh script.
>
> Depending on what plugins get installed, some dependent plugins with a
> much lower version get installed, because the do not exist for the script.
>
> When the container is started, the links from the base installation are
> created, which leads to an inconsistent and unusable jeknins instance.
>
Must be the move of the linking out of the Dockerfile and into the start up
path.  Our s2i testing did not catch the nuance above @bparees.

First blush, refactor the new (re)linking bash logic into a separate .sh
file that we call from the Dockerfile.

gabemontero:
[...]

PR #453, opened 2017-12-20
still link plugins in Dockerfile so plugins available to s2i assemble…

gabemontero:
… script in extension scenario

Fixes https://github.com/openshift/jenkins/issues/451

@bparees @jim-minter ptal

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -56,6 +56,11 @@ ADD release.version /tmp/release.version
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
+    # we symlink the rpm installed plugins from /usr/lib/jenkins to /opt/openshift/plugins so that
+    # future upgrades of the image and their RPM install automatically get picked by jenkins;
+    # we use symlinks vs. actual files to delineate whether the user has overridden a plugin (and
+    # by extension taken over its future maintenance)
+    for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
     chmod 775 /etc/passwd && \
     chmod -R 775 /etc/alternatives && \
     chmod -R 775 /var/lib/alternatives && \
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -212,11 +212,9 @@ for FILENAME in ${JENKINS_HOME}/plugins/*.jpi; do
     fi
 done
 
-# for rhel, some maintenance wrt copying plugins is done on every startup;  this logic
-# covers the following scenarios:
-# 1) this is the initial start of the image and the JENKINS_HOME/configured file is not
-# yet present
-# 2) this is a release upgrade, and new plugins need to be added
+# for rhel, while the plugins are initially linked to the image volume on initial start,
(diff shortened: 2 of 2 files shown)


gabemontero:
[test]

gabemontero:
@bparees - assuming we get to a consensus with this, do we want to initiate the flow for a redhat.access registry refresh ?

bparees:
> @bparees - assuming we get to a consensus with this, do we want to initiate the flow for a redhat.access registry refresh ?

yeah, we're gonna have to backport it to the 3.7 branch of this repo and then work with justin's team to spin another 3.7 rhel image.

Thanks for jumping in to fix it so quickly.

2/Dockerfile.rhel7
```diff
@@ -56,6 +56,11 @@ ADD release.version /tmp/release.version
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
+    # we symlink the rpm installed plugins from /usr/lib/jenkins to /opt/openshift/plugins so that
+    # future upgrades of the image and their RPM install automatically get picked by jenkins;
+    # we use synlinks vs. actual files to delineate whether the user has overriden a plugin (and
```
bparees: synlinks->symlinks

bparees: overriden->overridden

2/Dockerfile.rhel7
```diff
@@ -56,6 +56,11 @@ ADD release.version /tmp/release.version
...
+    # by extension taken over its future maintenance)
+    for FILENAME in /usr/lib/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
```
bparees: does this mean we can go back and remove the logic in "run" that copies from /usr/lib/jenkins?  (https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/run#L229)

gabemontero: the comment in `run` should be adjusted, but we still need that code, since a release upgrade could introduce new plugins and the PV needs to be updated ...the links here only account for the initial image start

gabemontero:
spelling / comments adjusted

2/Dockerfile.rhel7
bparees: hm.  So this logic symlinks from /usr/lib/jenkins to /opt/openshift/plugins so that the install_plugins.sh script (as executed in assemble) sees them?

and then the run script symlinks them from /usr/lib/jenkins to /var/lib/jenkins to Jenkins itself sees them?

gabemontero: yes and yes

and we tolerate the additional hop with rhel to facilitate a common code path with both centos and rhel for install_plugins.sh (why /usr is not used for centos was a decision that predated my forays into this stuff, or there is an obvious explanation as to why that I'm unaware of)

gabemontero: duh .... it was file permissions :-) ... the non-root user can modify /opt based on the fix permission stuff 

now, we've since modified permissions in /usr for the 32 bit vs. 64 bit stuff; so if we wanted, we could switch gears here and just use /usr

but not this PR :-)

bparees:
lgtm, merge whenever you're happy w/ the testing status.

gabemontero:
thanks ... yeah the upgrade scenario, issue inducing scenario, and automated tests all look good.

[merge]

PR #454, opened 2017-12-21
use v3.7 release tagged base image

bparees:
note that i'm intentionally not doing this for the rhel images.  it would require changes to our CI jobs that aren't worth doing since the rhel-built images produced from this repo aren't what we ship anyway.

fixes https://github.com/openshift/jenkins/issues/456

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin
+FROM openshift/origin:v3.7.0
 
 # Jenkins image for OpenShift
 #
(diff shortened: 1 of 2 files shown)


bparees:
@gabemontero ptal
[test]

bparees:
[merge]

gabemontero:
yep makes sense in hindsight

gabemontero:
should we tag the rhel image Dockerfile FROM clauses accordingly?

bparees:
> should we tag the rhel image Dockerfile FROM clauses accordingly?

see my comment in the PR summary.

gabemontero:
ah - sorry missed that

PR #462, opened 2018-01-02
add variable for additional Java user options

drhelius:
Added a new variable so it's possible to pass additional Java options to Jenkins, like those [useful for the Kubernetes plugin](https://github.com/jenkinsci/kubernetes-plugin#over-provisioning-flags) without disturbing or redefining other Java flags.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -326,6 +326,7 @@ if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
   set -x
   exec java $JENKINS_JAVA_OPTIONS -Duser.home=${HOME} \
             -Djavamelody.application-name=${JENKINS_SERVICE_NAME} \
+            $JENKINS_JAVA_OVERRIDES \
             -jar /usr/lib/jenkins/jenkins.war $JENKINS_OPTS "$@"
 fi
 


jim-minter:
Can you use the pre-existing $JENKINS_OPTS variable for this?  The intention is that this is appended to the command line such that it can override values previously specified on the command line, if necessary.

drhelius:
I'm afraid you can't, those are Java options that need to be passed before the ```-jar``` command.

jim-minter:
@drhelius please could you give me an example? (I'm no expert in this...)

jim-minter:
@drhelius ah, I see, the syntax is java [options] -jar file.jar [arguments]

jim-minter:
@bparees I'm wondering if JENKINS_OPTS was intended to do this and either never worked or hasn't worked for a very long time :disappointed: - do you think we should move it to before the -jar ?

bparees:
@jim-minter no, JENKINS_OPTS is where it's supposed to be (it's args to the jar file, effectively, ie arguments to jenkins itself, not to the JVM).

bparees:
but can't JAVA_TOOLS_OPTIONS be used for what @drhelius wants?

bparees:
or _JAVA_OPTIONS
https://stackoverflow.com/questions/28327620/difference-between-java-options-java-tool-options-and-java-opts

jim-minter:
Perhaps, but not without potential side-effects.  For the JNLP slave I added "JNLP_JAVA_OVERRIDES" for this; I didn't do so for the master because I wrongly thought that JENKINS_OPTS covered it.  I'm in favour of adding a $JENKINS_JAVA_OVERRIDES just before the -jar to match the JNLP slave behaviour.

bparees:
> Perhaps, but not without potential side-effects. 

side effects such as?  it seems like between those to env vars (one that gets overridden by other flags and one that overrides other flags) users should be able to get whatever jvm args they want to be set.

jim-minter:
> side effects such as

being applied to any subsequently spawned JVM.

bparees:
> being applied to any subsequently spawned JVM.

ah, true.  Based on that i'm ok w/ adding a JENKINS_JAVA_OVERRIDES to align w/ the slave behavior, as you proposed.

jim-minter:
@drhelius would #463 work for you?

drhelius:
Yes, it's perfectly fine, it let us pass JVM options to Jenkins without messing with the other variables. 
I fixed it in this PR as you proposed, can I contribute it?

bparees:
lgtm
[merge]

jim-minter:
Thanks @drhelius !

Issue #447, opened 2017-12-19
Update install-plugins script to support version specific update centers


bakito:
The install-plugins.sh from jenkinsci/docker does support version specific update centers for LTS versions. 
https://github.com/jenkinsci/docker/blob/678f3ec669666bada2f448f9da3f33c4eb0417a9/install-plugins.sh

        # If version-specific Update Center is available, which is the case for LTS versions,
        # use it to resolve latest versions.

It would be nice if this would be applied to the OpenShift Jenkins too.

bakito:
This issue hits me when extending the jenkins with s2i.

gabemontero:
thanks @bakito !

I should have a PR up shortly to verify that change is copacetic in our env.  Seems fine / safe upon visual examination.

bakito:
@gabemontero Thank you very much

gabemontero:
you're welcome @bakito and thanks again for alerting us to the update in jenkinsci/docker

bakito:
@gabemontero your pull request was incomplete, please see my pull request #466 for a fix

PR #466, opened 2018-01-04
fix version specific update center handling

bakito:
The solution for issue #447 #460 was missing the JENKINS_UC_LATEST handling from https://github.com/jenkinsci/docker/blob/678f3ec669666bada2f448f9da3f33c4eb0417a9/install-plugins.sh#L218-L225
Also fixed some tab -> space indents

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -114,7 +114,7 @@ copy_reference_file() {
             action=${action:-"INSTALLED"}
             log=true
             mkdir -p "$JENKINS_HOME/${dir:23}"
-	    # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
+            # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
             cp --remove-destination -r "${f}" "$JENKINS_HOME/${rel}";
             # pin plugins on initial copy
             touch "$JENKINS_HOME/${rel}.pinned"
@@ -129,7 +129,7 @@ copy_reference_file() {
             action="INSTALLED"
             log=true
             mkdir -p "$JENKINS_HOME/${dir:23}"
-	    # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
+            # if done on rhel, we may need to override a link to /usr/lib/jenkins, so include --remove-destination
             cp --remove-destination -r "${f}" "$JENKINS_HOME/${rel}";
         else
             action="SKIPPED"
@@ -147,6 +147,8 @@ copy_reference_file() {
 
 REF_DIR=${REF:-/opt/openshift/plugins}
 FAILED="$REF_DIR/failed-plugins.txt"
+
+JENKINS_WAR=/usr/lib/jenkins/jenkins.war
 
 function getLockFile() {
     echo -n "$REF_DIR/${1}.lock"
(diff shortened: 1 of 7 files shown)


2/contrib/jenkins/install-plugins.sh
```diff
@@ -343,14 +344,27 @@ function installedPlugins() {
     done
 }
 
+function jenkinsMajorMinorVersion() {
+    if [[ -f "$JENKINS_WAR" ]]; then
+        local version major minor
+        version="$($(alternatives --display java | awk '/family.*x86_64/ { print $1; }') -jar $JENKINS_WAR --version)"
```
bakito: @gabemontero java is needed here is there a better way to evaluate the java-binary in the s2i assemble step?

gabemontero: @bakito thanks for checking ... but we reference java in the same way in the s2i run script, though you need to account for the fact that we may be running with 32 bit

See https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/run#L116-L122

bparees: since we don't care about 32/64bit for a call like this, we should definitely just invoke java and get whatever we get.

bakito: java is not in the path during the assemble script, that is why I copied the alternatives call.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -347,7 +347,7 @@ function installedPlugins() {
 function jenkinsMajorMinorVersion() {
     if [[ -f "$JENKINS_WAR" ]]; then
         local version major minor
-        version="$($(alternatives --display java | awk '/family.*x86_64/ { print $1; }') -jar $JENKINS_WAR --version)"
+        version="$(/etc/alternatives/java -jar $JENKINS_WAR --version)"
```
bakito: java is not in the path in the assemble step, it is set in the run script https://github.com/openshift/jenkins/blob/626247110ad9c23e7e238185fd0dcba539179499/1/contrib/s2i/run#L18-L28

32/64 bit does not matter for just getting the version of the war

gabemontero: yep actually getting at java was also part of my reference in https://github.com/openshift/jenkins/pull/466#discussion_r159677972 @bakito - apologies if it was not clear

and yeah, if you've got an alternatives usage with your latest commit that does not awk on the family, all the better, and you do not have to be concerned with the 32 vs. 64 bit questions

I'll initiate a test momentarily.  Assuming it passes, we'll ask that you squash the commits and we'll merge.

gabemontero:
[test]

bakito:
@gabemontero  I'd like to add another plugin to the test plugin.txt with version latest. I think because only the ansicolor plugin with a specific version is installed,  the tests where fine in #460 and JENKINS_UC_LATEST was not used. I' ll try to extend the test later this day.

bakito:
[test]

gabemontero:
sounds good @bakito - and the last commit looked fine to me at first blush - let's see how the test looks

and it did not seem to fire when you posted the comment (hmm ... is there a permission thing I'm not aware of ???)

I'll try

[test]

bakito:
@gabemontero Test seems good. Is something else needed from me to merge?

gabemontero:
hey @bakito if you could squash your commits we'll be good to go (squashing commits is a dev process rule we have)

bakito:
@gabemontero squashed

gabemontero:
great - thanks @bakito 

[merge]

bakito:
@gabemontero Hi do you have any idea why the build failed?
the last time it was green, I do no changes since then.

I can only test centos and these tests are still sucessfully.

gabemontero:
On Fri, Jan 12, 2018 at 9:53 AM, Marc Brugger <notifications@github.com>
wrote:

> @gabemontero <https://github.com/gabemontero> Hi do you have any idea why
> the build failed?
> the last time it was green, I do no changes since then.
>
We just had another PR test fail as well.  Something underlying may be
afoot.  I've got
this on today's to-do list, and will report back when I have a chance to
take a look.

gabemontero:
Yeah a few more PRs have failed with various environmental issues.  In this particular case, we hit a devmapper space one:

```
devmapper: Thin Pool has 2100 free data blocks which is less than minimum required 2128 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior
make: *** [build] Error 1

```

The main openshift/origin jobs are experience various issues as well.

I'll retry this merge again in a bit, after our ci/cd team gets a chance to sort through a few things.

gabemontero:
[merge]

gabemontero:
[merge]

gabemontero:
ok, our ci system is healed, was able to merge @bakito (whew)

bakito:
@gabemontero thx a lot.
Do you have any plans on releasing a new rhel jenkins image soon?

gabemontero:
YW and thanks again for the contribution @bakito 

This will go into the upcoming 3.9 release for sure.

Otherwise, we generally release errata on a 3 week cycle.  I believe the next errata for 3.7 was scheduled for Jan 18.  This is most likely too late for that, so in theory it would be 3 weeks after that, if the errata goes out on time.

PR #485, opened 2018-01-19
adsf

gabemontero:
I've got an old note about a bot tag that would initiate ext tests for non-origin repos ... going to see if it is still valid

will nuke this PR afterward

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 Jenkins Docker Image
 ====================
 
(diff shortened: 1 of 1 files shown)


gabemontero:
[test][openshift]

gabemontero:
the answer is no, the [openshift] tag no longer has (if it ever did) any effect.

Issue #477, opened 2018-01-11
document the embedded oc compatability fallout 


gabemontero:
Follow up to the recent 3.7 GA fallout, and various openshift-sme email threads, including https://mail.google.com/mail/u/0/#label/openshift-sme/16036b7c1d422563

1) @bparees 's last summation was I *think* pretty concise and comprehensive:

```
There has been a bit of churn in this space lately, but the thing to check is what version of the oc client is in each of the slave images you're using.  The v3.7 oc client is *not* backwards compatible to 3.5.

The current state of the world should be:
jenkins-*-rhel7:v3.7 - contains oc 3.7 binary
jenkins-*-rhel7:v3.6 - contains oc 3.7 binary
jenkins-*-rhel7:latest - contains oc 3.6 binary  ("latest" points to the v3.6 image and is going to stay that way going forward).

jenkins-*-centos7:v3.7 - contains the oc 3.7 binary
jenkins-*-centos7:v3.6 - contains the oc 3.6 binary
jenkins-*-centos7:latest - contains the oc 3.7 binary (i know this seems inconsistent, but we don't really make compatibility guarantees about our centos images.)

On top of that, because the slave pod configurations do not specify "pull always" you may have older/different images on your nodes depending when the images were pulled.  We've just changed that default to pull always to avoid this problem in the future, but for you now you should update your own slave configurations to pull always to get consistency around what is being used on each node.

```

2) @bparees - was there a bugzilla or something already in play to handle this that I'm forgetting?  also, I'm thinking references the details here in this repo's readme and a link to it, referencing the issue generically, in doc.openshift.io.  Thoughts / is there another required path?

@bmcelvee fyi

bparees:
@gabemontero no existing BZ, putting it in the jenkins repo readme for now seems like a reasonable starting point.  There's probably something that could be said here too about being aware of what oc client your jenkins image contains (and aligning your jenkins image version w/ your cluster version):
https://docs.openshift.org/latest/using_images/other_images/jenkins.html

possibly in here:
https://docs.openshift.org/latest/using_images/other_images/jenkins.html#client-plugin-in

but possibly just in the general image overview since the oc client can be used outside of the plugin too.

Note that @jim-minter is currently reworking/refactoring some of that doc, so you might be able to talk him into adding some content, if not you'll want to wait until his changes land.

gabemontero:
sounds good / thanks @bparees 

[...]

PR #486, opened 2018-01-19
add clarification around oc version and image tags

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/477

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,38 @@ You can also set the following mount points by passing the `-v /host:/container`
 directory has the appropriate permissions and that the owner and group of the directory
 matches the user UID or name which is running inside the container.**
 
+Inclusion of the `oc` binary
+---------------------------------
+
+To assist in interacting with the OpenShift API server while using this image, the `oc` binary, the CLI command for OpenShift, has been
+installed in the master and slave images defined in this repository.
+
+
+However, it needs to be noted that backward compatibility is not guaranteed between different versions of `oc` and the OpenShift
+API Server.  As such, it is recommended that you align versions of this image present in the nodes of your cluster with your 
+OpenShift API server.  In other words, you should use the version specific tag instead of the `latest` tag.
+
+|  Jenkins image version      | `oc` client version      |
+| :-------------------------- | ------------------------ |
+|  `jenkins-*-centos7:v3.7`   | 3.7 `oc` binary          |
+|  `jenkins-*-centos7:v3.6`   | 3.6 `oc` binary          |
+|  `jenkins-*-centos7:latest` | `oc` binary from `docker.io/openshift/origin:latest` image          |
+|  `jenkins-*-rhel7:v3.7`     | 3.7 `oc` binary          |
+|  `jenkins-*-rhel7:v3.6`     | 3.6 `oc` binary          |
+|  `jenkins-*-rhel7:latest`   | 3.6 `oc` binary \*\*     |
+
+
+**Notice: the `latest` tag for the RHEL7 images will point to 3.6 indefinitely in order to support users on older clusters with older slave 
+configurations that point to the "latest" tag.  This way, they will have an older `oc` client which should be able to communicate with both 3.6
+and newer versions of OpenShift API Servers.  As the support policy is less stringent for the CentOS7 image, the `latest` tag there will 
+make the more obvious correlation to the latest built version of OpenShift (which can include pre-GA versions). 
+
+**Notice:  There is an additional consideration with the pod configurations for the Kubernetes Plugin; earlier versions of this image
+did not specify the "pull always" policy for the default agents/slaves configured.  As a result, users may have older/different images on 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,6 +96,37 @@ You can also set the following mount points by passing the `-v /host:/container`
 directory has the appropriate permissions and that the owner and group of the directory
 matches the user UID or name which is running inside the container.**
 
+Inclusion of the `oc` binary
+---------------------------------
+
+To assist in interacting with the OpenShift API server while using this image, the `oc` binary, the CLI command for OpenShift, has been
+installed in this image.
```
bparees: installed in the master and slave images defined in this repository.

README.md
```diff
@@ -96,6 +96,37 @@ You can also set the following mount points by passing the `-v /host:/container`
...
+
+However, it needs to noted that backward compatibility is not guaranteed between different versions of `oc` and the OpenShift
```
bparees: s/to noted/to be noted/

README.md
```diff
@@ -96,6 +96,37 @@ You can also set the following mount points by passing the `-v /host:/container`
...
+
+
+**Notice:  There is an additional consideration with the pod configurations for the Kubernetes Plugin; earlier versions of this image
+did not specify the "pull always" policy for the default agents/slaves configured.  As a result, users may have older/different images on 
+your nodes depending when the images were pulled.  Staring with the 3.7 release, the default changed to "pull always" to avoid this problem 
+in the future.  But if you started using this image prior to 3.7, verification of your Kubernetes plugin configurations for the image pull 
+policy used is warranted to guarantee consistency around what image is being used on each of your nodes.
+
+**Notice: the `latest` tag for the RHEL7 images will point to 3.6 for the time being in order to support users whose pull policies for
+their Kubernetes Plugin configuration lack the "pull always" setting.  This way, they will have an older `oc` client which should be able to
```
bparees: it's unrelated to pullalways.  The rhel7 images point to the 3.6 tag (not just for the time being but probably forever) to ensure users on older clusters and who have older slave configurations that point to ":latest" will continue to work (will continue to pick up the v3.6 oc binary).  

New slave configurations use version specific tags (as well as pullalways) to ensure the right image is used for their cluster.

README.md
```diff
@@ -96,6 +96,37 @@ You can also set the following mount points by passing the `-v /host:/container`
...
+
+However, it needs to noted that backward compatibility is not guaranteed between different versions of `oc` and the OpenShift
+API Server.  As such, it is recommended that you align versions of this image present in the nodes of your cluster with your 
+OpenShift API server.  In other words, you should use the version specific tag instead of the `latest` tag.
+
+|  Jenkins image version      | `oc` client version      |
+| :-------------------------- | ------------------------ |
+|  `jenkins-*-centos7:v3.7`   | 3.7 `oc` binary          |
+|  `jenkins-*-centos7:v3.6`   | 3.6 `oc` binary          |
+|  `jenkins-*-centos7:latest` | 3.9 `oc` binary          |
```
jim-minter: I think that if you state 3.9 here, it'll need to be changed next release.  Might be better to say "whatever the latest version is" (or words to that effect)?

README.md
```diff
@@ -96,6 +96,37 @@ You can also set the following mount points by passing the `-v /host:/container`
...
+|  `jenkins-*-rhel7:v3.7`     | 3.7 `oc` binary          |
+|  `jenkins-*-rhel7:v3.6`     | 3.6 `oc` binary          |
+|  `jenkins-*-rhel7:latest`   | 3.6 `oc` binary          |
```
jim-minter: Might be worth reversing the two paragraphs that follow and putting an asterisk here to link the following paragraph more closely

README.md
gabemontero: Sure ... it is literally "the version of oc present in the docker.io/openshift/origin:latest image at the time the jenkins images with the latest tag were built".

I'll work on some consumable form of that.  Suggestions are of course welcome.

README.md
gabemontero: I imagine there is some trick with markdown language to employ the asterisk like you are suggesting .... unless you know different, I'll start hunting for examples / doc references

README.md
gabemontero: I was trying to be as non-committal as possible wrt "time being" vs. "probably forever"; just seemed rash to say "forever"... perhaps just remove the "time being" phrase and not substitute any other indication of time frame ?

otherwise sure I'll remove the pull always piece, just reference "to ensure users on older clusters and who have older slave configurations that point to ":latest" will continue to work (will continue to pick up the v3.6 oc binary)."

bparees: Maybe "indefinitely" or "for the foreseeable future" instead.  "For the time being" implies to me that we have plans to change it at some point when we get to it, which we don't.

gabemontero: I'll go with indefinitely

gabemontero:
response to comments pushed in separate commits

jim-minter:
lgtm

gabemontero:
[merge]

Issue #481, opened 2018-01-16
Image tagging on docker hub


spicysomtam:
I notice on docker hub  that you have now started tagging the centos images with the openshift release. Thats good since there was no tagging before. But how about also tagging them with the Jenkins release too? As a devops engineer, the Jenkins version is pretty important, and tbh we need to upgrade fairly regularly the Jenkins release because of security vulnerabilities, etc. You can have multiple tags per image (in case you did not know this). Otherwise I need to download the image, spin it up, and then see what version of Jenkins it runs. Since openshift runs as non root it can be tricky updating the Jenkins.war to bring it up to date. Or you might add a wiki table somewhere which says image v3.6 runs Jenkins blah.

gabemontero:
Adding say a table to this repo's readme with the jenkins info seems reasonable.

As to pushing multiple tags for a given image ... I suspect we might be bumping up against openshift conventions, but don't know for sure .... before saying "no", any insight / opinions @bparees ?

bparees:
It wouldn't exactly cause a convention problem but i don't love it.

After we start making tags like "jenkins-2-centos7:openshiftv3.7-jenkinsv2.89" (which in itself is a mess) someone's going to start asking us to "please update the jenkins-2-centos7:v3.7-jenkinsv2.73" image to include the fix that was put into jenkins-2-centos7:v3.7-jenkinsv2.89".

The table is ok but someone's going to have to remember to update it every time we bump the LTS and we're probably going to forget to do that from time to time too.

Not to mention if you just look at:
https://github.com/openshift/jenkins/blob/master/2/Dockerfile#L37

You can see what LTS we're currently installing.  (jenkins-2.89.2-1.1 as of this comment).

spicysomtam:
I could see a problem where you have two version tags (say `v3.7` and `v3.9`)  and they have the same Jenkins version, so multiple tagging could be an issue? I suppose the problem is you want to use image `v3.7` but have the ability to upgrade the Jenkins version, since Jenkins is often updated due to security vulnerabilities. Since openshift runs as non root for containers its not easy to replace the Jenkins.war without building a new image? Suppose you could mount a volume where the war file lives and then update it that way?

bparees:
> Since openshift runs as non root for containers its not easy to replace the Jenkins.war without building a new image? 

Changing the war should always involve building a new image. Mounting the war from a volume is possible, but it's a far less stable/reproducible deployment model.  In general you want to treat your images as being as immutable as possible.

[...]

PR #491, opened 2018-01-23
Add JENKINS_RELEASE var, used for label and rpm

grdryn:
This is a proposal based on discussion in #481. It allows you to have
one place to update the version (the new env var), and it'll be used
for both the label and the RPM installation.

The idea is that one could query the Jenkins version without having to
pull the image. I might be completely misunderstanding what's desired
in #481, but I suggested that one could use skopeo to check a label
rather than pulling the image and running it, so I figured I'd give
this example of how the label could exist, and still only have one
place to update.


grdryn:
@gabemontero I'm sure if the label is at all useful, and if it is you've probably already thought of this way. I figured I'd put this example forward just in case :)

PR #493, opened 2018-01-24
pull in 3.7 9b5f7bf 7ef705b 9db8ae2

gabemontero:
@jupierce @adammhaile @bparees for recent brew failures

in addition to the raw noarch change, I pulled in migration fixes we had to drop in 3.7 after initially releasing with just the raw noarch change

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -51,9 +51,11 @@ ADD ./contrib/s2i /usr/libexec/s2i
 # and its presence in the rhel image is confusing.
 RUN rm /opt/openshift/base-plugins.txt && \ 
     mkdir -p /opt/openshift/plugins && \
-    # NOTE: When adding new Jenkins plugin, you have to create the symlink for the
-    # HPI file created by rpm to /opt/openshift/plugins folder.
-    for FILENAME in /usr/lib64/jenkins/*hpi ; do ln -s $FILENAME /opt/openshift/plugins/`basename $FILENAME .hpi`.jpi; done &&\
+    # we symlink the rpm installed plugins from /usr/lib/jenkins to /opt/openshift/plugins so that
+    # future upgrades of the image and their RPM install automatically get picked by jenkins;
+    # we use symlinks vs. actual files to delineate whether the user has overridden a plugin (and
+    # by extension taken over its future maintenance)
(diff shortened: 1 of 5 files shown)


gabemontero:
[test]

gabemontero:
Well our update of the aos-cd-jobs to tag the rhel7 image seems to have helped;  we no longer get the rhel subscription problems that we saw with our last 3.6 PR.

However the test of the registry.access.redhat.com/openshift3/jenkins-2-rhel7-candidate image failed, and in my initial scan of the console logs it is not obvious to me as to why.  The dumps of the container logs look OK to me.

Diving into the test scripts, will report back when I think I found something.

gabemontero:
ahh .... forgot, we did not deprecate the jenkins v1 image until 3.7 ... so the 3.6 job is looking into both of them

the PR test job is failing on the v1 image ... hasn't gotten to the v2 image currently affected by this change

I'll diff the files in question, see how painful making the same changes to v1 is 

but a due diligence question @bparees  .... does our deprecation of jenkins v1 apply to errata of 3.6?  instinctual response for me is "no" but don't wan to presume ...

gabemontero:
... it would be nice though :-)

merging the requisite noarch changes into the jenkins v1 context dir is not going to be easy-peasy

gabemontero:
for kicks in the interim ... going to adjust the makefile to ignore v1

bparees:
> but a due diligence question @bparees .... does our deprecation of jenkins v1 apply to errata of 3.6? instinctual response for me is "no" but don't wan to presume ...

I would say effectively yes. If we hadn't hit the issues we did which forced us into unique version tags, we would have never updated the jenkins-1 latest tag again anyway.

So I don't see a reason to go updating the jekins-1 v3.6 image now.

gabemontero:
OK ... turns out the comment in the xml file is getting stripped ... @bparees remembers this now ... see https://github.com/openshift/jenkins/commit/7f4a28f219f1478ad74ae06eac29bf9ef8019047

I'll add that label change, clean up my PR debug, try again

gabemontero:
ok the test is still cooking but we got past the problem noted above ... in fact all the tests have passed for rhel and the centos tests look good so far

most like I'll merge this after the test is complete

@smunilla per your inquiry this AM ... should be good to go ^^

gabemontero:
[merge]

PR #516, opened 2018-02-28
Fix migration rhel

gabemontero:
@openshift/sig-developer-experience fyi / ptal

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1550193

turns our a post 3.7 GA fix broke previous fixes to git image migration working with the openshift jenkins rhel image on PVs

i.e.
1) someone has deployed the 3.6 or 3.7 image to a PV
2) they then update their image stream to point to the 3.9 image

@bparees I don't have the env override we just talked about in place yet; I wanted to start getting input on the core changes, and validate the manual test runs from the AWS rhel instance match what the CI system will produce.

I'll add the env override (including README) once we get over the hump with the initial review

NOTE: will also need to backport this to 3.6 and 3.7 branches.


gabemontero:
[test]

2/contrib/s2i/run
```diff
@@ -105,6 +105,28 @@ function migrate_slave_configuration() {
   sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
 }
 
+function force_copy_plugins() {
+   # if rhel
+   if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
+      echo "This openshift jenkins image is starting in a deployment where the image version"
+      echo "was not recoreded.  A one time replacement of all the plugins from the image to the"
```
bparees: s/recoreded/recorded/

2/contrib/s2i/run
```diff
@@ -235,8 +277,39 @@ if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
           continue
         fi
 	echo "Linking ${FILENAME} RPM installed Jenkins plugins to ${JENKINS_HOME} ..."  
+        # note note a force link, to not override upgrades when not upgrading the image
```
bparees: s/note note/note not/

2/contrib/s2i/run
```diff
@@ -235,8 +277,39 @@ if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
...
+   # extension scenario, so that the existing plugins are available during 
+   # the s2i assemble step (which precedes the s2i run step handled by this
+   # script); however, we must keep any plugins pulled in by s2i assemble, so
+   # we only remove files which are the same as /usr/lib/jenkins
+   #
+   # also, this will preserve the links if nothing has changed so that subsequent
+   # release to release migrations can work since the links will result in picking
+   # up the new content in /usr/lib/jenkins
+   if [ ! -e ${JENKINS_HOME}/configured ]; then
+      echo "Cleaning up links to base image plugins in /opt/openshift/plugins as part of this initial image start"
```
bparees: this won't accomplish anything.  Every time the container starts from the image again, these files will be back.

2/contrib/s2i/run
```diff
@@ -281,6 +354,8 @@ else
   fi  
 fi
 
+echo ${OPENSHIFT_JENKINS_IMAGE_VERSION} > ${JENKINS_HOME}/image-version
```
bparees: presumably could do this up where you're checking for this value.  (if it's not there/doesn't match, do this)

2/contrib/s2i/run
```diff
@@ -105,6 +105,28 @@ function migrate_slave_configuration() {
...
+      echo "This openshift jenkins image is starting in a deployment where the image version"
+      echo "was not recoreded.  A one time replacement of all the plugins from the image to the"
+      echo "jenkins plugin directory will occur."
+      echo ""
+      ls -la /usr/lib/jenkins
+      for FILENAME in /usr/lib/jenkins/*.hpi ; do
+         basefilename=`basename $FILENAME .hpi`
+         plugin_image_volume_location="${JENKINS_HOME}/plugins/${basefilename}.jpi"
+         echo "Force linking ${FILENAME} to ${plugin_image_volume_location}"
+         ln -sf $FILENAME $plugin_image_volume_location
```
bparees: I guess having symlinks becomes somewhat moot in a world where you're always overwriting existing plugins w/ the plugins from the image, if the image is new (and hopefully not overwriting anything if the image is not new).  So does it make sense to make this a straight copy, so that this and the centos image behavior are aligned?

bparees:
hm, maybe the reason that won't work is that install-plugins.sh will pull down dependencies that might already exist in /var/lib/jenkins, if we wipe out /opt/openshift/plugins before running it.

Needs some more thought.

2/contrib/s2i/run
```diff
@@ -212,7 +234,27 @@ for FILENAME in ${JENKINS_HOME}/plugins/*.jpi; do
...
+   # then do the one time replace
+   currentversion=`cat ${JENKINS_HOME}/image-version`
+   if [ $currentversion != $OPENSHIFT_JENKINS_IMAGE_VERSION ]; then
+      force_copy_plugins
+   fi
+fi
+
+# otherwise, for rhel, while the plugins are initially linked to the image volume on initial start,
 # some maintenance wrt transferring plugins is done on every startup;  this logic
 # covers the case of a release upgrade that introduces new plugins.
```
bparees: is this actually still necessary then, in light of the other logic added above?

2/contrib/s2i/run
bparees: also this message is going to be printed when the imageversion is older than the current image also, so the text needs to be reworked.

2/contrib/s2i/run
```diff
@@ -105,6 +105,28 @@ function migrate_slave_configuration() {
...
 }
 
+function force_copy_plugins() {
+   # if rhel
+   if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
+      echo "This openshift jenkins image is starting in a deployment where the image version"
+      echo "was not recoreded.  A one time replacement of all the plugins from the image to the"
+      echo "jenkins plugin directory will occur."
+      echo ""
+      ls -la /usr/lib/jenkins
```
bparees: is this for debug or are you intending to leave it in?

gabemontero: nope suppose to remove ... part of next push

2/contrib/s2i/run
gabemontero: Message reworded, including dumping of the image and deployment versions.

part of next push

2/contrib/s2i/run
gabemontero: part of next push

2/contrib/s2i/run
gabemontero: disagree ... as testing verified, if we don't do this here, then the install_plugin call below will remove the the linking between /var/lib/jenkins/plugins and /usr/lib/jenkins ... I am pretty sure removing this commit is not viable given the second commit (but I did not explicitly try that).

To reword your review summary comments, the complexity in managing the multitude of injection mechanisms for plugins makes it difficult to keep track.

Most likely there are short term (get the migration functional) vs. longer term (redo this whole mess) approaches we could employ here.

I at least do not think it is advisable to tackle the longer term approach.

But yeah, walking through some things face to face, and maybe watching the tests progress, might be necessary for this one given the undo complexity.

2/contrib/s2i/run
gabemontero: Yeah I believe your comment below answers this one.  But certainly talking through it face to face to be sure may be prudent here.

2/contrib/s2i/run
gabemontero: To capture new plugins introduced, yes.  My other explanations elsewhere do lend relevant context to this statement.

2/contrib/s2i/run
```diff
@@ -235,8 +277,39 @@ if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then
...
+            rm -f $FILENAME
+         fi
+      done
+      if [ "$(ls /opt/openshift/plugins/* 2>/dev/null)" ]; then
+         echo "Remaining plugins in /opt/openshift/plugins are $(ls /opt/openshift/plugins)"
+      fi
+   fi
 fi
 
 if [ ! -e ${JENKINS_HOME}/configured ]; then
```
bparees: why didn't you just put the above block inside here, rather than introducing a second "if not configured" block?

gabemontero: Could have done that yeah, but would have had to duplicate the `if [ "$(ls /usr/lib/jenkins/*.hpi 2>/dev/null)" ]; then` if check that this is currently in.

Seems like a net zero gain to me.

bparees: ah, yep, missed the additional outer if block.

2/contrib/s2i/run
gabemontero: OK confirmed, this block is needed for bringing up this image initially.  Otherwise, the plugins in /var/lib/jenkins/plugins are not linked to /usr/lib/jenkins.

The `install_plugins` call below does a force copy from /opt/openshift/plugins as part of possible additions.  However, in the case of a new install, there are links for all the pluings in /opt/openshift/plugins to assist the s2i extension scenario noted above.  Trying to confirm the s2i extension scenario did not add anything is essentially this same or similar logic in that path.  This way seemed more intuitive.

I can also confirm this block is *NOT* needed in the upgrade scenario (i.e. we brought up jenkins on a PV with the 3.6 image, then upgraded).

gabemontero:
OK pushed the message changes

gabemontero:
@bparees - pushed the update stemming from our code walk through in separate commit

2/contrib/s2i/run
bparees: bump

bparees: (just trying to get all the version marking logic as close together as possible)

bparees:
ok i think i've walked through at least the rhel codepath in the run script and convinced myself of its sanity.

i wrote up a little walkthrough as i was going through the code, @gabemontero, maybe you can confirm this looks like what you think it's supposed to be doing:

```
remove stale/dead symlinks from /var/lib/jenkins (plugins we've removed from /usr/lib/jenkins)

if version change or no version present:
  - force a link from /usr/lib to /var/lib
  - remove /opt/openshift/plugins/filename (for every plugin we linked above)

if plugin in /usr/lib and missing from /var/lib
  - link /usr/lib to /var/lib

first time(!configured)
  - remove /opt/openshift/plugins that exists(w/ matching checksum) in /usr/lib
  - leave modified /opt/openshift/plugins (we'll copy them later)

first time or OVERRIDEPLUGINS
  - install plugins from env variable to /opt/openshift/plugins
  - forcibly copy all content from /opt/openshift/plugins to /usr/lib
  - clear /opt/openshift/plugins (pointless, I think, but we can revisit that in the future)
```

2/contrib/s2i/run
gabemontero: sorry missed that one ... agreed, will push momentarily

gabemontero:
OK, pushed the commit to move the setting of `image-version`.

On @bparees 's write up, I think it is spot on but for one minor correction.

Instead of 

`- forcibly copy all content from /opt/openshift/plugins to /usr/lib`

it should be 

`- forcibly copy all content from /opt/openshift/plugins to /var/lib/jenkins/plugins`

gabemontero:
Pending any other tweaks to the rhel side, I'll still try a centos based set of tests around upgrade to make sure we are good there before we merge this.

Thanks.

bparees:
> On @bparees 's write up, I think it is spot on but for one minor correction.

sorry, yeah.  going to edit my write up for posterity.

gabemontero:
Ok I just completed the centos7 based migration testing.  I've pushed a new commit with a couple of small tweaks to facilitate that.

@bparees - ptal and if you are good, I'll squash and merge.  Thanks.

gabemontero:
Ok I just completed the centos7 based migration testing.  I've pushed a new commit with a couple of small tweaks to facilitate that.

@bparees - ptal and if you are good, I'll squash and merge.  Thanks.

bparees:
still needs logic to let users opt-out of having everything stomped?

gabemontero:
yep forgot ... working on that now

gabemontero:
override pushed

bparees:
lgtm

gabemontero:
cool - thx ... running a final regression of rhel/centos now

assuming passes, will squash and merge

gabemontero:
ah regression testing .. problem with the new flag ... the test on it needed to be `-z`, not `! -z` 

from the bash manual:

```
-z string

    True if the length of string is zero.

```

gabemontero:
regression test now passes

[merge]

gabemontero:
the new go tests timed out for some reason I'm having trouble parsing

running locally first, see if I see the same thing

will check back in later this evening

gabemontero:
On Fri, Mar 2, 2018 at 5:46 PM OpenShift Bot <notifications@github.com>
wrote:

> continuous-integration/openshift-jenkins/test SUCCESS (
> https://ci.openshift.redhat.com/jenkins/job/test_pull_request_jenkins_images/179/)
> (Base Commit: 814dfbb
> <https://github.com/openshift/jenkins/commit/814dfbb824cc324ba504aa9e463de2abafaa3284>)
> (PR Branch Commit: 4b4c6ee
> <https://github.com/openshift/jenkins/commit/4b4c6ee00f0ae953e06e691339f900c4ee16a444>
> )
>
ah but the test passed

Perhaps our test is just long enough now
That we are skirting the current timeout

gabemontero:
[merge]

PR #534, opened 2018-03-07
add docker related debug around s2i test

gabemontero:
@bparees ptal

this is the debug change to this repo's tests I mentioned to you earlier

highlights:
0) if the `s2i` invocation completes within a minute, nothing is printed
1) otherwise, it prints out the results from ps -ef within each container it finds
2) it prints out with nice formatting the json from `docker inspect`

like I mentioned, while it certainly would not have uncovered the docker flaw @jim-minter found, this changes would have ruled out a couple of our theories prior to that

### DIFF
--- a/2/test/jenkins_test.go
+++ b/2/test/jenkins_test.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"net/http"
@@ -187,6 +188,12 @@ var _ = Describe("Jenkins testing (v2)", func() {
 		By("running s2i build")
 		destImage := fmt.Sprintf("jenkins-test-s2i-%d", rand.Intn(1e9))
 
+		By("set up docker debug")
+		ctx, cancel := context.WithTimeout(context.Background(), 25*time.Minute)
+		cmdstrs := []string{"ps", "-ef"}
+		go dockercli.ExecInActiveContainers(GinkgoWriter, ctx, cmdstrs)
+		go dockercli.InspectActiveContainers(GinkgoWriter, ctx)
+
 		cmd := exec.Cmd{
 			Path: s2i,
 			Args: []string{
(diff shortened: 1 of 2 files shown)


bparees:
it still feels strange to me that we're putting logic into our jenkins tests to diagnose potential bugs in s2i or docker(and a pretty narrow case at that), but i won't block it.

2/test/jenkins_test.go
```diff
@@ -187,6 +188,12 @@ var _ = Describe("Jenkins testing (v2)", func() {
 		By("running s2i build")
 		destImage := fmt.Sprintf("jenkins-test-s2i-%d", rand.Intn(1e9))
 
+		By("set up docker debug")
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute*time.Duration(25))
```
bparees: I think `25*time.Minute` is sufficient

gabemontero:
sometimes you need to sort out if your dependencies are breaking you, no ?

2/test/jenkins_test.go
gabemontero: yep will update

gabemontero:
[merge]

bparees:
> sometimes you need to sort out if your dependencies are breaking you, no ?

ideally that sort of thing would be caught in the s2i repo by the s2i tests is my point.

The reason it wasn't caught in this case is mostly because it was a combination of dependency updates, not all of which went through the s2i PR process (namely the AMI update to docker 1.13).

The lesson here is that when we saw s2i command line builds hanging, we should have immediately attempted to directly reproduce the s2i build hands manually.  That said, because of the timing of docker 1.12/1.13, we likely still would have said "yup, it hangs on 1.13.  Nope, it's not hanging on 1.12.  must be a 1.13 problem" instead of the actual root cause which was "it was a docker client change that didn't cause problems until we updated to docker 1.13"

gabemontero:
yep fair enough ... but yep, as the redundancy and change here is small and logging usually benign ... (as you can see I posted merge)

curious if we get a hang while this merges and if the debug has any inkling as to what is up that I may have missed while testing

PR #537, opened 2018-03-08
New agents

gabemontero:
For https://trello.com/c/YQm7osws/1493-5-jenkins-slave-scl-dep-updatesjenkinsintegration

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -35,13 +35,13 @@ function docker_build_with_version {
 dirs=${VERSION:-$VERSIONS}
 
 # enforce building of the slave-base image if we're building any of
-# the slave images.  Note that we might build the slave-base
+# the slave/agent images.  Note that we might build the slave-base
 # twice if it was explicitly requested.  That's ok, it's
 # cheap to build it a second time.  The important thing
 # is we have to build it before building any other
 # slave image.
 for dir in ${dirs}; do
-  if [[ "$dir" =~ "slave" ]]; then
+  if [[ "$dir" =~ "slave" || "$dir" =~ "agent" ]]; then
     dirs=( "slave-base ${dirs[@]}")
     break
   fi
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -36,6 +36,7 @@ echo "OPENSHIFT_JENKINS_JVM_ARCH='${OPENSHIFT_JENKINS_JVM_ARCH}', CONTAINER_MEMO
 
 # Configure the slave image
 source /usr/local/bin/configure-slave
+source /usr/local/bin/configure-agent
 
 set -e
 
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 # version (at least that is the initial goal).  This naming system
 # can be revisited in the future if we decide we need either jenkins
 # or <platform> version numbers in the names.
-VERSIONS="2 slave-base slave-maven slave-nodejs"
+VERSIONS="2 slave-base slave-maven slave-nodejs agent-maven-3.5 agent-nodejs-8"
 
 ifeq ($(TARGET),rhel7)
 	OS := rhel7
--- a/README.md
+++ b/README.md
@@ -6,7 +6,10 @@ use with [OpenShift v3](https://github.com/openshift/origin)
 
 For an example of how to use it, [see this sample.](https://github.com/openshift/origin/blob/master/examples/jenkins/README.md)
 
-The images are pushed to DockerHub as openshift/jenkins-2-centos7, openshift/jenkins-slave-base-centos7, openshift/jenkins-slave-maven-centos7, and openshift/jenkins-slave-nodejs-centos7.
+The images are pushed to DockerHub as openshift/jenkins-2-centos7, openshift/jenkins-slave-base-centos7, openshift/jenkins-slave-maven-centos7, openshift/jenkins-slave-nodejs-centos7, openshift/jenkins-agent-35-centos7, and openshift/jenkins-agent-nodejs-8-centos7.
+
+The slave-maven and slave-nodejs for both centos7 and rhel7  are being deprecated as part of v3.10 of OpenShift.
+Additionally, development of these images will cease as of v3.10.
 
 Development of the V1 Jenkins images has ceased.
 
@@ -37,6 +40,8 @@ Choose either the CentOS7 or RHEL7 based image:
     * https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/jenkins-slave-base-rhel7
     * https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7
     * https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7
+    * https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/jenkins-agent-maven-35-rhel7
+    * https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/jenkins-agent-nodejs-8-rhel7
 
     To build a RHEL7 based image, you need to run Docker build on a properly
     subscribed RHEL machine.
--- a/agent-maven-3.5/Dockerfile
+++ b/agent-maven-3.5/Dockerfile
@@ -1,0 +1,36 @@
+FROM openshift/jenkins-slave-base-centos7
+
+MAINTAINER Gabe Montero <gmontero@redhat.com>
+
+ENV MAVEN_VERSION=3.5 \
+    GRADLE_VERSION=4.2.1 \
+    BASH_ENV=/usr/local/bin/scl_enable \
+    ENV=/usr/local/bin/scl_enable \
+    PROMPT_COMMAND=". /usr/local/bin/scl_enable" \
+    PATH=$PATH:/opt/gradle/bin
+
+# Install Maven
+RUN INSTALL_PKGS="java-1.8.0-openjdk-devel.x86_64 java-1.8.0-openjdk-devel.i686 rh-maven35*" && \
+    yum install -y centos-release-scl-rh && \
+    yum install -y --enablerepo=centosplus $INSTALL_PKGS && \
+    curl -LOk https://services.gradle.org/distributions/gradle-${GRADLE_VERSION}-bin.zip && \
+    unzip gradle-${GRADLE_VERSION}-bin.zip -d /opt && \
+    rm -f gradle-${GRADLE_VERSION}-bin.zip && \
+    ln -s /opt/gradle-${GRADLE_VERSION} /opt/gradle && \
+    rpm -V ${INSTALL_PKGS//\*/} && \
+    yum clean all -y && \
+    mkdir -p $HOME/.m2 && \
+    mkdir -p $HOME/.gradle
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ADD contrib/bin/scl_enable /usr/local/bin/scl_enable
+ADD contrib/bin/configure-agent /usr/local/bin/configure-agent
+ADD ./contrib/settings.xml $HOME/.m2/
+ADD ./contrib/init.gradle $HOME/.gradle/
+
+RUN chown -R 1001:0 $HOME && \
+    chmod -R g+rw $HOME
+
+USER 1001
--- a/agent-maven-3.5/Dockerfile.rhel7
+++ b/agent-maven-3.5/Dockerfile.rhel7
@@ -1,0 +1,40 @@
+FROM registry.access.redhat.com/openshift3/jenkins-slave-base-rhel7
+
+MAINTAINER Gabe Montero <gmontero@redhat.com>
+
+# Labels consumed by Red Hat build service
+LABEL com.redhat.component="jenkins-slave-maven-rhel7-docker" \
+      name="openshift3/jenkins-slave-maven-rhel7" \
+      version="3.10" \
+      architecture="x86_64" \
+      release="4" \
+      io.k8s.display-name="Jenkins Agent Maven" \
+      io.k8s.description="The jenkins agent maven image has the maven tools on top of the jenkins slave base image." \
+      io.openshift.tags="openshift,jenkins,agent,maven"
+
+ENV MAVEN_VERSION=3.5 \
+    BASH_ENV=/usr/local/bin/scl_enable \
+    ENV=/usr/local/bin/scl_enable \
+    PROMPT_COMMAND=". /usr/local/bin/scl_enable"
+
+# Install Maven
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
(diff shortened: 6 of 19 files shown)


gabemontero:
Note I have manually run this new images with our sample pipeline jobs that leverage nodejs/maven agents.

2/contrib/jenkins/kube-slave-common.sh
```diff
@@ -24,12 +24,12 @@ JNLP_PORT=${!T_PORT}
...
-NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7:${JENKINS_SLAVE_IMAGE_TAG}}
-MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7:${JENKINS_SLAVE_IMAGE_TAG}}
+NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-registry.access.redhat.com/openshift3/jenkins-agent-nodejs-8-rhel7:${JENKINS_SLAVE_IMAGE_TAG}}
+MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-registry.access.redhat.com/openshift3/jenkins-agent-maven-35-rhel7:${JENKINS_SLAVE_IMAGE_TAG}}
 # if the master is running the centos image, use the centos slave images.
 if [[ `grep CentOS /etc/redhat-release` ]]; then
-  NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-openshift/jenkins-slave-nodejs-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
-  MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-openshift/jenkins-slave-maven-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
+  NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-openshift/jenkins-agent-nodejs-8-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
+  MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-openshift/jenkins-agent-maven-35-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
```
bparees: suggest you break this out to a separate PR so we can first merge a PR that starts building/publishing the new images, then update the jenkins master to actually consume them after we are sure they exist (and work).

gabemontero:
[(5) Jenkins slave SCL dep updates[jenkins_integration]](https://trello.com/c/YQm7osws/1493-5-jenkins-slave-scl-dep-updatesjenkinsintegration)

gabemontero:
[(5) Jenkins slave SCL dep updates[jenkins_integration]](https://trello.com/c/YQm7osws/1493-5-jenkins-slave-scl-dep-updatesjenkinsintegration)

coreydaley:
Should we rename the base to jenkins-agent-base-centos7 ?

gabemontero:
Would cause more churn than it is worth IMO @coreydaley

2/contrib/s2i/run
```diff
@@ -102,12 +102,12 @@ function install_plugins() {
...
 
   # replace registry.access.redhat.com/openshift3/jekins-slave-xxxx-rhel7[:optionaltag] with registry.access.redhat.com/openshift3/jenkins-slave-xxxx-rhel7:VersionTag
-  sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
-  sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-slave-nodejs-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-agent-maven-35-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-agent-maven-35-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
+  sed -i "s#<image>registry.access.redhat.com/openshift3/jenkins-agent-nodejs-8-rhel7\(:.*\)\{0,1\}</image>#<image>registry.access.redhat.com/openshift3/jenkins-agent-nodejs-8-rhel7:${JENKINS_SLAVE_IMAGE_TAG}</image>#" /var/lib/jenkins/config.xml
```
bparees: We need to think through this.

For someone who already has a `registry.access.redhat.com/openshift3/jenkins-slave-maven-rhel7:v3.x` reference in their PV, what do we do?

1) we know we don't want to move them to maven35 (might not be compatible)
2) at some point (probably after v3.10) we won't have a new maven:versiontag image to migrate them to

But at least for v3.10, I think we still want to leave the existing migration logic in place so that someone who's got a PV that references the :v3.7 or :v3.9 tag, will get updated to the maven:v3.10 tag (not maven35:v3.10).

When we ship a new jenkins master tag and do NOT ship a new corresponding slave versiontag, we'll need to remove the old migration logic.  I am assuming we are still going to ship a maven:v3.10 tag(likely the last time we do so).

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -36,6 +36,7 @@ echo "OPENSHIFT_JENKINS_JVM_ARCH='${OPENSHIFT_JENKINS_JVM_ARCH}', CONTAINER_MEMO
 
 # Configure the slave image
 source /usr/local/bin/configure-slave
+source /usr/local/bin/configure-agent
```
bparees: isn't this going to cause file not found errors?  (might not fail the script but it's going to look ugly).

how about wrapping each call if a test to see if the file exists and only source it if so?

2/contrib/s2i/run
bparees: (leave the existing migration sed logic *and* add your new sed invocations is what i am suggesting)

slave-base/contrib/bin/run-jnlp-client
gabemontero: there were no error messages that I saw when running the images manually, looking at pod logs, so I kept it vanilla

2/contrib/jenkins/kube-slave-common.sh
gabemontero: ah good point ... will do

2/contrib/s2i/run
gabemontero: ok yep  ... the background/scenario coupled with the "leave existing and add new sed" was helpful thanks

slave-base/contrib/bin/run-jnlp-client
bparees: i'm skeptical
```
$ cat /tmp/t.sh 
#!/bin/sh

source foo
```

```
$ /tmp/t.sh 
/tmp/t.sh: line 3: source: foo: file not found
```


slave-base includes a dummy configure-slave:
https://github.com/openshift/jenkins/tree/master/slave-base/contrib/bin

so you won't see it for your new agent images.

but you will see the error on the old slave images that don't have a configure-agent, unless you add a dummy one.

gabemontero: ah didn't catch what slave-base did ... ok that would explain things

OK ... so from a bash pespective and the if test, do you mean:

```
if ! source /usr/local/bin/configure-agent ; then
   source /usr/local/bin/configure-slave
fi
```
or some such?

bparees: i meant 
`
if [ -e /usr/local/bin/configure-slave ]; then
  source /usr/local/bin/configure-slave
fi
`

but i think i'd rather you just introduce the dummy file in the base image like we did for configure-slave.

gabemontero:
@stevekuznetsov - first two questions / request around our migration to prow

1) take a look at the contents of https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_jenkins/537/test_pull_request_jenkins_images/191/ ... how can we get that to be populated with more useful info like we see in the origin jobs?

2) what is this "tide" thing under the checks?  do we need it and if so how do we get in the tide pool?  if we don't need it, can we get it removed?

gabemontero:
On the image test failure, it was our hang running s2i cause of the docker client failure to impl an interface.

@bparees just updated s2i 2 days ago for this.

I believe this is either a question of needing an AMI update to get that fix, or maybe an aos-cd-jobs update ... I don't remember which.  @stevekuznetsov @bparees ? 

I believe @coreydaley saw this as well with a sync plugin test.

gabemontero:
Note https://github.com/openshift/jenkins/pull/538 did not hit the docker/s2i error ... don't know if that is timing (i.e. it got the s2i update for @bparees 's fix) or the fact that the docker/s2i error can be intermittent.

bparees:
> I believe this is either a question of needing an AMI update to get that fix, or maybe an aos-cd-jobs update ... I don't remember which. @stevekuznetsov @bparees ?

if i recall correctly this test job pulls down + builds s2i itself from master.  so i would not expect that to be the issue if the test ran after the fix merged to s2i.

gabemontero:
Here is the bash snippet @bparees :+1: 
```
#!/bin/bash
SCRIPT_START_TIME="$( date +%s )" && export SCRIPT_START_TIME && echo "########## STARTING STAGE: BUILD AND INSTALL S2I ##########" && trap 'export status=FAILURE' ERR && trap 'set +o xtrace; SCRIPT_END_TIME="$( date +%s )"; ELAPSED_TIME="$(( SCRIPT_END_TIME - SCRIPT_START_TIME ))"; echo "########## FINISHED STAGE: ${status:-SUCCESS}: BUILD AND INSTALL S2I [$( printf "%02dh %02dm %02ds" "$(( ELAPSED_TIME/3600 ))" "$(( (ELAPSED_TIME%3600)/60 ))" "$(( ELAPSED_TIME%60 ))" )] ##########"' EXIT && set -o errexit -o nounset -o pipefail -o xtrace && if [[ -s "${WORKSPACE}/activate" ]]; then source "${WORKSPACE}/activate"; fi
script="$( mktemp )"
cat <<SCRIPT >"${script}"
#!/bin/bash
set -o errexit -o nounset -o pipefail -o xtrace
cd "\${GOPATH}/src/github.com/openshift/source-to-image"
make build
sudo cp /data/src/github.com/openshift/source-to-image/_output/local/bin/linux/amd64/s2i /usr/bin
SCRIPT
chmod +x "${script}"
scp -F ./.config/origin-ci-tool/inventory/.ssh_config "${script}" openshiftdevel:"${script}"
ssh -F ./.config/origin-ci-tool/inventory/.ssh_config -t openshiftdevel "bash -l -c \"timeout 300 ${script}\"" 
```

I don't see an explicit download / git clone

bparees:
@gabemontero the repo sync is implicit:
https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/common/test_cases/s2i.yml#L7-L8

bparees:
@gabemontero @stevekuznetsov can confirm, but that should be syncing the latest s2i repo to what is in the ami before building.

gabemontero:
OK the easier answer would be there is a delay in synching the repo and that there isn't still a docker issue in s2i ;-)

bparees:
@gabemontero can you point me to the jenkins job that you think hit the s2i hang issue?

gabemontero:
it was at https://ci.openshift.redhat.com/jenkins/job/test_pull_request_jenkins_images/191/consoleFull#-121518783756bf4006e4b05b79524e5923 @bparees 

Note this from the gostacks:

```
os/exec.(*Cmd).Run(0xc420448160, 0xc4201e8460, 0xc42000e450)
	/usr/lib/golang/src/os/exec/exec.go:289 +0x5c
github.com/openshift/jenkins/2/test.glob..func2.7()
	/data/src/github.com/openshift/jenkins/2/test/jenkins_test.go:211 +0x4e7
```

Also, the test has, fwiiw, the periodic `ps -ef` and `docker inspect` invocations ... of course that wasn't pertinent for the missing method item ... but if this is in fact something else, maybe they have value.

bparees:
I sshe'd into another ec2 instance, one running the wildfly push job which also syncs s2i and indeed the source is not updated.  @stevekuznetsov this job:
https://ci.openshift.redhat.com/jenkins/job/push_wildfly_images/9/consoleFull

should have synced the s2i repo:
```
########## FINISHED STAGE: SUCCESS: BUILD AND INSTALL S2I [00h 00m 30s] ##########
```

https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/common/test_cases/s2i.yml#L7-L11

and yet on the ec2 instance the last commit in:
/data/src/github.com/openshift/source-to-image

is from feb 28th.

(the last commit in s2i master is from march 7th)

gabemontero:
On Fri, Mar 9, 2018 at 11:07 AM, Ben Parees <notifications@github.com>
wrote:

> I sshe'd into another ec2 instance, one running the wildfly push job which
> also syncs s2i and indeed the source is not updated. @stevekuznetsov
> <https://github.com/stevekuznetsov> this job:
> https://ci.openshift.redhat.com/jenkins/job/push_wildfly_
> images/9/consoleFull
>
> should have synced the s2i repo:
>
> ########## FINISHED STAGE: SUCCESS: BUILD AND INSTALL S2I [00h 00m 30s] ##########
>
> https://github.com/openshift/aos-cd-jobs/blob/master/sjb/
> config/common/test_cases/s2i.yml#L7-L11
>

OK whew at least we know the docker/s2i bug is not still out there

I noticed a fair amount of `oct sync remote ` invocations for various repos
in aos-cd-jobs ...
should we have one of those in the s2i.yml ?

gabemontero:
OK, to the specifics of the PR, I've updated per @bparees 's comments 

1) added configure-agent to slave-base 
2) removed the k8s plugin config change, seds associated with new image, to save for separate PR for after we publish some images

bparees:
@gabemontero since these new image names don't have "slave" in them, you're going to need to tweak this:
https://github.com/openshift/jenkins/blob/master/hack/build.sh#L37-L48

bparees:
(I think that's my last comment)

gabemontero:
thanks @bparees 

pushed update ... didn't squash yet

hack/build.sh
```diff
@@ -45,6 +45,10 @@ for dir in ${dirs}; do
     dirs=( "slave-base ${dirs[@]}")
     break
   fi
+  if [[ "$dir" =~ "agent" ]]; then
+    dirs=( "slave-base ${dirs[@]}")
+    break
+  fi
```
bparees: if you're building a "slave" image and an "agent" image this is going to end up with a very duplicated array.

just make the existing if do an || check.

gabemontero: just pushed

bparees:
lgtm

gabemontero:
commits squashed / merge label added

gabemontero:
failure was docker/s2i

/retest

bparees:
> failure was docker/s2i

@stevekuznetsov's changes should have addressed that.

PR #544, opened 2018-03-15
incur jenkins update center pain during startup vs. login via groovy …

gabemontero:
…init

@openshift/sig-developer-experience fyi ptal

this a a redo of an end of 3.9 release attempt we made to incur update center delay during startup vs. login

however we quickly met resistance in that image extenders were leveraging the default init.groovy.d/init.groovy

this employ an atypical name which should not conflict, and delves into manipulating the permissions to allow assemble scripts to mess with this

I stopped short of chmod 777 both the dir and the file, but may reconsider

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -84,6 +84,7 @@ RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
     unlink /usr/share/man/man1/unpack200.1.gz && \
     chown -R 1001:0 /opt/openshift && \
     /usr/local/bin/fix-permissions /opt/openshift && \
+    /usr/local/bin/fix-permissions /opt/openshift/configuration/init.groovy.d && \
     /usr/local/bin/fix-permissions /var/lib/jenkins && \
     /usr/local/bin/fix-permissions /var/log
 
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -101,6 +101,7 @@ RUN rm /opt/openshift/base-plugins.txt && \
     chown -R 1001:0 /opt/openshift && \
     # the prior chown doesn't traverse the /opt/openshift/plugins links .. this one will assist fix-permission/assemble for extension builds like master/slave
     chown 1001:0 /usr/lib/jenkins/*hpi && \ 
+    /usr/local/bin/fix-permissions /opt/openshift/configuration/init.groovy.d && \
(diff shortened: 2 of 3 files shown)


gabemontero:
fyi @openshift/sig-developer-experience 

since this image update, at least when running extended tests locally on my laptop (which could be CPU thing since I'm running other things in parallel on my laptop), I occasionally see the jenkins deployment fail because it is "not progressing"

i.e. see this portion of the test framework:  https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L833-L846

and this explanation of the "progressing" monitoring: https://github.com/openshift/origin/blob/master/pkg/apps/apis/apps/types.go#L462-L467

Sure enough, when looking at the jenkins pod log dump, it is still in the middle of our new groovy init script action around downloading all the info from the jenkins update center)

At first blush in looking at https://github.com/openshift/origin/blob/master/pkg/apps/apis/apps/types.go#L462-L467 this timeout seems separate from the liveness/readiness probes.

I'm going to continue to investigate/experiment, but short term, wanted to give a heads up for folks monitoring the overnight extended test runs or submitting PRs which result in running tests against pipeline.go.

bparees:
yeah we may have to bump the deployment timeout in the template as a result of this.  I assume the deployment is looking for the pod to become "ready" and of course that will take longer now.

gabemontero:
default appears to be 10 mintues ... see https://github.com/openshift/origin/blob/master/pkg/apps/apis/apps/types.go#L11-L14

gabemontero:
though I'm still digging ... it may be more nuanced than that

gabemontero:
Yeah everything I see confirms the comment at https://github.com/openshift/origin/blob/master/pkg/apps/apis/apps/types.go#L462 that the progressing == false (i.e. slow) is purely informational

I think this rather boils down to a test change ... at least for jenkins, we need to be a little more forgiving wrt the progressing flag and not fail as soon as we see that ... either checking less often, or allowing for a few consecutive scans of not progressing before we fail

gabemontero:
btw ... being able to run the openshift/origin extended tests out of this repo quite possibly would have caught this

gabemontero:
first occurrence with our regression runs on ci / aws:  https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_image_ecosystem/417/consoleFull#-153881154656cbb9a5e4b02b88ae8c2f77

hope to have a origin pr which addresses soon

PR #545, opened 2018-03-16
add yarn 1.5.1

arnaud-deprez:
Hi,

- [x] Add [Yarn](https://yarnpkg.com/en) for those like me who prefer it over npm
- [x] Add support for NODE_MIRROR_URL (similar to MAVEN_MIRROR_URL but for node repositories)

/assign @bparees @gabemontero

### DIFF
--- a/agent-nodejs-8/test/agent_nodejs_test.go
+++ b/agent-nodejs-8/test/agent_nodejs_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"os"
+	"strings"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
@@ -101,6 +102,32 @@ var _ = Describe("NodeJS agent testing", func() {
 			&container.Config{
 				Image:      imageName,
 				Entrypoint: []string{"/bin/bash", "-c", "node --version"},
+				Tty:        true,
+			},
+			nil)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = dockercli.ContainerStart(id)
+		Expect(err).NotTo(HaveOccurred())
+
+		code, err := dockercli.ContainerWait(id)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(code).To(Equal(0))
+	})
+
+	It("should contain a runnable yarn", func() {
+		// the default entrypoint for the image assumes if you supply more than
+		// one arg, you are trying to invoke the slave logic, so we have to
+		// override the entrypoint to run complicated commands for testing.
+		if strings.Contains(imageName, "rhel") {
+			Skip("n/a on RHEL image")
+		}
+
+		var err error
+		id, err = dockercli.ContainerCreate(
+			&container.Config{
+				Image:      imageName,
+				Entrypoint: []string{"/bin/bash", "-c", "yarn --version"},
 				Tty:        true,
 			},
 			nil)
--- a/agent-nodejs-8/Dockerfile
+++ b/agent-nodejs-8/Dockerfile
@@ -3,6 +3,7 @@ FROM openshift/jenkins-slave-base-centos7
 MAINTAINER Gabe Montero <gmontero@redhat.com>
 
 ENV NODEJS_VERSION=8 \
+    YARN_VERSION=1.5.1 \
     NPM_CONFIG_PREFIX=$HOME/.npm-global \
     PATH=$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:$PATH \
     BASH_ENV=/usr/local/bin/scl_enable \
@@ -17,6 +18,8 @@ RUN yum install -y centos-release-scl-rh && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
+    source /usr/local/bin/scl_enable && \ 
+    npm install -g yarn@${YARN_VERSION} && \
     yum clean all -y
 
 RUN chown -R 1001:0 $HOME && \


agent-nodejs-8/Dockerfile.rhel7
```diff
@@ -31,6 +32,8 @@ RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
+    source /usr/local/bin/scl_enable && \ 
+    npm install -g yarn@${YARN_VERSION} && \
```
bparees: we can't do this.  all content in the rhel image must come from rpms, we can't source rhel image content from the internet.

bparees:
> @bparees can this be back ported for openshift 3.8 ?

the new nodejs8 agent image isn't even a part of 3.9 and 3.8 isn't really a "real" release, so no.

also before this can be merged we need to confirm there is a rhel package available for yarn to install in our rhel version of this image.

arnaud-deprez:
Hi @bparees,

I cannot access to the https://access.redhat.com/downloads/content/package-browser right now, I can't find back my account and I need to ask a reset in my company.

On the other hand I found this: https://bugzilla.redhat.com/show_bug.cgi?id=1518262
They said in that topic that it's actually not recommended to install yarn via npm, so I tried to install it with their [recommended way](https://yarnpkg.com/lang/en/docs/install): 
```sh
curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | sudo tee /etc/yum.repos.d/yarn.repo
yum install yarn
```
But I have this error: 
```sh
Resolving Dependencies
--> Running transaction check
---> Package yarn.noarch 0:1.5.1-1 will be installed
--> Processing Dependency: nodejs for package: yarn-1.5.1-1.noarch
--> Finished Dependency Resolution
Error: Package: yarn-1.5.1-1.noarch (yarn)
           Requires: nodejs
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```
But node and npm are already available in the image.

Any idea how to solve that so it can be compatible with your policies ?

bparees:
I did a search, there does not appear to be a rhel packaged version of yarn currently available.

arnaud-deprez:
Ok so I propose this PR.
I removed yarn from rhel image and leave it in centos image.
Keep the NODE_MIRROR_URL feature for both.

agent-nodejs-8/README.md
```diff
@@ -0,0 +1,9 @@
+Node Slave Image
+====================
+
+This repository contains Dockerfiles for a Jenkins Agent Docker image intended for 
+use with [OpenShift v3](https://github.com/openshift/origin)
+
+Node Reregistry Support
```
bparees: s/Reregistry/Registry/

bparees:
I don't love diverging our centos and rhel images (among other things it means subscribing customers end up w/ a downgraded experience unless they choose to run the unsupported image), but we did already do it w/ the maven images (centos includes gradle, rhel does not, for similar reasons) so it's not w/o precedent.  I think we can do this.

@gabemontero any concerns?

arnaud-deprez:
Typo fixed, let me know about divergency.

arnaud-deprez:
Hi, 
I'm not sure to understand why the test is failing...
I cannot reproduce it on my machine neither because of: 
```
./agent_nodejs_test.go:58:8: cannot use "github.com/docker/engine-api/types/container".Config literal (type *"github.com/docker/engine-api/types/container".Config) as type *"github.com/openshift/jenkins/vendor/github.com/docker/engine-api/types/container".Config in argument to dockercli.ContainerCreate
./agent_nodejs_test.go:81:8: cannot use "github.com/docker/engine-api/types/container".Config literal (type *"github.com/docker/engine-api/types/container".Config) as type *"github.com/openshift/jenkins/vendor/github.com/docker/engine-api/types/container".Config in argument to dockercli.ContainerCreate
./agent_nodejs_test.go:104:8: cannot use "github.com/docker/engine-api/types/container".Config literal (type *"github.com/docker/engine-api/types/container".Config) as type *"github.com/openshift/jenkins/vendor/github.com/docker/engine-api/types/container".Config in argument to dockercli.ContainerCreate
./agent_nodejs_test.go:121:6: undefined: strings
./agent_nodejs_test.go:130:8: cannot use "github.com/docker/engine-api/types/container".Config literal (type *"github.com/docker/engine-api/types/container".Config) as type *"github.com/openshift/jenkins/vendor/github.com/docker/engine-api/types/container".Config in argument to dockercli.ContainerCreate
```
I'm afraid not familiar with golang to solve this right now.

agent-nodejs-8/test/agent_nodejs_test.go
```diff
@@ -113,4 +113,30 @@ var _ = Describe("NodeJS agent testing", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(code).To(Equal(0))
 	})
+
+	It("should contain a runnable yarn", func() {
+		// the default entrypoint for the image assumes if you supply more than
+		// one arg, you are trying to invoke the slave logic, so we have to
+		// override the entrypoint to run complicated commands for testing.
+		if strings.Contains(imageName, "rhel") {
```
bparees: you need to add an import for the "strings" package.

yarn-error.log
```diff
@@ -0,0 +1,32 @@
+Arguments:
```
bparees: I assume you didn't intend for this file to be added to your PR.

gabemontero:
On Sat, Mar 17, 2018 at 9:43 PM, Ben Parees <notifications@github.com>
wrote:

> I don't love diverging our centos and rhel images (among other things it
> means subscribing customers end up w/ a downgraded experience unless they
> choose to run the unsupported image), but we did already do it w/ the maven
> images (centos includes gradle, rhel does not, for similar reasons) so it's
> not w/o precedent. I think we can do this.
>
> @gabemontero <https://github.com/gabemontero> any concerns?
>
Yeah I think this is the reality of things.  The centos image for all
intents and purposes is our early release, rapid delivery vehicle.
As long as we anticipate the yarn rpm eventually arriving, I'm ok.

In the maven/gradle case for example, my understanding is that it is just a
matter of time.  i.e. we got some sort of assurance that
was the case (unless I'm mis-remembering).

That said, is there a way we can confirm that the yarn is in fact coming?

bparees:
> That said, is there a way we can confirm that the yarn is in fact coming?

probably need to ask our friends in SCL, at least as a starting point.

agent-nodejs-8/test/agent_nodejs_test.go
arnaud-deprez: thanks

bparees:
@omron93 @hhorak who would know if there are plans to produce a yarn rpm?

hhorak:
@bparees, I'll ask @notroj or @kasicka about the yarn, but I'm not seeing it in fedora either at this point.

kasicka:
There is currently package review request open for yarn in fedora.

hhorak:
> @bparees, I'll ask @notroj or @kasicka about the yarn, but I'm not seeing it in fedora either at this point.

Just checked with @notroj, it looks like there is no official request that we track, so please, fill it in with reasoning into the Bugzilla.

bparees:
> Maybe (not checked) it can be made available similarly to pipenv, i.e. when yarn needed, it could be installed first via npm, and then the yarn is used? @kasicka, do you think something like this would work?

the problem is these are jenkins slave images, so they get spun up repeatedly.  Having to install yarn every time a slave starts is going to be a bad (and probably fragile) experience.

>  it looks like there is no official request that we track, so please, fill it in with reasoning into the Bugzilla.

what bz product/component?

arnaud-deprez:
Any news ?

omron93:
>what bz product/component?

@bparees I guess:

Product: Red Hat Software Collections
Component: rh-nodejs8
Version: rh-nodejs8

gabemontero:
I opened https://bugzilla.redhat.com/show_bug.cgi?id=1568080 

Based on how that discussion goes, I may solicit additional info or reqs, etc. here to put in that bugzilla (since interested folks here may not have access).

arnaud-deprez:
Thanks, keep me inform.
Normally, I should get a new account soon, but it's still on process queue somewhere in my company...

AlginMaduro:
Maybe its best to split the Yarn and the Node mirror into 2 different PR's. I am currently doing the same(node mirror) because this is not merged yet. If you split it into 2 PR's it becomes smaller and easier to merge.

arnaud-deprez:
Yep, I thought about that too. I can take care or toi can if hou have more time ;-)

arnaud-deprez:
@AlginMaduro please have a look at #611

akram:
we will close this one as we are deprecating node8 image
The PR get very old now, we are sorry for that
@arnaud-deprez if you feel that you still need this, I suggest you to open a new PR with the nodejs10 image that we are releasing also

PR #551, opened 2018-03-19
allow for the disabling of memory intensive admin monitors via env va…

gabemontero:
…r and template param

while these update center admin monitors are on by default, these changes will allow us to explicitly turn them off minimally in our extended tests, and possibly in online starter if we chose to do so

Details in https://github.com/openshift/origin/pull/19008

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -28,9 +28,22 @@ function create_jenkins_config_xml() {
     if [ -f "${image_config_path}.tpl" ]; then
       export KUBERNETES_CONFIG=$(generate_kubernetes_config)
       echo "Generating kubernetes-plugin configuration (${image_config_path}.tpl) ..."
+      export ADMIN_MON_CONFIG=$(generate_administrative_monitors_config)
+      echo "Generating administrative monitor configuration (${image_config_path}.tpl) ..."
       envsubst < "${image_config_path}.tpl" > "${image_config_path}"
     fi
   fi
+}
+
+function generate_administrative_monitors_config() {
+  if [[ "${DISABLE_ADMINISTRATIVE_MONITORS}" == "true" ]]; then
+      echo '
+        <string>hudson.model.UpdateCenter$CoreUpdateMonitor</string>
+        <string>jenkins.security.UpdateSiteWarningsMonitor</string>
+      '
+      return
+  fi
+  echo ""
 }
 
 function create_jenkins_credentials_xml() {
@@ -377,6 +390,20 @@ else
 
     install_plugins
   fi  
+fi
+
+if [[ "${DISABLE_ADMINISTRATIVE_MONITORS}" == "true" ]]; then
+    echo "Disabling administrative monitors that contact the update center"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" ]; then
+	echo "Turning off startup groovy init update center administrative monitor initialization"
+	mv "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" "${JENKINS_HOME}/init.groovy.d/update-center-init.disabled"
+    fi
+else
+    echo "Administrative monitors that contact the update center will remain active"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.disabled" ]; then
+	echo "Turning on startup groovy init update center administrative monitor initialization"
+	mv "${JENKINS_HOME}/init.groovy.d/update-center-init.disabled" "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy"
+    fi
 fi
 
 if [ -e ${JENKINS_HOME}/password ]; then
--- a/openshift/templates/jenkins-ephemeral.json
+++ b/openshift/templates/jenkins-ephemeral.json
@@ -118,6 +118,10 @@
                     "value": "true"
                   },
                   {
+                    "name": "DISABLE_ADMINISTRATIVE_MONITORS",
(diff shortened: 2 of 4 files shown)


2/contrib/s2i/run
```diff
@@ -28,11 +28,25 @@ function create_jenkins_config_xml() {
...
       envsubst < "${image_config_path}.tpl" > "${image_config_path}"
     fi
   fi
 }
 
+function generate_administrative_monitors_config() {
+  DOLLAR='$'
+  if [[ "${DISABLE_ADMINISTRATIVE_MONITORS}" == "true" ]]; then
+      echo "
+        <string>hudson.model.UpdateCenter${DOLLAR}CoreUpdateMonitor</string>
```
bparees: is this ${DOLLAR} right?  this value represents something that's being substituted into the config.xml.  I wouldn't expect it to need to be escaped.

2/contrib/s2i/run
```diff
@@ -379,6 +393,20 @@ else
   fi  
 fi
 
+if [[ "${DISABLE_ADMINISTRATIVE_MONITORS}" == "true" ]]; then
+    echo "Disabling administrative monitors that contact the update center"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" ]; then
+	echo "Turning off startup groovy init update center administrative monitor initialization"
+	mv "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" "${JENKINS_HOME}/init.groovy.d/update-center-init.save"
```
bparees: s/save/disabled/

2/contrib/s2i/run
```diff
@@ -379,6 +393,20 @@ else
...
+    fi
+else
+    echo "Administrative monitors that contact the update center will remain active"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.save" ]; then
+	echo "Turning on startup groovy init update center administrative monitor initialization"
+	mv "${JENKINS_HOME}/init.groovy.d/update-center-init.save" "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy"
```
bparees: s/save/disabled/

2/contrib/s2i/run
```diff
@@ -379,6 +393,20 @@ else
...
 
+if [[ "${DISABLE_ADMINISTRATIVE_MONITORS}" == "true" ]]; then
+    echo "Disabling administrative monitors that contact the update center"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" ]; then
+	echo "Turning off startup groovy init update center administrative monitor initialization"
+	mv "${JENKINS_HOME}/init.groovy.d/update-center-init.groovy" "${JENKINS_HOME}/init.groovy.d/update-center-init.save"
+    fi
+else
+    echo "Administrative monitors that contact the update center will remain active"
+    if [ -e "${JENKINS_HOME}/init.groovy.d/update-center-init.save" ]; then
```
bparees: s/save/disabled/

2/contrib/s2i/run
bparees: or rather, you can use echo with single quotes to avoid variable resolution here.

gabemontero: Yes this works.

And so does `hudson.model.UpdateCenter\$CoreUpdateMonitor`, without setting of `DOLLAR='$'` on the same line as the `envsubst` call

I employed the current approach based on the discussion that occurred in the original PR, but am perfectly find just switching to `hudson.model.UpdateCenter\$CoreUpdateMonitor`

What does not work:  A string of `hudson.model.UpdateCenter$CoreUpdateMonitor`

Followed later on by `DOLLAR='$' envsubst....` ... the `$` is lost on the above echo even and then there is nothing for `DOLLAR='$' envsubst ...`

bparees: > the $ is lost on the above echo 

single quotes would solve that.

```
#!/bin/sh
  
function foo() {
  echo 'a$b'
}
export bar=$(foo)
envsubst < file.json
```

```
$ cat file.json 
a
$bar
foo
```

```
$ ./t.sh 
a
a$b
foo
```

gabemontero:
updates pushed @bparees 

i'll squash if you are good

openshift/templates/jenkins-ephemeral.json
```diff
@@ -277,6 +281,12 @@
       "description": "The OpenShift Namespace where the Jenkins ImageStream resides.",
       "value": "openshift"
     },
+    {
+      "name": "DISABLE_ADMINISTRATIVE_MONITORS",
+      "displayName": "Disable memory intensive administrative monitors",
+      "description": "Whether to perform memory intensive, possibly slow, synchronization with the Jenkins Update Center on starte.  If true, the Jenkins core update monitor and site warnings monitor are disabled.",
```
bparees: s/starte/start/

openshift/templates/jenkins-persistent.json
```diff
@@ -301,6 +305,12 @@
       "description": "The OpenShift Namespace where the Jenkins ImageStream resides.",
       "value": "openshift"
     },
+    {
+      "name": "DISABLE_ADMINISTRATIVE_MONITORS",
+      "displayName": "Disable memory intensive administrative monitors",
+      "description": "Whether to perform memory intensive, possibly slow, synchronization with the Jenkins Update Center on starte.  If true, the Jenkins core update monitor and site warnings monitor are disabled.",
```
bparees: starte typo

gabemontero:
cool thanks ... fixed typos and squashed

PR #552, opened 2018-03-19
move update center init groovy to background thread

gabemontero:

moving these downloads to a separate threads will accelerate jenkins startup to some degree, if sufficient cpu/cores are available, though my testing shows the pod may still not be deemed ready until the download completes, based on whether the readiness probes can squeeze out enough cpu even after the jenkins is up and ready message appears

### DIFF
--- a/2/contrib/openshift/configuration/init.groovy.d/update-center-init.groovy
+++ b/2/contrib/openshift/configuration/init.groovy.d/update-center-init.groovy
@@ -1,3 +1,27 @@
 import jenkins.model.Jenkins
-Jenkins.getInstance().getPluginManager().doCheckUpdatesServer()
-Jenkins.getInstance().getUpdateCenter().getCoreSource().getData()
+import hudson.security.ACL;
+import jenkins.security.NotReallyRoleSensitiveCallable;
+
+ACL.impersonate(ACL.SYSTEM, new NotReallyRoleSensitiveCallable<Void, Exception>() {
+    public Void call() throws Exception {
+        Thread thread = new Thread(){
+            public void run(){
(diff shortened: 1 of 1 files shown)


2/contrib/openshift/configuration/init.groovy.d/update-center-init.groovy
```diff
@@ -1,3 +1,27 @@
...
+                                e.printStackTrace();
+                            }
+                            Jenkins.getInstance().getUpdateCenter().getCoreSource().getData();
+                        return null;
+                    }
+                });
+            }
+        };
+        
+        thread.start();
```
bparees: will this still block until the updates are complete?

gabemontero: No my testing shows it is concurrent / non-blocking vs. just calling those 2 methods, which is good since that was my recollection of how java threads worked ;-)

But per my note in the description, the cpu cost since the activity is still occurring and generating the need for GC, and possible impact to the readiness probe ... but offloading to another thread appears to help somewhat

bparees: i'm just wondering since it's not blocking, if that means now a user can get hung at the login screen again if this is not done yet?

gabemontero: I have not seen such timing, where the login prompt occurs super early, but this background processing is still occurring.  The main startup thread is still doing plenty of other stuff.

But granted have only been creating pods in some form or fashion with it on my laptop for a couple of days.

I have observed the startup speed up *some*, whether by manually launching a pod, or starting the extended tests.

gabemontero:
`Stage FORWARD GCS CREDENTIALS TO REMOTE HOST [00h 00m 00s] failed.`

/retest

PR #555, opened 2018-03-21
add verification for KUBE_CA contents

gabemontero:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1548619

### DIFF
--- a/2/contrib/jenkins/kube-slave-common.sh
+++ b/2/contrib/jenkins/kube-slave-common.sh
@@ -74,6 +74,11 @@ function generate_kubernetes_config() {
     [ -z "$oc_cmd" ] && return
     [ ! has_service_account ] && return
     local crt_contents=$(cat "${KUBE_CA}")
+    # do not dump contents of cert here, but allow any error messages to be seen
+    openssl x509 -in $KUBE_CA > /dev/null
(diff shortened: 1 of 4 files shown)


2/contrib/jenkins/kube-slave-common.sh
```diff
@@ -74,6 +74,11 @@ function generate_kubernetes_config() {
     [ -z "$oc_cmd" ] && return
     [ ! has_service_account ] && return
     local crt_contents=$(cat "${KUBE_CA}")
+    # do not dump contents of cert here, but allow any error messages to be seen
+    openssl x509 -in $KUBE_CA > /dev/null
+    if [ $? -eq 1 ] ; then
+      crt_contents=""
```
bparees: i take it setting an empty cert in the xml is not a problem?  if so, lgtm.

gabemontero: yep good question @bparees ....,I have verified by setting the cert field to empty and then restarted Jenkins.  It came up fine.  No exceptions.

gabemontero:
oops there was some feedback in the bugzilla that I was about to push .... I now realize @bparees was not copied on the bugzilla 

I'll create a follow up PR .... this change as-is it benign

PR #558, opened 2018-03-22
Jenkins 1: Update `oc` binary to v3.7.2

mhutter:


bparees:
the jenkins:1 image is deprecated, we will not be updating it w/ newer oc clients.

PR #561, opened 2018-03-23
remove all v1 related content

bparees:


bparees:
/assign @gabemontero 

our Makefile and CI jobs were already ignoring v1 (which is too bad, i was hoping we could speed them up by removing the v1 build/test step)

gabemontero:
do we want to delete the openshift/imagestream json file references?

gabemontero:
There is also this comment in the Makefile:  `:# the image name (jenkins-1-centos7 in this case).  For the slave`

bparees:
> do we want to delete the openshift/imagestream json file references?

no, we can never remove those.  People can still have resources that reference that imagestream tag.

> There is also this comment in the Makefile: :# the image name (jenkins-1-centos7 in this case). For the slave

fixed

Issue #528, opened 2018-03-06
S2I build fails if init.groovy.d/ present


jlebon:
It looks like #440 might have caused a regression on client builds that want to bake in `init.groovy.d/` scripts:

```
---> Installing new Jenkins configuration ...
mv: cannot move ‘/tmp/tmp.isnFJKrzxzjenkins/configuration/init.groovy.d’ to ‘/opt/openshift/configuration/init.groovy.d’: Directory not empty
error: build error: non-zero (13) exit code from openshift/jenkins-2-centos7@sha256:8e5a8cfb470ac744b903cc2fc6607a78d3d98b7ee6af5e718a39f7730c6d9663
```

bparees:
@gabemontero please revert this out of the 3.9 branch while we sort this out.  it's not needed for 3.9 anyway (the template changes are sufficient).

gabemontero:
sure ... while users laying claim to the init.groovy file is speculative IMO, either in the top level dir or init.groovy.do subdir, the time needed to sort through that debate conflicts with our 3.9 timetable.

jlebon:
I posted an easy fix for this in #529. Though I'm fine with reverting as well. Ideally, we would have "merge" semantics there.

gabemontero:
On Tue, Mar 6, 2018 at 3:05 PM, Jonathan Lebon <notifications@github.com>
wrote:

> I posted an easy fix for this in #529
> <https://github.com/openshift/jenkins/pull/529>. Though I'm fine with
> reverting as well. Ideally, we would have "merge" semantics there.
>
Yep saw that - thanks.  That said, in theory there might be users who
assumed they could lay claim to the other file as well.
We'll want to give the issue proper consideration.

Thanks for the quick heads up by the way ;-)

jlebon:
> Ideally, we would have "merge" semantics there.

I just updated #529 to do this.

gabemontero:
On Tue, Mar 6, 2018 at 3:10 PM, Jonathan Lebon <notifications@github.com>
wrote:

> Ideally, we would have "merge" semantics there.
>
> I just updated #529 <https://github.com/openshift/jenkins/pull/529> to do
> this.
>
OK cool.  We'll certainly cross reference with that when we revisit this.
Thanks again.

gabemontero:
This has been backed out with the recent centos image updates and is part of 3.9 rhel

[...]

PR #565, opened 2018-04-04
switch from mv to rsync in assemble script to better account for exis…

gabemontero:
…ting directories

(Re)Fixes https://github.com/openshift/jenkins/issues/528

In 3.9, we ended up removing the init.groovy.d stuff for GA.

Added it back in 3.10, with a name change for the script, but still did not account for mv not dealing with existing dirs.

@jlebon fyi


jlebon:
Awesome, LGTM as well!

PR #576, opened 2018-04-18
Adding checkmarx plugin with version 8.60.0

manoharendla:


bparees:
@manoharendla what is the motivation for adding this plugin?  

In general we only want to include plugins in the image that specifically support the openshift pipeline feature.  We make it very easy for users to add their own plugins to this image[1][2] if the base set doesn't cover their use case, but we don't want to increase the maintenance burden or image size by including arbitrary plugins that aren't relevant to a significant percentage of users.

[1]https://docs.openshift.org/latest/using_images/other_images/jenkins.html#jenkins-as-s2i-builder
[2]https://docs.openshift.org/latest/using_images/other_images/jenkins.html#jenkins-environment-variables  (see the INSTALL_PLUGINS env var)

manoharendla:
Will be making use of S2i to have an image with custom plugins

PR #584, opened 2018-04-25
temp remove of rpm -V until origin int rhel 7.5 repos picks up fedora…

gabemontero:
… fix for java rpms and rpm -V

per request from @jupierce


bparees:
(though shouldn't 3.8 be being built using rhel 7.4 repos?)

jupierce:
@bparees We build images on the latest GA RHEL regardless of release.

PR #585, opened 2018-04-25
head off monitor plugin interference (via filter induced session crea…

gabemontero:
…tion) with login plugin bearer token support

@scoheb discovered this interference to our login plugin's bearer token support recently while building his team's extension to our image

With my ever increasing disdain for more tuning knobs for this image, we are always setting this.  If users want the image to play nice with this monitor plugin and allow the feature we have turned off, they can extend out image.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -453,7 +453,9 @@ JENKINS_SERVICE_NAME=${JENKINS_SERVICE_NAME:-JENKINS}
 JENKINS_SERVICE_NAME=`echo ${JENKINS_SERVICE_NAME} | tr '[a-z]' '[A-Z]' | tr '-' '_'`
 
 if [[ -z "${JENKINS_JAVA_OPTIONS}" ]]; then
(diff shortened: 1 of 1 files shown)


2/contrib/s2i/run
```diff
@@ -466,6 +466,7 @@ if [[ $# -lt 1 ]] || [[ "$1" == "--"* ]]; then
   set -x
   exec java $JENKINS_JAVA_OPTIONS -Duser.home=${HOME} \
             -Djavamelody.application-name=${JENKINS_SERVICE_NAME} \
+            -Djavamelody.displayed-counters=log,error \ # a discover was made upstream that if the monitor plugin is installed, it creates httpsession's via its filter, which impact the login plugin bearer token support, this setting turns that off
```
bparees: how about setting it in JAVA_OPTIONS so users can at least override the entire JAVA_OPTIONS to change the behavior? above

gabemontero:
@bparees comment addressed, fixed bash parsing error

PR #588, opened 2018-04-30
do not enable i686 unless we are x86_64 based

gabemontero:
for bug https://bugzilla.redhat.com/show_bug.cgi?id=1572710

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -35,9 +35,10 @@ RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/j
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
     curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.107.3-1.1" && \
-    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
-    rpm -V $INSTALL_PKGS && \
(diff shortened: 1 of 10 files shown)


gabemontero:
if logic refactored @bparees

gabemontero:
agreed ... the main reason I did not cite the bugzilla here, etc. was that I ultimately wanted to hold pending some of those updates from planning, etc.

if we end up doing this I'll link the items then, etc.

bparees:
@gabemontero at this point i'm somewhat leaning towards merging this (After the rebase conflict is addressed) and getting it off our plate...thoughts?

gabemontero:
yeah let's do it ... i'll rebase in a couple

gabemontero:
@bparees ptal

2/Dockerfile
```diff
@@ -35,8 +35,9 @@ RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/j
     rpm --import https://pkg.jenkins.io/redhat-stable/jenkins-ci.org.key && \
     yum install -y centos-release-scl-rh && \
     curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.107.3-1.1" && \
-    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
+    x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \
+    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk jenkins-2.107.3-1.1" && \
+    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS $x86_EXTRA_RPMS && \
     rpm -V $INSTALL_PKGS && \
```
bparees: need to verify $x86_EXTRA_RPMS too.

bparees: alternatively include $x86_EXTRA_RPMS when you set INSTALL_PKGS above and then you don't need to pass it explicitly to `yum install` or `rpm -V`

gabemontero: update pushed

PR #589, opened 2018-04-30
Backport Jenkins latest stable jenkins-2.107.2-1.1

jtescher:


gabemontero:
@jtescher - to date we have not been backporting jenkins lts updates to existing releases because resource limitations wrt regression testing

Can I assume you have seen that we are at 2.107.2-1.1 in our 3.10 version of this image?

jtescher:
Ah ok, so users of 3.9 should just use the 3.10 tag 👍

PR #599, opened 2018-05-10
Fix the value of -Djavamelody.displayed-counters

psiroky:
The single quotes around the value were causing a plugin init fail:

  Caused by: java.lang.IllegalArgumentException: Unknown counter: 'log
      at net.bull.javamelody.FilterContext.setDisplayedCounters(FilterContext.java:237)
      at net.bull.javamelody.FilterContext.initCounters(FilterContext.java:196)


gabemontero:
thanks @psiroky !

PR #600, opened 2018-05-12
don't configure javac in base slave image

bparees:
the run-jnlp-client script in slave base tries to enable the javac alternative, but there's no JDK installed (only a JRE) so it fails:

```
$ docker run -it brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/openshift3/jenkins-agent-nodejs-8-rhel7 /bin/sh
alternatives version 1.7.4 - Copyright (C) 2001 Red Hat, Inc.
This may be freely redistributed under the terms of the GNU Public License.

usage: alternatives --install <link> <name> <path> <priority>
                    [--initscript <service>]
                    [--family <family>]
                    [--slave <link> <name> <path>]*
       alternatives --remove <name> <path>
       alternatives --auto <name>
       alternatives --config <name>
       alternatives --display <name>
       alternatives --set <name> <path>
       alternatives --list

common options: --verbose --test --help --usage --version --keep-missing
                --altdir <directory> --admindir <directory>
OPENSHIFT_JENKINS_JVM_ARCH='', CONTAINER_MEMORY_IN_MB='8796093022207', using /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.171-7.b10.el7.x86_64/jre/bin/java and 

```

### DIFF
--- a/agent-maven-3.5/contrib/bin/configure-agent
+++ b/agent-maven-3.5/contrib/bin/configure-agent
@@ -8,3 +8,20 @@ if [ -n "$MAVEN_MIRROR_URL" ]; then
   </mirror>"
   sed -i "s|<!-- ### configured mirrors ### -->|$xml|" $HOME/.m2/settings.xml
 fi
+
+CONTAINER_MEMORY_IN_BYTES=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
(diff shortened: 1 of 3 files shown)


bparees:
(the failure is benign but it looks bad)

PR #602, opened 2018-05-15
Fixes WAR bundled plugins

goober:
The regex for finding war bundled plugins is wrong, which having the effect that it never finds any bundled plugins when install-plugins.sh is executed.

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -312,7 +312,7 @@ function bundledPlugins() {
     if [ -f $JENKINS_WAR ]
     then
         TEMP_PLUGIN_DIR=/tmp/plugintemp.$$
-        for i in $(jar tf $JENKINS_WAR | egrep '[^detached-]plugins.*\..pi' | sort)
+        for i in $(jar tf $JENKINS_WAR | egrep '[^detached-plugins].*\..pi' | sort)
         do
             rm -fr $TEMP_PLUGIN_DIR
             mkdir -p $TEMP_PLUGIN_DIR


gabemontero:
Interesting, what we currently have still matches what exists at https://github.com/jenkinsci/docker/blob/master/install-plugins.sh#L137, where we borrowed this logic from.

Perhaps they have a bug as well.

Let's look at the messages around `Analyzing war...` and `WAR bundled plugins:` from the test run, and compare with a run without this commit.  I'll kick off one of those locally.

thanks @goober

gabemontero:
Yeah the plugin loads failed in the image testing.

This change in its current form @goober is not something we can accept.  If you can reconcile the failures in the tests at https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_jenkins/602/test_pull_request_jenkins_images/298/ with the scenario you are trying to address, or if you can describe with some specific examples which bundled plugins were failing, so that maybe we can offer another suggestions, we can revisit.

thanks

bparees:
doesn't match:
```
~$ echo detached-plugins-foo.jpi | egrep '[^detached-]plugins.*\..pi' ; echo $?
1
```

does match:
```
~$ echo detached-plugins-foo.jpi | egrep '[^detached-plugins].*\..pi' ; echo $?
detached-plugins-foo.jpi
0
```

That said, i'm actually unclear on *why* that is the case(why the first one does not match) and i'm also unclear on the desired behavior.

I don't actually think either expression should be using square brackets (character class), i think it should be using grouping ()s, such as:
`echo detached-plugins-foo.jpi | egrep '^(detached-)plugins.*\..pi' ; echo $?`

but that's basically the same as:
`echo detached-plugins-foo.jpi | egrep '^detached-plugins.*\..pi' ; echo $?`


Assuming the goal is to match plugins that start with "detached".  If the goal is to *exclude* plugins that start with `detach-` then another approach is needed.

gabemontero:
ah, ok, that cross reference https://github.com/jenkinsci/docker/issues/319 helped at least some @goober 

I cannot speak with certainty to the "why does it even exist" query you have.  I *SUSPECT* this is some form of legacy code, and perhaps
- the name changed since some time during v1
- should be removed given what I read in https://wiki.jenkins.io/display/JENKINS/Bundling+plugins+with+Jenkins

That said, for now I don't envision us diverging from https://github.com/jenkinsci/docker/blob/master/install-plugins.sh, at least not for an edge case like bundled plugins.  We just don't have the cycles to fully monitor these sort of jenkins related snowflakes and manage/understand them.

goober:
Thanks for the quick response and discussion. I'll close this one and bring up the discussion in the official jenkins docker repo.

PR #603, opened 2018-05-15
Rename directories for more consistency

Gl4di4torRr:
We found the directory structure to be confusing. It took a little while for us to determine there where different versions of node. I have not changed the underlying naming scheme. For example, `jenkins-slave-maven-rhel7` is still the name of the image.


bparees:
The directory names are aligned to the image names they produce, I don't think we should change that, it's the same pattern we follow in our other repos.

Perhaps the readme can be enhanced to make it clearer what the directories represent.

Gl4di4torRr:
@bparees I thought that might be the case. I will look into updating the readme. Thanks!

gabemontero:
OK closing this PR out based on the prior discussion.

PR #606, opened 2018-05-16
Add Nodejs 6 agent

Gl4di4torRr:
Opening this PR to make sure I complete all the needed tasks earlier rather than later. 

What items would y'all like to see for this to be accepted?


Gl4di4torRr:
We did have a reason for the v6 image. However, since I now have a better understanding of the goals for this repository, I think it will be best to not include it here. 

From my companies perspective, it is easier to use images if they come straight from the RedHat container registry.

Gl4di4torRr:
@gabemontero thanks! I am glad to try and help out anywhere I can :)

PR #610, opened 2018-05-17
Add save-artifacts s2i script for incremental builds

psiroky:
I was trying to keep the changes to a minimum to avoid a big divergence from the upstream `install-plugins.sh`. However, that means the change is not as clean as it could be. Let me know what you think - I could refactor the `download` and `doDownload` functions a bit more to avoid duplication at the cost of diverging from the upstream.

I've tested this with our image, which installs about 80 plugins, and the (incremental) build time got from 200s to about 35s.

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -150,6 +150,8 @@ FAILED="$REF_DIR/failed-plugins.txt"
 
 JENKINS_WAR=/usr/lib/jenkins/jenkins.war
 
+INCREMENTAL_BUILD_ARTIFACTS_DIR="/tmp/artifacts"
+
 function getLockFile() {
     echo -n "$REF_DIR/${1}.lock"
 }
@@ -199,6 +201,17 @@ function doDownload() {
         echo "Using provided plugin: $plugin"
         return 0
     fi
+
+    # Check if the plugin is cached and in correct version; if so, use it instead of downloading
+    # Some plugins do not follow the naming conventions and include the "-plugin" suffix; both cases need to be covered
+    for pluginFilename in "$plugin.jpi" "$plugin-plugin.jpi"; do
+        local cachedPlugin="$INCREMENTAL_BUILD_ARTIFACTS_DIR/plugins/$pluginFilename"
+        if test -f "$cachedPlugin" && [[ $(get_plugin_version "$cachedPlugin") == "$version" ]]; then
+            echo "Copying plugin from a cache created by s2i: $cachedPlugin"
+            cp "$cachedPlugin" "$jpi"
+            return 0
+        fi
+    done
 
     if [[ "$version" == "latest" && -n "$JENKINS_UC_LATEST" ]]; then
         # If version-specific Update Center is available, which is the case for LTS versions,
--- a/2/contrib/s2i/assemble
+++ b/2/contrib/s2i/assemble
@@ -3,6 +3,8 @@
 JENKINS_DIR=/opt/openshift
 JENKINS_INSTALL_DIR=$(mktemp -d --suffix=jenkins)
 
+INCREMENTAL_BUILD_ARTIFACTS_DIR=/tmp/artifacts
+
 shopt -s dotglob
 
 echo "---> Copying repository files ..."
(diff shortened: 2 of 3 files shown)


2/contrib/jenkins/install-plugins.sh
```diff
@@ -200,6 +202,17 @@ function doDownload() {
         return 0
     fi
 
+    # Check if the plugin is cached and in correct version; if so, use it instead of downloading
+    # Some plugins do not follow the naming conventions and include the "-plugin" suffix; both cases need to be covered
+    for pluginFilename in "$plugin.jpi" "$plugin-plugin.jpi"; do
+        local cachedPlugin="$INCREMENTAL_BUILD_ARTIFACTS_DIR/plugins/$pluginFilename"
+        if test -f "$cachedPlugin" && unzip -p "$cachedPlugin" META-INF/MANIFEST.MF | tr -d '\r' | grep "^Plugin-Version: ${version}$" > /dev/null; then
```
gabemontero: Did you consider using `get_plugin_version()` to get the cached version, and then just comparing that output to `${version}` ?

Or was that related to the further change in install_plugins.sh that you were referring to?  

I'd rather make that work, vs. duplicating/mimicking  the `tr` and `grep` logic

psiroky: I basically just copied the same check which is already there few lines above. `get_plugin_version` is definitely nicer.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -200,6 +202,17 @@ function doDownload() {
         return 0
     fi
 
+    # Check if the plugin is cached and in correct version; if so, use it instead of downloading
+    # Some plugins do not follow the naming conventions and include the "-plugin" suffix; both cases need to be covered
+    for pluginFilename in "$plugin.jpi" "$plugin-plugin.jpi"; do
+        local cachedPlugin="$INCREMENTAL_BUILD_ARTIFACTS_DIR/plugins/$pluginFilename"
+        if test -f "$cachedPlugin" && [[ $(get_plugin_version "$cachedPlugin") == "$version" ]]; then
+            echo "Copying plugin from a cache created by s2i: $cachedPlugin"
+            cp "$cachedPlugin" "$jpi"
```
bparees: i think this should be a move rather than a cp, so you don't end up w/ two copies of the plugin in the output image.

bparees: even better, erase the entire /tmp/artifacts dir when done installing plugins, that way even plugins that didn't get reused are removed.

psiroky: Great point! For whatever reason I though s2i would clean that up automatically, but that's not the case. Fixing in now.

psiroky: I've updated the `assemble` script to delete the `tmp/artifacts/plugins` dir. It looks like that's the only place that needs this change - is that correct?

2/contrib/s2i/assemble
```diff
@@ -17,6 +19,8 @@ if [ -f ${JENKINS_INSTALL_DIR}/plugins.txt ]; then
     echo "Failed to install plugins."
     exit 1
   fi
+  echo "Deleting $INCREMENTAL_BUILD_ARTIFACTS_DIR/plugins"
+  rm -rf "$INCREMENTAL_BUILD_ARTIFACTS_DIR/plugins"
```
bparees: I think you don't want this inside the `if plugins.txt` block, it should just always be run.  Someone could in theory do an incremental s2i build with a previous image that has installed plugins, but without supplying a `plugins.txt` to the new build iteration, at which point you still want to remove the injected /tmp/artifact content from the newly produced image, even though you didn't invoke `install-plugins.sh`

psiroky: Fixed, thanks!

bparees:
@gabemontero lgtm but you might want to hold for post 3.10.

gabemontero:
yep good thinking @bparees 

Thinking about when to cut the 3.10 branch ... could be soon ... I'll bring it up in scrum today.

PR #611, opened 2018-05-22
add NPM_MIRROR_URL support to use a node mirror registry

arnaud-deprez:
As mentioned https://github.com/openshift/jenkins/pull/545#issuecomment-389445502, I split it in favour of 2 distinct PR.

This enable a similar customization as maven agent but for node registry instead of maven repository.

### DIFF
--- a/agent-nodejs-8/README.md
+++ b/agent-nodejs-8/README.md
@@ -1,0 +1,9 @@
+Node Slave Image
+====================
+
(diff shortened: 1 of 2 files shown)


bparees:
lgtm.  probably want to do it after we cut the 3.10 branch just in case there's some unforeseen consequence.

arnaud-deprez:
NP @bparees, when you're ready ;-)

Issue #401, opened 2017-10-17
explore using --XX:ErrorFile for core file location


gabemontero:
noticed in man page with open jdk 

consider if the jenkins jvm core dumps ... we want to files to go to the PV maybe so they are salvageable after Pod restarts 

@bparees @csrwng @stevekuznetsov fyi

bparees:
i'm thinking the best option might be an env var to configure the destination.  (possibly don't even set the flag unless the user provided an env var value)

Gl4di4torRr:
I would think an env var would be the correct approach too.

From CloudBees: 

```-XX:ErrorFile= You will need to specify the directory path for the hs_err_pidXXXX.log to be written to.
An example for Jenkins running on Linux is -XX:ErrorFile=locationOfYourChoice/hs_err_%p.log .
An example for Jenkins running on Windows is -XX:ErrorFile=“locationOfYourChoice\hs_err_%p.log” .```

Is this something you would still like to add? Maybe I can take try to implement this.

gabemontero:
We will accept PRs for this @Gl4di4torRr

bparees:
fixed in https://github.com/openshift/jenkins/pull/612

PR #612, opened 2018-05-22
add -XX:ErrorFile as option

Gl4di4torRr:
From the comments in #401 I thought we might just want to make this a switch the users can set in the environment variables. I have set the location to the PV. @bparees @gabemontero let me know what you think of this implementation. I will do some more testing if this is a good idea. Thanks!


2/contrib/s2i/run
```diff
@@ -230,12 +230,19 @@ if [[ "${CONTAINER_MEMORY_IN_BYTES}" -lt $((2**40)) ]]; then
...
+    FATAL_ERROR_OPTION="-XX:ErrorFile=/var/lib/jenkins/logs"
+fi
+
 if [[ -z "$JAVA_GC_OPTS" ]]; then
   # See https://developers.redhat.com/blog/2014/07/22/dude-wheres-my-paas-memory-tuning-javas-footprint-in-openshift-part-2/ .
   # The values are aggressively set with the intention of relaxing GC CPU time
   # restrictions to enable it to free as much as possible, as well as
   # encouraging the GC to free unused heap memory back to the OS.
-  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90"
+  JAVA_GC_OPTS="-XX:+UseParallelGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 $FATAL_ERROR_OPTION"
```
bparees: this seems like it belongs in JENKINS_JAVA_OPTIONS, not JAVA_GC_OPTIONS.

2/contrib/s2i/run
```diff
@@ -230,12 +230,19 @@ if [[ "${CONTAINER_MEMORY_IN_BYTES}" -lt $((2**40)) ]]; then
   fi
 fi
 
+FATAL_ERROR_OPTION=""
+
+if [[ "${FATAL_ERROR_LOG}" == "true" ]]; then
+    echo "Fatal error log is true. We need to set the Error File to reside in the PV."
+    FATAL_ERROR_OPTION="-XX:ErrorFile=/var/lib/jenkins/logs"
```
bparees: the PR discussion implied the path would be configured by the env var.  i'm not sure hardcoding it to the PV is wrong, but was there a reason you did it this way instead?

openshift/templates/jenkins-persistent.json
```diff
@@ -316,6 +320,12 @@
       "displayName": "Jenkins ImageStreamTag",
       "description": "Name of the ImageStreamTag to be used for the Jenkins image.",
       "value": "jenkins:2"
+    },
+    {
+      "name": "FATAL_ERROR_LOG",
+      "displayName": "Fatal Error Log",
+      "description": "When a fatal error occurs, an error log is created with information and the state obtained at the time of the fatal error.",
```
bparees: need to make it clearer that the valid values for this are true+false (pending my comment above about whether the path should be configurable)

2/contrib/s2i/run
Gl4di4torRr: I can move it. The issue has an extra `-` that was throwing it off... I realized that after I moved it to `JAVA_GC_OPTIONS`

2/contrib/s2i/run
Gl4di4torRr: 1. I was thinking the `-XX:ErrorFile` would be an `optional` argument in the deployment config. With that said, I wanted a way to have the log file off by default.
2. I believe the correct place for logs is `/var/log/jenkins`. Reading from https://wiki.jenkins.io/display/JENKINS/Logging. From the user's perspective, I was wondering how intuitive it would be to place the on the PV for `/var/lib/jenkins`. I started created a new PV for logs... But remembered your comment about this being a getting the developer running.

I am fine if we want this to be a ENV var that is defaulted to /var/lib/jenkins/logs. Let me know the direction here. Thanks!

openshift/templates/jenkins-persistent.json
Gl4di4torRr: Will do.

2/contrib/s2i/run
bparees: i think on further reflection i'm ok leaving it this way and someone who wants to control the path explicitly can set the -XX:ErrorFile arg themselves via the JENKINS_OVERRIDES (and not set the FATAL_ERROR_LOG env variable).

Gl4di4torRr: @bparees yes that is the conclusion I came to as well. I will make the changes shortly (this is to address all comments). Thanks!

Gl4di4torRr:
@bparees I have updated the name to follow the other true/false variables. I will add some documentation soon before this gets merged. Let me know if you want me to make more changes. Thanks!

bparees:
lgtm but i think we'll want to hold this until we cut the 3.10 branch.

gabemontero:
@Gl4di4torRr - can you update the README for the new environment variable ... there is a table at https://github.com/jenkinsci/kubernetes-plugin/pull/315/files

@bparees - I will run `make update-examples` in openshift origin after they cut the 3.10 branch

Gl4di4torRr:
@gabemontero Yes will do ASAP.

Gl4di4torRr:
@gabemontero added documentation. Would you like to see any more test results?

gabemontero:
README update looks good @Gl4di4torRr 

If you could squash the commits (that is our general practice), I'll go ahead and get this merged.

thanks again

Gl4di4torRr:
@gabemontero commits have been squashed. Thanks! Excited to help out!

gabemontero:
awesome thanks again @Gl4di4torRr

PR #614, opened 2018-05-24
Amend arguments if none passed to run-jnlp-client

scoheb:
As of version 1.6.0 of the Jenkins Kubernetes plugin,
args are no longer passed to jnlp container if not present in containerTemplate

See https://github.com/jenkinsci/kubernetes-plugin/pull/315

This is made worse if using Declarative pipeline and specifying yaml for the podTemplate.
since ${computer.*} is not resolvable.

To workaround this, we use JENKINS_SECRET and JENKINS_NAME for launcher arguments

See https://gist.github.com/scoheb/f29d263c240e8880648b920240635ed6 for an example Jenkinsfile

### DIFF
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -63,8 +63,25 @@ source /usr/local/bin/configure-agent
 
 set -e
 
+# As of version 1.6.0 of the Jenkins Kubernetes plugin,
+# args are no longer passed to jnlp container if not present in containerTemplate
+# See https://github.com/jenkinsci/kubernetes-plugin/pull/315
+# In this case, we use JENKINS_SECRET and JENKINS_NAME for launcher arguments
+# This is made worse if using Declarative pipeline and specifying yaml for the podTemplate
+# In this case, ${computer.*} is not resolvable.
+
+# If no args are given, we examine env vars and add to arg list.
+if [[ $# -eq 0 ]]; then
+  if [ ! -z "$JENKINS_SECRET" ]; then
+    set -- "${@}" "$JENKINS_SECRET"
+  fi
+  if [ ! -z "$JENKINS_NAME" ]; then
+    set -- "${@}" "$JENKINS_NAME"
+  fi
+fi
+
 # if `docker run` has 2 or more arguments the user is passing jenkins launcher arguments
-if [[ $# -gt 1 ]]; then
+if [[ $(echo "$@" | awk '{print NF}') -gt 1 ]]; then
   JAR="${JENKINS_HOME}/remoting.jar"
   PARAMS=""
 


scoheb:
Failure is not related:

error pulling image configuration: Get https://registry.access.redhat.com/v2/openshift3/ose/blobs/sha256:bf5608105fab14e26da22596ec750372463581dc99d0021e9d8d290f3db32a15: read tcp 172.18.8.68:40084->209.132.182.63:443: read: connection reset by peer

gabemontero:
hey @scoheb - Yeah @bparees ran into Carlos at KubeCon in Denmark a few weeks ago and we got wind of this.

What needs to be done for all this repo's scenarios is broader though then what you've got here, though what you got here does highlight the desire to assist folks who extend our image as well.  Do we some how allow users to specify the list of env vars to check for and transfer to args for example.

Also, we are close to cutting our 3.10 release ... we'd rather defer whatever change is done until 3.11.  I've been trying to hold how cutting the 3.10 branch here to avoid potential dual maintenance, but the post 3.10 work is piling up, so may have to give in soon.  But ultimately that is orthogonal.

Let's use this PR to drill down on the scenarios and possible implementations some more, and yeah, if it make sense to drive the solution here, or at least a logical subset of it, we'll do that, with again the thought it would be post 3.10.

thanks

gabemontero:
Got around to opening https://trello.com/c/j6wvmQPd/1568-jenkinsintegrationpipelineintegration-react-adjust-to-k8s-plugin-param-changes for this ... includes some of @bparees thoughts on resolving this that while at kubecon that I had forgotten.

gabemontero:
OK, @bparees and I were able to pow-wow this morning, and we landed on just using this change @scoheb, convincing ourselves it was the equivalent of what we had envisioned with the referenced trello card.

Also highlighted during that discussion that `run-jnlp-client` was already accounting for pulling JENKINS_URL and JENKINS_TUNNEL from env vars if not set as command line args, which is why our basic regression testing was fine when I did my testing for the k8s plugin pod toleration change I did recently.

The only last sanity check is to make sure that if our images with this change is running with an older k8s plugin, where JENKINS_NAME/JENKINS_SECRET are still passed in as args, thenthe fact that args are duplicated will not break things.

@scoheb - any chance you already tested this change against an older k8s plugin?

If not, I think I can get some cycles either later today or early next week to try it out.  And if it looks good, we'll merge this.

thanks

scoheb:
@gabemontero I have **not** tested it with older versions of the k8s plugin. However, older versions would add the args if not present and the docs suggest adding the args with ${computer.*} anyways.
This change only has an impact if the arg list is empty. Hence, I "believe" it will work with older versions.

I would only be able to test this out late next week too.

gabemontero:
ah - ok / didn't know that (re: what you found in the docs, etc.)

I actually found and started setting up a test a few minutes ago ... I should have results in a bit.

gabemontero:
Yep looks good runs fine

gabemontero:
[[jenkins_integration][pipeline_integration] react/adjust to k8s plugin param changes](https://trello.com/c/j6wvmQPd/1568-jenkinsintegrationpipelineintegration-react-adjust-to-k8s-plugin-param-changes)

scoheb:
Thanks! Is this part of a published image yet?

gabemontero:
On Fri, May 25, 2018 at 3:07 PM Scott Hebert <notifications@github.com>
wrote:

> Thanks! Is this part of a published image yet?
>
the centos images should be on docker hub within the hour if not sooner
unless there is an env hiccup

PR #623, opened 2018-06-05
🔒 Upgrade plugins for Jenkins advisory 2018-06-04

grdryn:
Advisory: https://jenkins.io/security/advisory/2018-06-04/

kubernetes 1.6.2 -> 1.7.1
https://github.com/jenkinsci/kubernetes-plugin/blob/master/CHANGELOG.md

git 3.8.0 -> 3.9.1
https://plugins.jenkins.io/git

github 1.26.0 -> 1.29.1
https://plugins.jenkins.io/github

github-branch-source 2.2.3 -> 2.3.6
https://plugins.jenkins.io/github-branch-source

PR #624, opened 2018-06-06

gabemontero:
See recent test failures in https://github.com/openshift/jenkins/pull/623

@openshift/sig-developer-experience @openshift/sig-continuous-delivery @grdryn fyi


gabemontero:
@Gl4di4torRr - fyi - this is what is blocking your PR

Issue #619, opened 2018-05-30
Set Jenkins identity in build slaves and master for pushing tags


springdo:
In a pipeline; I want to be able to push git tags to a repository. When doing this in the npm slaves (and assumedly maven & master too) you get an error that Git doesn't know who you are. As the CI server is a service account and the slave images are ephemeral; it doesn't have a proper `user.name` or `user.email` set causing the tagging to fail builds. I suggest we add some config for Jenkins such as the following

```
# Set Jenkins for git pushing tags etc in pipeline
RUN git config --global user.email "jenkins@my-ci-server.com" && \
    git config --global user.name "jenkins"
```

Open to debate and happy to raise a PR

bparees:
you can also work around this by setting the GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL env vars.

i'd sort of rather people do this explicitly lest we get questions about where the hardcoded value is coming from, but i'm not 100% against setting a default.

Gl4di4torRr:
@bparees what are you think the default values should be?

bparees:
`openshift@openshift.com` and `openshift` would be my recommendation i guess.

PR #627, opened 2018-06-12
Add default git config

Gl4di4torRr:
Opening this PR for #619.

Just to show the config is working in the centos image...
```
bash-4.2$ git config --list
user.email=openshift@openshift.com
user.name=openshift
```

I understand this is pending "what will/should this base out of the box jenkins master contain."

Thanks!

fixes https://github.com/openshift/jenkins/issues/619

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -89,7 +89,10 @@ RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
     /usr/local/bin/fix-permissions /opt/openshift && \
     /usr/local/bin/fix-permissions /opt/openshift/configuration/init.groovy.d && \
     /usr/local/bin/fix-permissions /var/lib/jenkins && \
-    /usr/local/bin/fix-permissions /var/log
+    /usr/local/bin/fix-permissions /var/log && \
+    git config --global user.email "openshift@openshift.com" && \
+    git config --global user.name "openshift"
+
 
 VOLUME ["/var/lib/jenkins"]
 
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -106,7 +106,10 @@ RUN rm /opt/openshift/base-plugins.txt && \
     chown 1001:0 /usr/lib/jenkins/*hpi && \ 
     /usr/local/bin/fix-permissions /opt/openshift/configuration/init.groovy.d && \
     /usr/local/bin/fix-permissions /var/lib/jenkins && \
-    /usr/local/bin/fix-permissions /var/log
+    /usr/local/bin/fix-permissions /var/log && \
+    git config --global user.email "openshift@openshift.com" && \
+    git config --global user.name "openshift"
+
 
 VOLUME ["/var/lib/jenkins"]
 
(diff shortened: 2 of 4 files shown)


bparees:
/ok-to-test

@Gl4di4torRr can you clean up/squash the commits (and also the typo in the commit) and then lgtm.

Gl4di4torRr:
@bparees when i squash the merge, it shows that I have changed all the files in the merge... I will work on cleaning this up this evening.

bparees:
@Gl4di4torRr it may help if you rebase on the latest master

Gl4di4torRr:
@bparees yes - will do. sorry other items have come up... i will do this asap.

Gl4di4torRr:
@bparees done

2/Dockerfile
```diff
@@ -91,6 +91,10 @@ RUN /usr/local/bin/install-plugins.sh /opt/openshift/base-plugins.txt && \
     /usr/local/bin/fix-permissions /var/lib/jenkins && \
     /usr/local/bin/fix-permissions /var/log
 
+# Set default git configuration
+RUN git config --global user.email "openshift@openshift.com" && \
+    git config --global user.name "openshift"
```
bparees: please append these commands to the existing RUN command above, so it doesn't create a new layer.

bparees:
@Gl4di4torRr were you going to do this in the slave-base dockerfile also?

2/Dockerfile
Gl4di4torRr: good idea

Gl4di4torRr:
@bparees yes i can do that as well

Gl4di4torRr:
@bparees i added the git config to the base slave images and added the config to the other layer.

gabemontero:
Shouldn't we have something in the README that at least briefly notes the fact that we are setting this and the why we are setting this?  I'm still shaking the cobwebs off after PTO, but users don't necessarily have to override, right?  If they did, obviously the README should get into that as well ....

bparees:
Hm, actually I need to revisit the problem we're solving here.  I *thought* #619 was reporting the known issue that the container is running with a random uid that isn't in /etc/passwd and as a result, git operations fail because they can't look up the current uid.

But on revisiting the description of the issue, it sounds like the problem was that when performing a write operation to git, it fails because you haven't identified yourself.  But it's not clear to me how setting a default git user would solve that, since presumably the default git user doesn't have permissions to perform the write (in this case git tag) operation anyway.

@Gl4di4torRr this is perhaps just a gap in my git knowledge, but how does setting a default/arbitrary git username/email help your scenario?  What additional steps do you still need to take even w/ this change, to make your flow work?

Gl4di4torRr:
@springdo will you please elaborate on this for @bparees

bparees:
@springdo bump

gabemontero:
@springdo @Gl4di4torRr bump on @bparees 's question in https://github.com/openshift/jenkins/pull/627#issuecomment-397639529 and if we move forward, a readme update per my https://github.com/openshift/jenkins/pull/627#issuecomment-397636413

Gl4di4torRr:
Unless @springdo provides justification for moving this forward, I am okay with just closing this out.

PR #635, opened 2018-07-03
remove dead link to java options

bparees:


bparees:
@gabemontero these links are dead and IT is complaining about us having internal hostnames exposed in public repos.

PR #636, opened 2018-07-12
bump container cap to 100

Gl4di4torRr:
Due to the nature of Jenkins, I have found the container cap to be quite low. Even for new consumers testing the OpenShift Jenkins, I believe will hit this cap very quickly which turns into a poor user experience. IMHO.


gabemontero:
thanks @Gl4di4torRr !

PR #643, opened 2018-07-16
replace the term docker with container

bparees:


bparees:
@gabemontero ptal, request from rcm.  unfortunately going to need to be backported as well.

gabemontero:
/lgtm

but the error in the failed test run is an odd one ... it is during the docker build, though I don't see how your change could have induced it

bparees:
indeed that is weird... let's try a retest and then i'll investigate further

/retest

2/Dockerfile.rhel7
```diff
@@ -27,11 +27,10 @@ LABEL io.k8s.description="Jenkins is a continuous integration server" \
       io.openshift.s2i.scripts-url=image:///usr/libexec/s2i
 
 # Labels consumed by Red Hat build service
-LABEL com.redhat.component="openshift-jenkins-2-docker" \
+LABEL com.redhat.component="openshift-jenkins-2-container" \
       name="openshift3/jenkins-2-rhel7" \
       version="3.11" \
       architecture="x86_64" \
-      release="7"
```
bparees: @gabemontero it was this. i need to remove the trailing slash above.

bparees: weird error/failure mode though.

PR #646, opened 2018-07-16

gabemontero:
@bparees @Gl4di4torRr fyi / ptal


gabemontero:
actually the merge fix left an extraneous rpm -V call ... updating ...

PR #647, opened 2018-07-16

gabemontero:
@bparees @Gl4di4torRr fyi / ptal

reminder - even though we did not ship 3.8, this change will help on the dist-git side of things, and the release still exists there and they have hit this issue there as well


2/Dockerfile
```diff
@@ -35,7 +35,9 @@ RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/j
     yum install -y centos-release-scl-rh && \
     INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip java-1.8.0-openjdk java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel java-1.8.0-openjdk-devel.i686 jenkins-2.73.3-1.1" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
-    rpm -V $INSTALL_PKGS && \
+    # have temporarily removed the validation for java to work around known problem fixed in fedora; jupierce and gmontero are working with
+    # the requisit folks to get that addressed ... will switch back to rpm -V $INSTALL_PKGS when that occurs
+    rpm -V dejavu-sans-fonts rsync gettext git tar zip unzip dumb-init jenkins-2.89.4-1.1 && \
```
gabemontero: ahhh ... need to fix this

PR #660, opened 2018-07-19
Adds support for http proxy for maven agent/slave

akram:


bparees:
@akram you'll need to add this to the agent-maven script too.

gabemontero:
@akram - address the comment from @bparees and I'll put the tag to merge this

gabemontero:
bump @akram  - any chance you can address @bparees 's comments so we can merge this?

akram:
@gabemontero PTAL again :)

Ananasr:
Can you merge please ?

akram:
rebased and squashed

gabemontero:
yep that is what @bparees meant @akram

bparees:
github lost my comment, but if you look here: https://github.com/openshift-s2i/s2i-wildfly/pull/171

we recently switched to using the standard "http_proxy" env var for maven proxy support in the wildfly image.  I think we should do the same here, rather than the non-standard http_proxy_host/port vars.  Since these images never supported http_proxy_host/port, we don't have to handle both (like wildfly does), just http_proxy.  The only annoying bit is having to handle splitting out the port (unless maven supports you passing host:port as a single field?)

PR #667, opened 2018-07-26
switch to new registry

bparees:


bparees:
@gabemontero sorry i think i answered your question incorrectly before.  we want the imagestream updated.  what i didn't care about updating was the slave pod template references inside the v1 image.

bparees:
(we want this updated so people don't get imagestream import failures at some point in the future if the old registry becomes unavailable)

gabemontero:
@bparees nw / 10-4

PR #673, opened 2018-08-03
Update README.md

gabemontero:
prow tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Jenkins Docker Image
 ====================
 
 This repository contains Dockerfiles for Jenkins Docker images intended for
-use with [OpenShift v3](https://github.com/openshift/origin)
+use with [OpenShift v3](https://github.com/openshift/origin)  
 
 For an example of how to use it, [see this sample.](https://github.com/openshift/origin/blob/master/examples/jenkins/README.md)
 


gabemontero:
So made some progress on launching origin extended tests from a jenkins PR via ci-operator.  The changes to access Jenkins REST via an execpod are working.

We even have one passing test.

However, our pipelines are now failing with errors like not being able to find the `oc` embedded in the Jenkins instance:

```
[workspace] Running shell script
+ oc start-build openshift-jee-sample-docker --from-file=target/ROOT.war -n e2e-test-jenkins-pipeline-czkqm
/var/lib/jenkins/jobs/e2e-test-jenkins-pipeline-czkqm/jobs/e2e-test-jenkins-pipeline-czkqm-openshift-jee-sample/workspace@tmp/durable-90635278/script.sh: line 2: oc: command not found
```

`oc` is something the jenkins image build injects into the image, injecting it into one of the directories in the path ... from what I see in the image build log, it seems we are injecting it where we usually do.

Will need to investigate the PATH variable that is created when the image is built via ci-operator, or other related factors, and look for any differences.

Also need to investigate all the failures, see if there is anything additional.

@bparees @smarterclayton @stevekuznetsov fyi ... and if something obvious to you comes in mind, of course chime in here.  Otherwise I'll report back when I have some findings.

bparees:
> Will need to investigate the PATH variable that is created when the image is built via ci-operator, or other related factors, and look for any differences.

yeah dumping the $PATH and also do and ls of the dir where you expect `oc` to be seem like reasonable next steps.  Since this command is part of a Jenkinsfile, i assume we're confident it's executing inside the jenkins container, right?

gabemontero:
On Fri, Aug 3, 2018 at 2:35 PM Ben Parees <notifications@github.com> wrote:

> Will need to investigate the PATH variable that is created when the image
> is built via ci-operator, or other related factors, and look for any
> differences.
>
> yeah dumping the $PATH and also do and ls of the dir where you expect oc
> to be seem like reasonable next steps. Since this command is part of a
> Jenkinsfile, i assume we're confident it's executing inside the jenkins
> container, right?
>
yeah the command not found error came from pipeline job run output

It without question is in a jenkins

gabemontero:
So first, a correction in my description of how `oc` gets injected into the jenkins image.
Today, the jenkins image starts with "FROM openshift/origin" (for our centos image) and
"FROM registry.access.redhat.com/openshift3/ose" (for our rhel image)

We hence pick up the `oc` binary from those base images.  We no longer download
the oc archive, then extract and copy over.

The `oc` binary is at /usr/bin in these cases.

It also is present at /usr/bin in the "IMAGE_TESTS" image, based
on my tests with ci-operator and using a reduced form of cluster-e2e
template.  That test image is based on openshift/origin-v3.11:tests.

However, I added ls -lR dumps of /usr/libexec and /usr/bin in a debug
version of the Jenkins Dockerfile in the openshift/jenkins:gabemontero-patch-2
branch.

Sure enough, there is no `oc` to be found.

In my master.json config file for ci-operator, I'm current basing
my Jenkins image off of openshift/origin-v3.11:base

Looking at the tags in the openshift namespace on api.ci
I saw origin-v3.11:cli.

After switching to that, sure enough, the `oc` binary is
at the same /usr/bin location it was in the other examples.

I'll be crafting a PR momentarily.

gabemontero:
(the jenkins image should include the `oc` binary now with the new master.json)

gabemontero:
OK, test failed, but progressed to the next layer of the onion, now that `oc` is available.  The tests did get much further along.  Will start posting analysis / next steps as I uncover them.

gabemontero:
The first type of error is trying to create a temp git repo for some of our build testing.  As part of that we call https://github.com/openshift/origin/blob/master/test/util/client.go#L35-L37, which is suppose to use the test namespace in creating the dir.

The `Namespace()` call is returning "integration" (typically the test namespaces are some form of random string) and then subsequent attempts to access the file fail:

```
stat /tmp/openshift-integration: no such file or directory
```
integration is the default name used after not finding any value for the env var `OS_TEST_NAMESPACE`

I do see in our extended builds controller compat test we explicitly set the env:  https://github.com/openshift/origin/blob/master/test/extended/builds/controller_compat.go#L25

My guess is this is a requirement for running ext tests either via the e2e framework, or on prow/gcp clusters, or some such.

Pending a course correction from someone, I'll craft a PR which sets that env var in the same fashion for the jenkins tests.

gabemontero:
The second flavor of error is trying to access files based on the `ARTIFACT_DIR` env var.  It is missing in these runs.  The jenkins tests actually use this setting as part of its env var subtitution for some of the jobs it defines (things like NAMESPACE for example).

In investigating how it is set, it appears to be set in the various extended test scripts in the openshift/origin repo.  Those scripts are bypassed with our e3e templates, which invoke `ginkgo` and the `extended.test` binary directly.

I could again test for the value of ARTIFACT_DIR early in the tests and set if it is unset to the test templates well known location .... maybe explicitly setting ARTIFACT_DIR in the test template is another option.

bparees:
> Pending a course correction from someone, I'll craft a PR which sets that env var in the same fashion for the jenkins tests.

seems like this should be set more universally so other tests don't run into this down the road.

gabemontero:
no more signs of the `ARTIFACT_DIR` complaint with the latest run, but the `stat /tmp/openshift-integration: no such file or directory` is still showing up.

Investigating.....

gabemontero:
nope ... e2e failure same as before.

waiting on https://github.com/openshift/origin/pull/20563 to merge, and the openshift/origin-v3.11:test IST on api.ci to get subsequently updated

gabemontero:
we have a successful e2e run against all the jenkins ext test in openshift/origin using the jenkins-2-centos:latest image built from this branch @bparees

gabemontero:
@stevekuznetsov @smarterclayton fyi as well ... I got the e2e test for jenkins working with the latest run https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_jenkins/673/pull-ci-jenkins-e2e-canary/13/

thanks again for all the assistance

PR #681, opened 2018-08-16
remove bad jsonpath reference

bparees:


bparees:
on further reflection this should not need to be done.

PR #689, opened 2018-09-10
Update the nodejs slave image to version 8

piyush1594:
Hey all,

Raising this pull request if it is feasible to update the nodejs slave image to version 8

If there is something blocker or breaking in this, feel free to advise as I am not aware of that.

It will be good if something like -> for current version ( node 4) -> image `openshift/jenkins-slave-nodejs-centos7

and after update, node 8 -> image `openshift/jenkins-slave-nodejs8-centos7` or may be vice versa

cc @gabemontero @sthaha @hrishin

### DIFF
--- a/slave-nodejs/Dockerfile
+++ b/slave-nodejs/Dockerfile
@@ -2,7 +2,7 @@ FROM openshift/jenkins-slave-base-centos7
 
 MAINTAINER Ben Parees <bparees@redhat.com>
 
-ENV NODEJS_VERSION=4.4 \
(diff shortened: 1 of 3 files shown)


bparees:
Thanks, but we already have one, see:
https://hub.docker.com/r/openshift/jenkins-agent-nodejs-8-centos7/

https://github.com/openshift/jenkins/tree/master/agent-nodejs-8

piyush1594:
Thanks a lot @bparees

PR #690, opened 2018-09-11
bump to v4.0

bparees:


bparees:
@gabemontero ptal, not sure i hit everywhere that needs to be touched.

the v3.11 branch has been created.

bparees:
> The Dockerfile (both centos and rhel) have version specific env vars

thanks, updated.

Issue #693, opened 2018-09-12
Config environment variables should understand 'false' values


atgreen:
Unlike many other config environment variables we see with the jenkins image, and openshift in general, the OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG, OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS, OVERRIDE_RELEASE_MIGRATION_OVERWRITE and SKIP_NO_PROXY_DEFAULT (at least) variables treat any value, even 'false', as affirmative.  Common practice, even with other variables used in jenkins deployments like OPENSHIFT_ENABLE_OAUTH, treat 'false' as false.

I've seen this trip people up, and consider it a usability problem.  Please allow for 'false' values for these variables.

gabemontero:
testing a change locally now ... should have a PR up soon

atgreen:
Awesome -- thanks for the quick change!   However, the documentation in the README file has to change as well.

PR #694, opened 2018-09-13
disable various override paths based on the associated env vars == false

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/693


atgreen:
Thanks.  The documentation in the README needs to change as well.

PR #706, opened 2018-09-28
Upgrade to job-dsl 1.70

svenstaro:
I think 1.70 has some [excellent changes](https://github.com/jenkinsci/job-dsl-plugin/wiki#release-notes). Please consider this.

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -35,7 +35,7 @@ htmlpublisher:1.16
 mailer:1.21
 config-file-provider:2.17
 docker-commons:1.11
-job-dsl:1.68
+job-dsl:1.70
 parameterized-trigger:2.35.2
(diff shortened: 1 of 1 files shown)


gabemontero:
thanks @svenstaro ... agree the updates seem ok

assuming this passes our regression testing the centos image on dockerhub should be upgraded later today; for the rhel image, this will be part of the 4.0 image shipped most likely at the end 2018

svenstaro:
@gabemontero can you check out the failure? I don't get it. I think it's just a timeout of some kind.

gabemontero:
I also have tests running again in https://github.com/openshift/jenkins/pull/708 ... a PR that has already passed once ... we'll see what happens now against a later openshift/origin

gabemontero:
now we are hitting ci hiccups

will see if there is a known env issue ...then retry in a bit

PR #708, opened 2018-09-28
bump centos to jenkins 2.138 (rhel handled by separate internal ci job)

gabemontero:
let's let our e2e ci job do some validation work for us

if our k8s plugin tests fail I'll try bumping that version (they are up to 1.12.6 now)

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -36,11 +36,11 @@ RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/j
     yum install -y centos-release-scl-rh && \
     curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
     x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk jenkins-2.121.3-1.1" && \
+    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk jenkins-2.138.1-1.1" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS $x86_EXTRA_RPMS && \
     # have temporarily removed the validation for java to work around known problem fixed in fedora; jupierce and gmontero are working with
     # the requisit folks to get that addressed ... will switch back to rpm -V $INSTALL_PKGS when that occurs
-    rpm -V  dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init jenkins-2.121.3-1.1 && \
+    rpm -V  dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init jenkins-2.138.1-1.1 && \
     yum clean all  && \
     localedef -f UTF-8 -i en_US en_US.UTF-8
 


gabemontero:
ooooh ... the ext tests passed ... those include the k8s agent tests

going to do a manual test to confirm, but whatever happened in august isn't happening now ... perhaps a remoting jar fix got dropped in the interim (see nothing on the k8s plugin side in this space).

gabemontero:
yep it works locally for me too (the exact test case I tried back in August)

gabemontero:
spoke too soon ... forgot that the error did not occur on initial startup

it occurs on shutdown/close

still curious why this did not show up in the prow e2e ... I'll have to examine the logs from a run of that job, do some cross referencing

gabemontero:
ah ... slight difference ... the build actually succeeded for me, even though the slave pod error'ed out:

```
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
```
and 

```
gmontero ~/Downloads/openshift-origin-client-tools-v3.10.0-dd10d17-linux-64bit $ ./oc get pods 
NAME                             READY     STATUS      RESTARTS   AGE
jenkins-1-8xc6l                  1/1       Running     0          14m
nodejs-fqvvh                     0/1       Error       0          9m
nodejs-mongodb-example-1-build   0/1       Completed   0          9m
nodejs-mongodb-example-1-s25pz   1/1       Running     0          8m
```

Undoubtedly this is what happened in the e2e run (though I'll still see about confirming).

gabemontero:
Hmm ... I'm seeing the same behavior at 2.121.3 as well, with the latest jenkins image.

Again, we this first came up, the job seemed to be failing as a result of the slave pod ending up in an error state.  That is not happening now.  

Starting to cycle back to older images, see if I can tell when the pattern changes.

gabemontero:
Yeah the behavior at 3.10 is different, which is jenkins 1.107.3 ... the termination path is different, and the pod gets deleted.

Variables to consider:
a) remoting jar changes from 2.107 to either 2.121.3 or 2.138.1 LTS ... the logs on termination do look different ... most likely a large amount of changes there
b) I don't *think* the pod retention stuff is coming into, as I *think* we left it off by default, but will need to verify that
c) other k8s plugin changes ... our 3.10 image is a 1.2 ... a far cry from 1.12.1 we ship in later releases

I'll start by bumping the k8s plugin with the 3.10 image.

gabemontero:
Couldn't help myself ... it defaults to deleting the pod after `activeDeadlineSeconds` has passed.  If this is not set, they per the openshift code, is 6 hours:

```
	// MaxDeploymentDurationSeconds represents the maximum duration that a deployment is allowed to run.
	// This is set as the default value for ActiveDeadlineSeconds for the deployer pod.
	// Currently set to 6 hours.
	MaxDeploymentDurationSeconds int64 = 21600
```

We are not setting activeDeadlineSeconds as part of our default pod templates for maven/nodejs.

So yeah, I'll explore b) from https://github.com/openshift/jenkins/pull/708#issuecomment-425582354 next week.... if I can make that to work to ignore post terminate issues, that would be the ideal

Otherwise, maybe some config change to change activeDeadlineSeconds in our pod templates at least is a suitable band aid.

And forgot to mention earlier ... addressing the timing windows I noted on slave pod terminate (various metadata removed so that they no longer find the id for our slave for example) is not something I think we should take on.

bparees:
Naive comments/questions:

It seems like there are two problems:

1) podretention behavior is occuring even though pod retention is not enabled?
2) pods are terminating in an error state (and thus being retained) when they should have exited cleanly?

Is that an accurate summary of the problems?

gabemontero:
On Fri, Sep 28, 2018 at 6:54 PM Ben Parees <notifications@github.com> wrote:

> Naive comments/questions:
>
> It seems like there are two problems:
>
>    1. podretention behavior is occuring even though pod retention is not
>    enabled?
>
> yes, the default when not explicitly specified is pod stays for
activeDeadlineSeconds; looks like
if that is not explicitly set, the pod sticks around long enough to be
noticed :-)

>
>    1. pods are terminating in an error state (and thus being retained)
>    when they should have exited cleanly?
>
> I'm note sure if the pod erroring out or not matters wrt retention, I'd
have to revisit the code change more closely,
but yes, it seems like they should have exited cleanly, but k8s plugin and
remoting jar have hiccups ... k8s plugin
most likely never worried about it given the nuke immediately approach.

I don't know if the test I ran, with our nodejs agent, is the exception or
the norm wrt exiting cleanly.

gabemontero:
Well, actually, with 0e7e9dc08bbb3d3ce4e9fb2a9d2ff455482efa76, it looks like the default always delete, so maybe it is only on error.

When I dive into things on Monday I'll add some tracing, etc. and confirm some of these particulars.

gabemontero:
Also, since I'm seeing the same thing on 2.121 and 2.107, we can make this update now if we want, and I'll work the k8s plugin jira separately.

I've opened https://github.com/openshift/aos-cd-jobs/pull/1597 to update the internal CI job to allow us the change the version for the 4.0 jenkins rhel image

bparees:
> Also, since I'm seeing the same thing on 2.121 and 2.107, we can make this update now if we want, and I'll work the k8s plugin jira separately.

seems reasonable to me.  @adambkaplan will be back in a week and can look into it since he implemented the retention feature.

gabemontero:
I'll take a crack at it this week ... hopefully at least have a working change ready.  But I can certainly hold off on a k8s plugin PR submission until @adambkaplan gets back and has a chance to provide input.

gabemontero:
Job https://buildvm.openshift.eng.bos.redhat.com:8443/job/devex/job/devex%252Fjenkins-bump-version/23/ should update the RPM used to install jenkins in our 4.0 images

gabemontero:
Got things sorted out ... https://buildvm.openshift.eng.bos.redhat.com:8443/job/devex/job/devex%252Fjenkins-bump-version/24/console did the update

PR #711, opened 2018-10-02
Fix bash syntax errors

### LINKED ISSUE #710: Bash syntax errors
We founds somme syntax errors:

/usr/libexec/s2i/run: line 283: [: missing `]'
/usr/libexec/s2i/run: line 284: : command not found
/usr/libexec/s2i/run: line 383: [: missing `]'
/usr/libexec/s2i/run: line 403: [: missing `]'

mbecca:
fixes #710

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -280,8 +280,8 @@ for FILENAME in ${JENKINS_HOME}/plugins/*.jpi; do
 done
 
 shopt -s nocasematch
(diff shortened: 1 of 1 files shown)


gabemontero:
a bit curious on why this only fails sometimes, as this is in the main path and so is part of everyone's startup of the jenkins image

@bparees @mbecca any thoughts?

bparees:
i suspect this only fails when run with /bin/sh, not /bin/bash, or some other nuance.

gabemontero:
thanks @bparees 

and thanks @mbecca

PR #715, opened 2018-10-09
Dockerfile edits and set java path for multiarch builds 

jeffdyoung:
Dockerfile changes needed to build ppc64le + x86 containers from same Dockerfile.

Adding  java/javac to 'alternatives' logic so ppc64le (and other archs) can successfully start jenkins.

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -42,7 +42,7 @@ RUN yum-config-manager --disable epel >/dev/null || : && \
     # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
     ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
     x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \
-    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
+    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk java-1.8.0-openjdk-devel atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
     yum install -y $INSTALL_PKGS $x86_EXTRA_RPMS && \
     # have temporarily removed the validation for java to work aroun known rhel 7.5 problem fixed in fedora; jupierce and gmontero are working with
     # the requisit folks to get that addressed ... will switch back to rpm -V $INSTALL_PKGS when that occurs
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -186,6 +186,10 @@ if [[ "$(uname -m)" == "x86_64" ]]; then
 	alternatives --set javac $(alternatives --display javac | awk '/family.*i386/ { print $1; }')
 	export MALLOC_ARENA_MAX=${MALLOC_ARENA_MAX:-1}
     fi
+#set JVM for all other archs
+else
+        alternatives --set java $(alternatives --display java | awk '/family.*'$(uname -m)'/ { print $1; }')
+        alternatives --set javac $(alternatives --display javac | awk '/family.*'$(uname -m)'/ { print $1; }')
 fi
 
 echo "OPENSHIFT_JENKINS_JVM_ARCH='${OPENSHIFT_JENKINS_JVM_ARCH}', CONTAINER_MEMORY_IN_MB='${CONTAINER_MEMORY_IN_MB}', using $(readlink /etc/alternatives/java) and $(readlink /etc/alternatives/javac)"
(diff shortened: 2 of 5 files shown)


jeffdyoung:
@gabemontero I'm not sure if these changes can be cherry picked into 3.10/3.11. If not, I will gladly open new PRs for each. Thanks!

2/Dockerfile.rhel7
```diff
@@ -42,7 +42,7 @@ RUN yum-config-manager --disable epel >/dev/null || : && \
     # will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images do *NOT* have a /usr/lib64/jenkins path
     ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
     x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \
-    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
+    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk java-1.8.0-openjdk-devel atomic-openshift-clients jenkins-2.* jenkins-2-plugins" && \
```
gabemontero: Can you elaborate on how adding the devel package assists us here?

2/contrib/s2i/run
```diff
@@ -186,6 +186,10 @@ if [[ "$(uname -m)" == "x86_64" ]]; then
 	alternatives --set javac $(alternatives --display javac | awk '/family.*i386/ { print $1; }')
 	export MALLOC_ARENA_MAX=${MALLOC_ARENA_MAX:-1}
     fi
+#set JVM for all other archs
```
gabemontero: github review won't let me add a review comment at line 176, but change `multi-arch check ... if not x86_64, leave alone` to `multi-arch check ... if x86_64, pick between 32 and 64 bit`

agent-maven-3.5/Dockerfile.rhel7
```diff
@@ -20,7 +20,7 @@ ENV MAVEN_VERSION=3.5 \
 RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-server-rhscl-8-rpms && \
     yum-config-manager --disable epel >/dev/null || : && \
-    INSTALL_PKGS="java-1.8.0-openjdk-devel.x86_64 rh-maven35*" && \
+    INSTALL_PKGS="java-1.8.0-openjdk-devel rh-maven35*" && \
```
gabemontero: Am I correct in inferring that `java-1.8.0-openjdk-devel` == `java-1.8.0-openjdk-devel.x86_64` ?

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -29,6 +29,9 @@ if [[ "$(uname -m)" == "x86_64" ]]; then
       alternatives --set java $(alternatives --display java | awk '/family.*i386/ { print $1; }')
       export MALLOC_ARENA_MAX=${MALLOC_ARENA_MAX:-1}
     fi
+#set JVM for all other archs
```
gabemontero: similar update to the comment in line 21 to what I asked for in the main image run script

gabemontero:
We do have a cherrypick command I can try once this merges @jeffdyoung  ... i'll mention you when I trie it ... but if it doesn't work, then yes, please update 3.10 and 3.11 as needed for both x86 and ppc64le to run

2/Dockerfile.rhel7
jeffdyoung: You are installing it on  x86:
`x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \`

I added it to INSTALL_PKGS so it could be picked up by all archs..unless you just the i686 version only for x86?

slave-base/contrib/bin/run-jnlp-client
jeffdyoung: ack. new commit coming soon.

agent-maven-3.5/Dockerfile.rhel7
jeffdyoung: yes, you are correct. And in our case:
`java-1.8.0-openjdk-devel == java-1.8.0-openjdk-devel.ppc64le`

2/contrib/s2i/run
jeffdyoung: ack. new commit coming soon.

2/Dockerfile.rhel7
gabemontero: ah missed that ... thanks

jeffdyoung: Actually.. this logic will install java-1.8.0-openjdk-devel.i686 and java-1.8.0-openjdk-devel.x86_64 on x86...I think that is still ok?

gabemontero: yes we want both for x86

gabemontero:
looks good to me

I'll wait on the lgtm label for
a) seeing if we get lucky re: infra resources and get a clean build
b) allow a few more cycles in case @bparees wants to chime in

bparees:
lgtm.

gabemontero:
env constraint pain with the e2e job  ... jenkins, registry

thinking we need to make the parallel job reduction changes @bparees and I discussed at lunch in our openshift/release artifacts

we got yum flake in dumb-init on the raw jenkins build

I did see both 32 and 64 bit java installed

gonna re-run that test just for safety sake ... assuming it passes, I'm having @bparees hit the green button on this one, while we sort out our e2e separately

gabemontero:
ok close enough re: latest test run

@bparees please merge button this one as part of expediting the recovery from the missing 32bit jvm corruption recently introduced

We can run e2e tests with new PRs after https://github.com/openshift/release/pull/1176 gets merged

PR #719, opened 2018-10-10
various new launcher related fixes; add verifyService to client plugin

gabemontero:

PR #721, opened 2018-10-11
Update openshift-client plugin to 1.0.20

svenstaro:
This updates the openshift-client plugin to 1.0.20 so that trailing builds in Jenkins builds will work.


gabemontero:
We already have https://github.com/openshift/jenkins/pull/719 for this @svenstaro 

We just have some issues with our ci so it has not merged yet.

PR #736, opened 2018-11-05
suggestion

WolverMinion:
if your plugins.txt is empty, then *.lock will not be created and your code in line 433 crashed with 
``rm: cannot remove ‘/opt/openshift/plugins/*.lock’: No such file or directory``

is not good for me 😉


gabemontero:
thanks @WolverMinion !

PR #748, opened 2018-12-05
Allow plugin download to use an insecure repository

akram:
Allow plugin download to use an insecure repository (signed by unknown CA or self signed certificate)

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -191,11 +191,11 @@ function download() {
 }
 
 function doDownload() {
-    local plugin version url jpi
+    local plugin version url jpi curl_options
     plugin="$1"
     version="$2"
     jpi="$(getArchiveFilename "$plugin")"
-
+    curl_options=""
     # If plugin already exists and is the same version do not download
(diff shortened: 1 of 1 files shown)


akram:
/assign @bparees 
@bparees can you PTAL ?
I tried to avoid introducing a security breach by using local variable.

bparees:
/lgtm

but we're in a bit of holding pattern until our CI is back up which should be soon.

/hold

gabemontero:
The install plugins command failed during the build:

```
[91m/usr/local/bin/install-plugins.sh: line 228: syntax error in conditional expression: unexpected token `;'
```

You change will need to be adjusted @akram

2/contrib/jenkins/install-plugins.sh
```diff
@@ -225,8 +225,13 @@ function doDownload() {
         url="$JENKINS_UC_DOWNLOAD/plugins/$plugin/$version/${plugin}.hpi"
     fi
 
+    if [[ "$JENKINS_UC_INSECURE" = true || "$JENKINS_UC_INSECURE" = 1]]; then
```
bparees: i think it's a missing space between the `=1` and the `]]`  ?

akram:
Due to deletion of my original branch and repo I recreated another pull request with the latest repo.

PR #749, opened 2018-12-07
bump lts for https://jenkins.io/security/advisory/2018-12-05/ and htt…

gabemontero:
…ps://bugzilla.redhat.com/show_bug.cgi?id=1656865

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -36,11 +36,11 @@ RUN curl https://pkg.jenkins.io/redhat-stable/jenkins.repo -o /etc/yum.repos.d/j
     yum install -y centos-release-scl-rh && \
     curl https://copr.fedorainfracloud.org/coprs/alsadi/dumb-init/repo/epel-7/alsadi-dumb-init-epel-7.repo -o /etc/yum.repos.d/alsadi-dumb-init-epel-7.repo && \
     x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk.i686 java-1.8.0-openjdk-devel.i686 ; fi) && \
-    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk jenkins-2.138.2-1.1" && \
+    INSTALL_PKGS="dejavu-sans-fonts rsync gettext git tar zip unzip openssl bzip2 dumb-init java-1.8.0-openjdk jenkins-2.138.4-1.1" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS $x86_EXTRA_RPMS && \
     # have temporarily removed the validation for java to work around known problem fixed in fedora; jupierce and gmontero are working with
     # the requisit folks to get that addressed ... will switch back to rpm -V $INSTALL_PKGS when that occurs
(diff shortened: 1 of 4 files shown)


gabemontero:
the same unit test 401 vs. 200 rc that @jlebon uncovered with 2.150.1 is happening with 2.138.4 as well

probably whatever security hole they plugged is messing with our tests

Reworking the tests is on the to-do lists now

gabemontero:
OK i've narrowed down the unit test failure to trying to override the non-oauth admin password

gabemontero:
And part of this sec advisory has to do with migrating user records ... not sure yet if that is applicable

gabemontero:
From the advisory:

```
Jenkins no longer uses directory names as a reference for user names, making the on-load migration of user records unnecessary. Instead, the new file users/users.xml is used to map user names to the directories containing the user metadata.
```

Our run script does:

```
function update_admin_password() {
  sed -i "s,<passwordHash>.*</passwordHash>,<passwordHash>$new_password_hash</passwordHash>,g" "${JENKINS_HOME}/users/admin/config.xml"
  echo $new_password_hash > ${JENKINS_HOME}/password
}
```

gabemontero:
I don't know why yet, but with 2.138.4, the following code we have at https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/jenkins-common.sh#L25-L34

```
function obfuscate_password {
    local password="$1"
    local salt="$2"
    local acegi_security_path=`find /tmp/war/WEB-INF/lib/ -name acegi-security-*.jar`
    local commons_codec_path=`find /tmp/war/WEB-INF/lib/ -name commons-codec-*.jar`

    # source for password-encoder.jar is inside the jar.
    # acegi-security-1.0.7.jar is inside the jenkins war.
    java -classpath "${acegi_security_path}:${commons_codec_path}:/opt/openshift/password-encoder.jar" com.redhat.openshift.PasswordEncoder $password $salt
}
```

generates the same hash regardless of using our default of `password` or if I pass in another string in for `JENKINS_PASSWORD`

neither of those jar files are present in neither the 2.138.4 image I'm building locally, or our latest 2.138.2 images on dockerhub.

still not 100% sure what is up yet

gabemontero:
OK, when I run the new image manually, I'm able to login with the new password at least ....

Could it be with these new changes they leverage different password files whatever for login vs. REST invocation? ... i.e. the new files and changes they have introduced ... not getting that from re-reading the advisory, but a lot of the jenkins internal gorp in that is just gorp to me

gabemontero:
Found it ... the container restart was mixing things up vs. starting from scratch with overriding JENKINS_PASSWORD ... it finally kicked in password change logic, and we got 

```
sed: can't read /var/lib/jenkins/users/admin/config.xml: No such file or directory
```

with the security change, they "randomize" the directory name to `/var/lib/jenkins/users/admin_8415419455379157904`

This name mapping is in another xml file ... we either have to parse that (in bash ...yuck!!) or via bash munge the random dir name, given it at least starts with admin.

gabemontero:
e2e-gcp will need https://github.com/openshift/release/pull/2341 plus an additional commit I'm about to push

gabemontero:
ok commit pushed ... openshift-jenkins is passing for me locally ... assuming it passes here let's see the e2e-gcp run after https://github.com/openshift/release/pull/2341

gabemontero:
ok unit tests now passing ... the e2e-gcp run was without the merged https://github.com/openshift/release/pull/2341 so we'll need to retest

PR #751, opened 2018-12-10
Using k8s service names instead of ip address

ErikLundJensen:
Rely on Kubernetes service abstraction instead of writing ip addresses to config.xml.
Note, services may get new ip addresses when updating or reinstalling Jenkins and as a result the config.xml will point to obsolete ip addresses.

### DIFF
--- a/2/contrib/jenkins/kube-slave-common.sh
+++ b/2/contrib/jenkins/kube-slave-common.sh
@@ -13,13 +13,9 @@
 
 export DEFAULT_SLAVE_DIRECTORY=/tmp
 export SLAVE_LABEL="jenkins-slave"
-JNLP_SERVICE_NAME=${JNLP_SERVICE_NAME:-JENKINS_JNLP}
-JNLP_SERVICE_NAME=`echo ${JNLP_SERVICE_NAME} | tr '[a-z]' '[A-Z]' | tr '-' '_'`
-T_HOST=${JNLP_SERVICE_NAME}_SERVICE_HOST
-# the '!' handles env variable indirection so we can resolve the nested variable
-# see: http://stackoverflow.com/a/14204692
-JNLP_HOST=${!T_HOST}
-T_PORT=${JNLP_SERVICE_NAME}_SERVICE_PORT
+JNLP_SERVICE_NAME=${JNLP_SERVICE_NAME:-jenkins-jnlp}
+JNLP_SERVICE_NAME_KEY=`echo ${JNLP_SERVICE_NAME} | tr '[a-z]' '[A-Z]' | tr '-' '_'`
+T_PORT=${JNLP_SERVICE_NAME_KEY}_SERVICE_PORT
 JNLP_PORT=${!T_PORT}
 
 export JNLP_PORT=${JNLP_PORT:-50000}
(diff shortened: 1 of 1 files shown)


2/contrib/jenkins/kube-slave-common.sh
```diff
@@ -149,8 +141,8 @@ function generate_kubernetes_config() {
       <addMasterProxyEnvVars>true</addMasterProxyEnvVars>
       <serverCertificate>${crt_contents}</serverCertificate>
       <namespace>${PROJECT_NAME}</namespace>
-      <jenkinsUrl>http://${JENKINS_SERVICE_HOST}:${JENKINS_SERVICE_PORT}</jenkinsUrl>
-      <jenkinsTunnel>${JNLP_HOST}:${JNLP_PORT}</jenkinsTunnel>
+      <jenkinsUrl>http://${JENKINS_SERVICE_NAME}:${JENKINS_SERVICE_PORT}</jenkinsUrl>
+      <jenkinsTunnel>${JNLP_SERVICE_NAME}:${JNLP_PORT}</jenkinsTunnel>
```
bparees: my main concern with this PR is here:  you're going to set the jenkinsurl to "jenkins:PORT".  Since you didn't put the namespace in the service name, no one outside this namespace will be able to resolve that url.

When we used the service ip, it would work.

you're going to need to include the namespace in the service name variables you are generating so they can be resolved outside the local namespace.

bparees: (it's possible that because this config is only used for the k8s plugin, it doesn't have to resolve outside the local project, i'd like @gabemontero to weigh in on that)

ErikLundJensen: Yes, I get your point. If the namespace of the agents differs in the <clouds> configuration then we need to use a qualified name like <service>.<ns>.svc.cluster.local where ns is the namespace of the Jenkins master.

I'll modify and push when I tested.

ErikLundJensen: Service name is actually resolved without need for further modifications.
This is the log from the agent when \<clouds\> configuration specifies another namespace.
The k8s plugin resolves the service names and supply the full DNS name to the agent at startup:

OPENSHIFT_JENKINS_JVM_ARCH='', CONTAINER_MEMORY_IN_MB='1536', using /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.171-8.b10.el7_5.i386/jre/bin/java and /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.171-8.b10.el7_5.i386/bin/javac
Downloading http://jenkins.demo-j.svc.cluster.local:80//jnlpJars/remoting.jar  ...

The Jenkins configuration should thereby only contain the short service name.

ErikLundJensen: Yes, the download line contain the IP address without my change. For example:

OPENSHIFT_JENKINS_JVM_ARCH='', CONTAINER_MEMORY_IN_MB='8796093022207', using /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.171-8.b10.el7_5.x86_64/jre/bin/java and /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.171-8.b10.el7_5.x86_64/bin/javac
--
  | Downloading http://172.30.193.155:80//jnlpJars/remoting.jar  ...

akram:
@ErikLundJensen This PR still seems valid. But, we will not back port it to 3.x branch.
Would you like to re-submit it with the master branch, so we can get it merged for 4.3 ?

akram:
another concern is also now that we support `http_proxy` and `https_proxy` variables, if we just set the service short name, then, it will cause issues because by default it is not added to `no_proxy` as opposite to `*.svc.cluster.loca`l or `172.16.0.0/16`

So to merge this PR we will also have to be sure that we append $JENKINS_SERVICE_NAME and JNLP_SERVICE_NAME to the list of `no_proxy`

akram:
This PR is stale for more than 1 year
/close

PR #767, opened 2019-01-11
🔒 Upgrade plugin versions for security-1266

grdryn:
From https://jenkins.io/security/advisory/2019-01-08/#SECURITY-1266:

```
SECURITY-1266 / CVE pending

Script Security sandbox protection could be circumvented during the
script compilation phase by applying AST transforming annotations such
as @Grab to source code elements.

Both the pipeline validation REST APIs and actual script/pipeline
execution are affected.

This allowed users with Overall/Read permission, or able to control
Jenkinsfile or sandboxed Pipeline shared library contents in SCM, to
bypass the sandbox protection and execute arbitrary code on the
Jenkins master.

All known unsafe AST transformations in Groovy are now prohibited in
sandboxed scripts.
```

On the Update page in the Plugin Manager inside Jenkins, for the
"Pipeline: Declarative" (pipeline-model-definition), there's the
following warning message in red:

> Warning: This plugin requires dependent plugins be upgraded and at
> least one of these dependent plugins claims to use a different
> settings format than the installed version. Jobs using that plugin
> may need to be reconfigured, and/or you may not be able to cleanly
> revert to the prior version without manually restoring old
> settings. Consult the plugin release notes for details.

PR #776, opened 2019-01-28
Add agent imagestreams

bparees:


bparees:
/assign @gabemontero 

@gabemontero you'll need to plumb this into openshift/library and then into the samples operator (adding logic to the samples operator to replace the image pullspecs w/ the payload values during upsert).  That also means the jenkins-* imagestreams should be exempt from the registry override behavior.

bparees:
note: i intentionally didn't do anything w/ PPC for now.

openshift/imagestreams/jenkins-agent-maven-rhel7.json
```diff
@@ -0,0 +1,48 @@
...
+        "annotations": {
+          "openshift.io/display-name": "Jenkins Maven Agent",
+          "openshift.io/provider-display-name": "Red Hat, Inc.",
+          "description": "Provides a Jenkins Agent with Maven tooling",
+          "iconClass": "icon-jenkins",
+          "tags": "jenkins"
+        },
+        "from": {
+          "kind": "DockerImage",
+          "name": "registry.redhat.io/openshift4/ose-jenkins-agent-maven:v4.0"
```
bparees: note that this doesn't exist/won't exist until GA.  but since the samples operator is going to replace this value during the install, it doesn't matter that much.

gabemontero:
@bparees am I correct in remembering that the openshift/library polling job should pick these up?  but you also mentioned recently that mechanism might be broken ... perhaps that pertains here

but one way or the other, we want to tag these in openshift/library such that they become content for the operator to manage, right?

bparees:
> @bparees am I correct in remembering that the openshift/library polling job should pick these up? 

Not these since they are net new.  If they were changes to the existing imagestreams, yes, polling should in theory pick them up (but may not be working).


> but one way or the other, we want to tag these in openshift/library such that they become content for the operator to manage, right?

yes.

gabemontero:
the latest `e2e-aws-jenkins` failure is the client plugin issue I just cut a fix for ... see https://github.com/openshift/jenkins/pull/767#issuecomment-458668387 on sorting out how we'll manage that bump and the sec advisory bump

bparees:
@gabemontero i'm going to manually merge this one too since it's net new and can't break anything, that way you can proceed w/ wiring up the library import and then samples operator import.

PR #782, opened 2019-02-01
sabotage plugins

gabemontero:
here's some sabotage @bparees

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -1,5 +1,3 @@
-openshift-login:1.0.14
-openshift-client:1.0.24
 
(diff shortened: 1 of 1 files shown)


bparees:
https://www.youtube.com/watch?v=z5rRZdiu1UE

gabemontero:
On Thu, Jan 31, 2019 at 8:48 PM Ben Parees <notifications@github.com> wrote:

> https://www.youtube.com/watch?v=z5rRZdiu1UE
>

Nice. :-)

gabemontero:
fails as we would expect ... can't authenticate for the curl ... i deleted the other two plugins just in case but it was unnecessary

PR #798, opened 2019-02-16
set default memory limit to 1gig in jenkins template

bparees:


bparees:
@smarterclayton per our discussion this afternoon.

@gabemontero @smarterclayton once this propagates to the samples operator+origin, we can remove the 32bit jdk+logic from the jenkins image.

Technically we'd normally still use a 32bit jvm for a 1gig pod (we switch over at 2gigs) but as long as jenkins will run on 64bit w/ 1 gig pod/512meg heap, this will allow us to shrink the image and simplify it.

Of course doing this will make it very difficult/impossible to run jenkins on starter, but that is already effectively the case.

/assign @gabemontero

gabemontero:
as an fyi I typically run with 4Gi (2Gi for the JVM heap per Jenkins recommendations, another 2Gi of OS memory for the various `oc` and `mvn` invocations that can occur in typical pipelines) ...

... but it finally moving off of 512Mi is still a win in my book :-)

bparees:
yeah, i'd love to go higher but i wanted something that won't flat out fail to deploy on starter (since there's a 1gig quota on starter).

PR #799, opened 2019-02-16
Remove the i686 optimization for Java for 4.0

smarterclayton:
/hold

Testing


bparees:
@smarterclayton i think you'll need to fix this up:
https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/run#L170-L190

gabemontero:
and https://github.com/openshift/jenkins/blob/master/slave-base/contrib/bin/run-jnlp-client#L21-L37 @bparees @smarterclayton

gabemontero:
Forgot about this PR ... ended up doing this with https://github.com/openshift/jenkins/pull/812

PR #805, opened 2019-02-25
explicitly cite 3.11 in FROM clauses

gabemontero:
/assign @bparees 

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -1,4 +1,4 @@
-FROM openshift/origin
+FROM openshift/origin:v3.11
 
 # Jenkins image for OpenShift
 #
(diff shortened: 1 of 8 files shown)


bparees:
/lgtm

you'll need to do this in the other branches as well but i think you can wait until you have a need to make changes in those branches to do it... (PRs against those branches will, i think, fail CI until the change is made)

gabemontero:
Yeah good point @bparees , but unless we have to fix a specific customer bugzilla / support case for anything before 3.11, I'm doing to hold on those.

We've already put the stack in the ground for jenkins security advisories that 3.11 and 4.x are the only destinations going forward (with the how-to on using new jenkins images against older clusters)

PR #810, opened 2019-03-05
restructure jenkins core/plugin installs along api.ci vs. osbs/brew l…

gabemontero:
…ines

See https://github.com/openshift/ocp-build-data/pull/48

/assign @bparees 
@adammhaile @sosiouxme fyi

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -20,7 +20,8 @@ ENV JENKINS_VERSION=2 \
     JENKINS_UC=https://updates.jenkins.io \
     OPENSHIFT_JENKINS_IMAGE_VERSION=4.0 \
     LANG=en_US.UTF-8 \
-    LC_ALL=en_US.UTF-8
+    LC_ALL=en_US.UTF-8 \
+    INSTALL_JENKINS_VIA_RPMS=false
 
 LABEL k8s.io.description="Jenkins is a continuous integration server" \
       k8s.io.display-name="Jenkins 2" \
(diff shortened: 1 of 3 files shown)


gabemontero:
/hold

this should merge after https://github.com/openshift/ocp-build-data/pull/48

2/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM quay.io/openshift/origin-cli
+FROM openshift/origin:v3.11
+#FROM quay.io/openshift/origin-cli
```
bparees: why is this changing?  (also referencing v3.11 seems odd)

2/contrib/jenkins/install-plugins.sh
```diff
@@ -1,436 +1,8 @@
...
-
-    if [[ -f $FAILED ]]; then
-        echo -e "\nSome plugins failed to download!\n$(<"$FAILED")" >&2
-        exit 1
-    fi
-
-    echo -e "\nCleaning up locks"
-    rm -rf "$REF_DIR"/*.lock
+    mainInstallPlugins "$@"
 }
```
bparees: doesn't seem like you need to define main() at all.  Just call mainInstallPlugins below.

2/Dockerfile
gabemontero: ah ... that was suppose to be temporary as I didn't want to wrestle with the UBI stuff as well ... will revert

2/contrib/jenkins/install-plugins.sh
gabemontero: Yeah I wasn't sure re: bash nuances if that would somehow mess up existing s2i extensions that leverate install-plugins.sh 

Assuming you don't think of any in the next few minutes, I'll make this change

gabemontero:
Getting rid of `main()` might address this as welll, but what happened when from `install-jenkins-core-plugins.sh`
- sourced install-plugins.sh
- called `install-jenkins-core-plugins.sh` from the dockerfile
it would bypass my logic in `install-jenkins-core.sh` and call the `main()` in `install-plugins.sh`

if, when getting rid of `main()` in `install-plugins.sh`, the all the logic in `install-jenkins-core-plugins.sh` is called, I'll converge back to just `install-plugins.sh`

bparees:
This seems to do what i'd expect and i think roughly maps to what you're trying to accomplish:


```
$ cat wrapper.sh 
#!/bin/sh

source ./inner.sh

echo wrapperlogic
# invokes main from inner.sh
main
```

```
$ cat inner.sh 
#!/bin/sh

main() {
  echo innermain
}

# this gets executed when inner.sh is run, or when inner.sh is sourced
echo inner not function scoped
```

```
$ ./wrapper.sh 
inner not function scoped
wrapperlogic
innermain
```

``
$ ./inner.sh 
inner not function scoped
```

**Comment by bparees** [comment]
so i think the issue you hit is that when you source the "inner" script, it contains an invocation of main itself.

File: 2/contrib/jenkins/install-jenkins-core-plugins.sh [hunk 3]
```diff
@@ -0,0 +1,20 @@
+#! /bin/bash -eu
+
+source /usr/local/bin/install-plugins-internals.sh
```
bparees: part of the solution to your issue might be to not source this unless you're in the rpms=false case.

gabemontero: I was about to suggest that ... we still want folks to be able to call inner.sh directly .. i.e. install-plugins.sh directly ... as they may have been doing that in their extension scenarios

gabemontero: assuming I can source within the logic like that and pass the parameters

bparees: https://stackoverflow.com/questions/39768712/how-to-pass-arguments-to-a-script-invoked-by-source-command

bparees: or you can just call the script rather than source it.

gabemontero: the latter is working ... will push that shortly

gabemontero:
update pushed @bparees ptal

gabemontero:
@adammhaile the jenkins PR for our ocp-build-data replace has merged ... please re-add it back

PR #814, opened 2019-03-11
Dockerfile.rhel7 cleanup for hybrid core/plugin install behavior on a…

gabemontero:
…pi.ci

/assign @bparees 

@smarterclayton fyi


gabemontero:
@bparees made a good point during devex arch ... specifically, the `FROM registry.access.redhat.com/openshift3/ose-cli` in Dockerfile.rhel7 should be changed to the new UBI equivalent

The current setting does not matter for api.ci per se (it is substituted there) but for building locally it appears to be prereq

For either this PR or a subequent one

gabemontero:
@bparees and I off line agreed changing the FROM and sorting through disabling repos is a subsequent PR

gabemontero:
ok tests look good

Issue #809, opened 2019-03-05
OpenShift 3.11: Build for centos: Base image in Dockerfile must be tagged to v3.11 


jomeier:
Hi,

```
FROM quay.io/openshift/origin-cli
```
does not work for builds for centos because the latest tag in quay.io is set to version 4.0. 

```
FROM quay.io/openshift/origin-cli:v3.11
```

does work.

Could you change that or mention in the documentation, please?

Best regards,

Josef

gabemontero:
hey @jomeier 

yeah, some big changes are happening as part of v4.0 of openshift 

among other things, the centos/rhel7 distinctions will be going away in large part

we won't be doing what you asked wrt switching to the v3.11, but PR https://github.com/openshift/jenkins/pull/819 includes README updates to start explaining things, as well as what if necessary if you want to build this repo locally (though that will only be supported for local dev now).

gabemontero:
I will be resolving this issue among other things with that PR

PR #819, opened 2019-03-13
get Dockerfile working for local buliding; more updates to README for v4

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/809

/assign @bparees 
(because of all the legacy impacts here)

I suspect the README updates minimally will need some massaging around UBI, but I think I at least found a valid URL for UBI in general.

But essentially the `Dockefile` 's are not for local dev only.

Also, as discussed previously @bparees , it probably makes sense to rename the `Dockerfile`'s to `Dockerfile.localdev` and `Dockerfile.rhel7` to `Dockerfile` but I have not taking that on just yet 
because 
a) that requires changes to openshift.release
b) doing it like this for now allowed me to leaving the Dockerfile.rhel7 allowed me to isolate what I could test locally vs. what I need tested in api.ci
c) I didn't want to go too far down the path if this was generally rejected


gabemontero:
yep was wondering about this when in api.ci and Dockerfile.rhel7

`/usr/local/bin/install-jenkins-core-plugins.sh: line 8: $2: unbound variable`

will need to tweak the optional passing in of the disabled repo list

gabemontero:
OK the image is building properly in api.ci now (meaning the local dev updates should be impacting the jenkins core/plugins install)

2/Dockerfile.rhel7
```diff
@@ -57,7 +57,7 @@ COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 ADD release.version /tmp/release.version
 
-RUN /usr/local/bin/install-jenkins-core-plugins.sh /opt/openshift/base-plugins.txt && \
+RUN /usr/local/bin/install-jenkins-core-plugins.sh /opt/openshift/base-plugins.txt " " && \
```
bparees: would be cleaner to make the script tolerate the absence of the second arg.

README.md
```diff
@@ -90,7 +87,27 @@ For example:
...
+Installation (OpenShift V4)
+---------------------------------
+
+    Starting with v4.0, the images are no longer hosted on DockerHub, but instead on Quay.  Their pull specs are:
+    * quay.io/openshift/origin-jenkins:v4.0
+    * quay.io/openshift/origin-jenkins-agent-nodejs:v4.0
+    * quay.io/openshift/origin-jenkins-agent-maven:v4.0
+    * quay.io/openshift/origin-jenkins-agent-base:v4.0
+
+    The images are also still available at the Red Hat Container Catlog.  Replace `openshift3` in the references above with `openshift4`.
```
bparees: catalog

and the container catalog images are(will be) ose-jenkins-*, not origin, right?

README.md
```diff
@@ -222,6 +241,8 @@ The complete list of plugins (i.e. including dependencies) needs to be provided
 Although this document will refrain on detailing the precise details, once the build on the internal Jenkins server is complete,
 the processes will be set in motion to build the `jenkins-2-plugins` RPM that is installed by the [RHEL7 Dockerfile](2/Dockerfile.rhel7) when the next version of the RHEL7 based OpenShift Jenkins image is built.  When new versions of OpenShift are released, associated versions of the RHEL7 based versions of the images produced by this repository are pushed to the Docker registry provided to RHEL7 subscribers.
 
+This plugin RPM is not for public consumption.  If a customer with a subscription wants to get a plugin update, they have to get an update to the OpenShift Jenkins image from the Red Hat Container Catalog.  They do not get access to a new version of the RPM and install it themselves via `yum` or `dnf`.
```
bparees: i'd leave this out, it belongs in the kbase article(that i guess is still pending) but it seems weird to be in our public/community readme.

2/Dockerfile.rhel7
bparees: (shift would probably get you there: https://www.tldp.org/LDP/Bash-Beginners-Guide/html/sect_09_07.html)

bparees:
i think leaving the file names alone for now makes sense, we want to keep this repo in sync w/ all our other repos and in all the other repos dockerfile.rhel7 is the file we care about for both OKD and OCP builds.

you could probably rename dockerfile to dockerfile.localdev (and not have a plain "dockerfile" file at all in the repo), that would be a nice way to confirm none of our CI is dependent on it.

README.md
gabemontero: Yeah this stemmed from lack of optimism that the kbase article won't happen any time soon

I don't know if it is weird to have this in the README, but don't feel strongly about it.

I can remove it.

README.md
gabemontero: I thought the ose-* ones would only show up in the TBR.  That said, I should list those as well.

Will make some updates in the next push.

gabemontero:
pushed 2 new commits @bparees 
- responses to your in line comments
- move of Dockerfile to Dockerfile.localdev

ptal

gabemontero:
unrelated flakes in e2e-aws

README.md
bparees: the TBR is the redhat container catalog.  I assumed that's what you intended to refer to.

The available locations are/will be:

quay.io/openshift/origin-jenkins:v4.0  (community)
registry.redhat.io/openshift4/ose-jenknis:v4.0 (ocp) (I think).

though since they're going to have the same UBI base image i'm not even sure what meaningful difference there will be.

I think maybe we are talking about different things though.  Might be better to just specify the actual urls so there's no room for confusion.

bparees: (also there's still a typo in Catalog above)

gabemontero:
next round of updates pushed @bparees ptal

README.md
```diff
@@ -96,10 +96,15 @@ Installation (OpenShift V4)
...
 
-    The images are also still available at the Red Hat Container Catlog.  Replace `openshift3` in the references above with `openshift4`.
+    The images are also still available at the Red Hat Container Catalog for customers with subscriptions, 
+    though with some changes in the naming.
+    
     Also, the `registry.access.redhat.com` host is being transitioned to `registry.redhat.io`.  Both exist at this time, but eventually
-    only `registry.redhat.io` will remain.  See [Transitioning the Red Hat container registry](https://www.redhat.com/en/blog/transitioning-red-hat-container-registry)
-    for details.  Lastly, replace `origin` in the quay.io references above with `ose`.
+    only `registry.redhat.io` will remain.  See [Transitioning the Red Hat container registry](https://www.redhat.com/en/blog/transitioning-red-hat-container-registry) for details:
+    * registry.redhat.io/openshift4/ose-jenknis:v4.0
```
bparees: typo

README.md
```diff
@@ -96,10 +96,15 @@ Installation (OpenShift V4)
...
+    * registry.redhat.io/openshift4/ose-agent-nodejs:v4.0
+    * registry.redhat.io/openshift4/ose-agent-maven:v4.0
+    * registry.redhat.io/openshift4/ose-agent-base:v4.0
```
bparees: dont these have jenkins in the name?

gabemontero: ahhh ... yes they do

README.md
gabemontero: part of next push

gabemontero:
next round pushed @bparees

bparees:
looks right to me... you may get a few confused people trying to pull the new images from registry.redhat.io before 4 GAs though.

gabemontero:
thanks ... I'll risk the pre-GA confusion and adjust if need be

will squash then tag for merge

PR #817, opened 2019-03-11
Add adambkaplan as approver

adambkaplan:
/assign @gabemontero

PR #821, opened 2019-03-19
fix up owners files

bparees:


adambkaplan:
Covered in https://github.com/openshift/jenkins/pull/817

Issue #823, opened 2019-03-22
default agents using registry.redhat.io


Gl4di4torRr:
I just deployed the latest `v3.11` version of Jenkins on OpenShift and ran a quick pipeline for testing. Looks like the default agents are now using [registry.redhat.io](https://github.com/openshift/jenkins/blob/4fa776906048e2fda7b57b1f93af9d3fb05ab895/2/contrib/jenkins/kube-slave-common.sh#L28). Now my agents are failing to spin up because this requires a login to the container catalog. Looking at the [docs](https://github.com/openshift/jenkins/tree/openshift-3.11#installation), it seems like they should be pointing to `access.redhat.com`. Am I missing something or could this be a gap in the docs vs agents? Thanks!

Gl4di4torRr:
I found https://access.redhat.com/RegistryAuthentication but doesn't this break the out-of-the-box functionality of jenkins? Maybe we should update the docs to use `registry.redhat.io`?

gabemontero:
Yeah we recently updated the readme for this ( see https://github.com/openshift/cluster-samples-operator/blob/master/pkg/apis/samples/v1/types.go#L403-L405 ), but only in the 4.0 context there.

But the quick FYI is registry.redhat.access.com is going away in the pretty near future.

I honestly had totally forgot about https://github.com/openshift/jenkins/pull/665 from last summer

@bparees - I forget now, were there concerns about folks needing creds when we executed https://github.com/openshift/jenkins/pull/665?  ... or did we just miss updating the docs to explain that?

Also, switching that ref to quay.io was not considered since it wasn't available yet and the jenkins based UBI available yet (though that stuff of course is not available for 3.x, just 4.x).... running the UBI with subscriptions/entitlements *may* provide some wiggle room (though I am not optimistic).

Minimally @Gl4di4torRr , yes, some more hits to the doc seem warranted, unless some @bparees and I sort out some kind of accommodation for 3.x re: the UBI and which image we reference in the config for the sample agent defs.

@openshift/sig-developer-experience @waveywaves fyi

bparees:
> But the quick FYI is registry.redhat.access.com is going away in the pretty near future

registry.redhat.access.com will probably be around for a long time but it will not get rhel8 based images.

> @bparees - I forget now, were there concerns about folks needing creds when we executed #665? ... or did we just miss updating the docs to explain that?

[...]

PR #827, opened 2019-03-27
adjust agent config based on cluster version; augment readme around r…

gabemontero:
…hel registry and security advisories

Fixes https://github.com/openshift/jenkins/issues/823

/assign @bparees 

So the *"parsing"* of the `oc version` output is not a clean as I was hoping for given the change in behavior between 3.10 and 3.11, but I would think `oc` is relatively stable at this point in the 3.x stream.

And relative to the pain I'm anticipating as more users apply the 3.11 image against older clusters for the security advisory pickup.

I've tested against 3.6 through 3.11 clusters.

### DIFF
--- a/2/contrib/jenkins/kube-slave-common.sh
+++ b/2/contrib/jenkins/kube-slave-common.sh
@@ -31,6 +31,25 @@ if [[ `grep CentOS /etc/redhat-release` ]]; then
   NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-openshift/jenkins-agent-nodejs-8-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
   MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-openshift/jenkins-agent-maven-35-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
 fi
+# if the master is running the rhel image, but we are on a pre-3.11 cluster, use the non-TBR images.
+if [[ `grep Enterprise /etc/redhat-release` ]]; then
+  versions=`oc version`
+  echo "OpenShift client and server versions are ${versions}"
+  countForInPodCheck=`echo $versions | awk -F"v3" '{print NF-1}'`
+  # if 2 instance of 'v3', then we are running in a pre3.11 pod;
+  # in the 3.11 image `oc`, the server version reporting only lists kubernetes 1.11
+  if [ "${countForInPodCheck}" -eq "2" ]; then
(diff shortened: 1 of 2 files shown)


2/contrib/jenkins/kube-slave-common.sh
```diff
@@ -31,6 +31,18 @@ if [[ `grep CentOS /etc/redhat-release` ]]; then
   NODEJS_SLAVE=${NODEJS_SLAVE_IMAGE:-openshift/jenkins-agent-nodejs-8-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
   MAVEN_SLAVE=${MAVEN_SLAVE_IMAGE:-openshift/jenkins-agent-maven-35-centos7:${JENKINS_SLAVE_IMAGE_TAG}}
 fi
+# if the master is running the rhel image, but we are on a pre-3.11 cluster, use the non-TBR images.
+if [[ `grep Enterprise /etc/redhat-release` ]]; then
+  versions=`oc version`
+  echo "OpenShift client and server versions are ${versions}"
+  countForInPodCheck=`echo $versions | awk -F"v3" '{print NF-1}'`
+  # if 2 instance of 'v3', then we are running in a pre3.11 pod;
+  # in 3.11, the server version reporting only lists kubernetes 1.11
```
bparees: I'm seeing something different, though it may be a client issue:

```
$ oc version
oc v3.11.0+0c1c28f-124
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://api.ci.openshift.org:443
openshift v3.11.0+84fe0ad-23
kubernetes v1.11.0+d4cacc0
```

so i do see v3 twice in the output.

bparees: might be better to look for "openshift v3" and then check if it's "openshift v3.11" or not.

gabemontero: yeah my tests were with the `oc` in the image from within a jenkins pod ... is that what you did?  

that said, I debated the extra tests like you noted vs. KISS and initially sided with the latter

I'll play with the extra verifications and see what it looks like

gabemontero: also, I *think* it looked different from within a pod when I `oc rsh ...` in  vs. what came out of the jenkins start up script 

I'll double check that as well and post the results

bparees: i built an oc client binary from origin/release-3.11.

gabemontero: ok ... so yeah, maybe *eventually* a 3.11 refresh of the rhel image will pick up a later version of `oc` and it will have the openshift v3 line for the server as well.

I just brought up the 3.11 rhel image and `oc` is at `3.11.88`.  And only one `v3.11` line.

gabemontero:
ok ptal @bparees .... believe I account for both `3.11` output patterns now

PR #830, opened 2019-03-28
some more '--setopt=protected_multilib=false ' now needed for the pus…

gabemontero:
…h jenkins image job

got by with updates to 2/ slave-base/ agent-maven-35/ last time ...need slave-maven/ now

/assign @bparees

### DIFF
--- a/slave-maven/Dockerfile
+++ b/slave-maven/Dockerfile
@@ -13,7 +13,7 @@ ENV MAVEN_VERSION=3.3 \
 RUN INSTALL_PKGS="java-1.8.0-openjdk-devel.x86_64 rh-maven33*" && \
     x86_EXTRA_RPMS=$(if [ "$(uname -m)" == "x86_64" ]; then echo -n java-1.8.0-openjdk-devel.i686 ; fi) && \
     yum install -y centos-release-scl-rh && \
-    yum install -y --enablerepo=centosplus $INSTALL_PKGS $x86_EXTRA_RPMS && \
+    yum install -y --setopt=protected_multilib=false --enablerepo=centosplus $INSTALL_PKGS $x86_EXTRA_RPMS && \
     curl -LOk https://services.gradle.org/distributions/gradle-${GRADLE_VERSION}-bin.zip && \
     unzip gradle-${GRADLE_VERSION}-bin.zip -d /opt && \
     rm -f gradle-${GRADLE_VERSION}-bin.zip && \


agent-nodejs-8/Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY contrib/bin/configure-agent /usr/local/bin/configure-agent
 RUN yum install -y centos-release-scl-rh && \
     INSTALL_PKGS="rh-nodejs${NODEJS_VERSION} rh-nodejs${NODEJS_VERSION}-npm rh-nodejs${NODEJS_VERSION}-nodejs-nodemon make gcc-c++" && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    yum install -y --setopt=protected_multilib=false --setopt=tsflags=nodocs $INSTALL_PKGS && \
```
bparees: why does nodejs need this? it doesn't install javac right?

slave-nodejs/Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY contrib/bin/scl_enable /usr/local/bin/scl_enable
 RUN yum install -y centos-release-scl-rh && \
     INSTALL_PKGS="rh-nodejs4 rh-nodejs4-npm rh-nodejs4-nodejs-nodemon" && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    yum install -y --setopt=protected_multilib=false --setopt=tsflags=nodocs $INSTALL_PKGS && \
```
bparees: not needed for nodejs?

agent-nodejs-8/Dockerfile
gabemontero: tried to explain in the description, apologies for lack of clarity ... this was a preemptive decision in case something else popped up in the future ...

but if we should wait until the instances pop up, I'll remove

bparees: you caught me not reading pr descriptions again. :)

i'd be inclined to wait in case there are other decisions that should be made when we get to that point.

bparees:
> needs agent-maven too right?

read your description now, sounds like it was done previously.

gabemontero:
nodejs removed @bparees ptal

bparees:
/lgtm

update the pr summary though since it still mentions nodejs

gabemontero:
description updated thanks @bparees

PR #834, opened 2019-04-12
run yum update to avoid package mismatch between archs

jeffdyoung:
Run yum update to avoid package version mismatch between archs


jeffdyoung:
@gabemontero

PR #862, opened 2019-05-30
add ubi7 based Dockerfile

mbach04:


gabemontero:
Hey @mbach04 

So initial comments:
- with what you have done, the `oc` binary is not included in the image
Con with that:  it makes it decidedly different from our images that get pushed to quay.io/registry.redhat.io
Pro with that: we have had reasons to regret including the `oc` binary in our image ... perhaps this could be a back door gateway around getting some traction for moving off of that :-)

- supposedly the `quay.io/openshift/origin-cli`image used in our https://github.com/openshift/jenkins/blob/master/2/Dockerfile.localdev file is also UBI based .... were you aware of that?  If so, can you clarify the rationale for wanting this new form of a UBI based image vs. the existing one?

Maybe related to the various repo enabled/disabled in each form ... i.e. https://github.com/openshift/jenkins/blob/master/2/Dockerfile.localdev#L39

mbach04:
Interesting. I pushed this version up b/c the requirement on this project is to layer it on top of a modified ubi7 base image. But it's close enough that there's no impact on testing with the ubi7 hosted directly by Red Hat. I wasn't aware the origin-cli image was based on ubi either. I'm still working through all of this to get a self hosted version for all binaries and such, but this is good to know as I'll need to layer in the oc binary as well...assuming I can find where it's normally located.

gabemontero:
An old example of how we used to install the `oc` binary vs. pick it up in the image we extended from is https://github.com/openshift/jenkins/blob/openshift-3.6/2/Dockerfile#L44-L47

The relase of the binary downloaded would have to be updated of course.

gabemontero:
@bparees @adambkaplan fyi ^^^

gabemontero:
@mbach04 the technical head honcho for the openshift project is @smarterclayton  .... if you have not already, or if @bparees did not after he pointed you to me, it might be good to pick his brain on some of the internal details of openshift v4.x and your reqs.

@smarterclayton - @mbach04 is a RH'er heading a team focused on a particular customer vertical ... I'll forward you an email as a means of introduction.

pabrahamsson:
> Hey @mbach04
> 
> So initial comments:
> 
> * with what you have done, the `oc` binary is not included in the image
>   Con with that:  it makes it decidedly different from our images that get pushed to quay.io/registry.redhat.io
>   Pro with that: we have had reasons to regret including the `oc` binary in our image ... perhaps this could be a back door gateway around getting some traction for moving off of that :-)
> * supposedly the `quay.io/openshift/origin-cli`image used in our https://github.com/openshift/jenkins/blob/master/2/Dockerfile.localdev file is also UBI based .... were you aware of that?  If so, can you clarify the rationale for wanting this new form of a UBI based image vs. the existing one?
> 
> Maybe related to the various repo enabled/disabled in each form ... i.e. https://github.com/openshift/jenkins/blob/master/2/Dockerfile.localdev#L39

Hey @gabemontero,
Would you be able to confirm that `quay.io/openshift/origin-cli` image is indeed UBI based? Not sure how I can tell but I can see it being used for the slave-base image as well which I'm looking to use to update some of the [jenkins agents that we have in the Containers & PaaS CoP](https://github.com/redhat-cop/containers-quickstarts/issues/226) to be based on UBI (preferably 8 but 7 will be ok as well).

bparees:
> Would you be able to confirm that quay.io/openshift/origin-cli image is indeed UBI based? Not sure how I can tell


$ docker inspect quay.io/openshift/origin-cli | grep ubi
                "com.redhat.component": "ubi7-container",
                "name": "ubi7",
                "url": "https://access.redhat.com/containers/#/registry.access.redhat.com/ubi7/images/7.6-123",
                "com.redhat.component": "ubi7-container",
                "name": "ubi7",
                "url": "https://access.redhat.com/containers/#/registry.access.redhat.com/ubi7/images/7.6-123",

akram:
Hi @mbach04 
thank you for submitting this PR. As you see, current conclusion is that the current image is already UBI based.
Please feel free to post any other PR that would be adding other requirements on top of the existing image.

Also, FYI an update is pending for the Jenkins image to use OpenJDK 11 instead of Java 8.

PR #896, opened 2019-07-23
remove all yum-config-manager invocations from Dockerfiles

bparees:
let's see what breaks. In theory nothing should since yum-config-manager
is no-op'ing anyway.

OSBS strips these yum-config-manager invocations and on the origin/CI side of the house, they are being no-op'd, so they should have no effect there either.

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -40,12 +40,11 @@ LABEL com.redhat.component="openshift-jenkins-2-container" \
 # 8080 for main web interface, 50000 for slave agents
 EXPOSE 8080 50000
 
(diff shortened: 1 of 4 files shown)


waveywaves:
@bparees noted

bparees:
CI infra AWS issues.
/retest

bparees:
confirmed this is safe for OSBS, so removing [WIP].  if it passes CI i assume we're good to go.

bparees:
@gabemontero @akram same issue here (jenkins e2es are legitimately failing)

bparees:
/hold

Please unhold when the e2e is fixed.


Ben Parees | OpenShift

PR #899, opened 2019-07-25
Prometheus Monitoring Integration

waveywaves:
## Description
_Enable users to tract the resource utilization by Jenkins by integrating with Openshift Monitoring._

### Architecture
https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md#related-resources
https://github.com/coreos/prometheus-operator/blob/master/Documentation/troubleshooting.md#overview-of-servicemonitor-tagging-and-related-elements

## Checkpoint
- [x] Add metrics-plugin
- [x] Add prometheus-plugin
- [x] Let `ServiceMonitor` CR discover the Jenkins `svc` which would allow connection to `/prometheus` endpoint for data ingestion
- [ ] Use Prometheus + Grafana to display dashboard
- [ ] Add Authentication via https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/basic-auth.md and https://github.com/jenkinsci/prometheus-plugin/issues/23

cc @akram @gabemontero

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -91,3 +91,7 @@ blueocean:1.10.2
 # 2.5 now includes pipeline-model-definition (declaritive pipeline)
 # 2.4 brought in pipeline-milestone-step
 workflow-aggregator:2.6
+
+# Monitoring plugins
+metrics:4.0.2.5
+prometheus:2.0.0
--- a/openshift/templates/jenkins-ephemeral-monitored.json
+++ b/openshift/templates/jenkins-ephemeral-monitored.json
@@ -1,0 +1,339 @@
+{
+  "kind": "Template",
+  "apiVersion": "v1",
+  "metadata": {
+    "name": "jenkins-ephemeral-monitored",
+    "annotations": {
+      "openshift.io/display-name": "Jenkins (Ephemeral)",
+      "description": "Jenkins service, without persistent storage.\n\nWARNING: Any data stored will be lost upon pod destruction. Only use this template for testing.",
+      "iconClass": "icon-jenkins",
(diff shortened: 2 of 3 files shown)


openshift/templates/jenkins-ephemeral-monitored.json
```diff
@@ -0,0 +1,339 @@
+{
+  "kind": "Template",
```
akram: We need to put these templates in openshift/origin .
@gabemontero does origin still accept new templates?

2/contrib/openshift/base-plugins.txt
```diff
@@ -91,3 +91,7 @@ blueocean:1.10.2
 # 2.5 now includes pipeline-model-definition (declaritive pipeline)
 # 2.4 brought in pipeline-milestone-step
 workflow-aggregator:2.6
+
+# Monitoring plugins
```
akram: good

openshift/templates/jenkins-ephemeral-monitored.json
gabemontero: it does, but as we commented in the openshift/origin PR, you add the template here in this repo, and then update https://github.com/openshift/library 's jenkins artifacts to reference your new template here once this PR merges.

Eventually, as part of 4.3, I'll pull in your new template into the samples operator.

the template refs in openshift/origin are pretty much historical at this point ... all the "samples" repos and the openshift/library content are the "source of truth"

akram: ok @gabemontero , we will do that then.

akram:
As discussed with @gabemontero and @waveywaves , we will put the templates here now so they can be handled by the samples operator.

/LGTM

PR #901, opened 2019-08-02
Bug 1729522: Bump openshift-sync-plugin to v1.0.40

akram:


gabemontero:
@enj @adambkaplan fyi this should pull in the fabric client change where they user `/apis` vs. root path traversal to determine if they are running on openshift or not 

After this merges, the jenkins image on api.ci should get updated.

@akram will then make the additional updates for brew/OSBS/RPMs and what QA will verify

gabemontero:
@akarm you either say just `/retest` or you say `/test e2e-aws-jenkins`

PR #908, opened 2019-08-11
Bump plugin versions wrt workflow-cps fix in client plugin

waveywaves:
Updates: 

Wrt the fix in https://github.com/openshift/jenkins-client-plugin/pull/287 workflow-cps plugin versions above 2.71 are now usable as they do not give the warning it used to before which was due to a bug in the jenkins client plugin itself which was also solved in the PR above.

Plugins updating:
workflow-cps FROM 2.65 -> 2.73
jenkins-client-plugin FROM 2.65 -> 2.73

/assign @akram 
/assign @gabemontero

### DIFF
--- a/2/contrib/openshift/base-plugins.txt
+++ b/2/contrib/openshift/base-plugins.txt
@@ -60,7 +60,7 @@ mercurial:2.3
 subversion:2.10.3
 github:1.29.2
 github-branch-source:2.3.6
-workflow-cps:2.65
+workflow-cps:2.73
 workflow-cps-global-lib:2.15
 pipeline-model-definition:1.3.4.1
 token-macro:2.8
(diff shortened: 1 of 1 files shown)


akram:
Hi @waveywaves 
could you also add the new version of the openshift-sync plugin in this same PR ?
openshift-sync => 1.0.39

waveywaves:
@Akram sync plugin is at 1.0.40 https://github.com/openshift/jenkins/blob/0af9e14c4b7aff4b9652d813a65d8a206a2766d6/2/contrib/openshift/base-plugins.txt#L12

akram:
> @akram sync plugin is at 1.0.40
> 
> https://github.com/openshift/jenkins/blob/0af9e14c4b7aff4b9652d813a65d8a206a2766d6/2/contrib/openshift/base-plugins.txt#L12

yep, sorry, I had in mind the new version of the sync plugin, but there is still a PR pending before we can cut the new release that would be 1.0.41 actually.

PR #939, opened 2019-10-03
Add Nodejs 10 Agent

waveywaves:


gabemontero:
yeah same as  4.1 ... not sure backporting a new feature like this to 3.11 is such a good idea

it hasn't even soaked in 4.3/master yet

is there some background here?

/hold

waveywaves:
@gabemontero Nodejs 8 goes EOL end of this year, that's the only reason we are introducing this.

gabemontero:
ok that is what I expected @waveywaves 

but the established pattern is to introduce the new nodejs/maven in our newer releases and eventually remove the EOL ones

we explicitly did NOT backport newer nodejs/maven images to older openshift releases

for example, if you look at https://github.com/openshift/jenkins/tree/openshift-3.6, it doesn't even have nodejs8.... it has the old/original 4.4 version of nodejs

There are several concerns around such backports, which we would consider features.  QE coverage for example.

I would advise maintaining this pattern.  But perhaps a conversation between us, @akram , @siamaksade , is warranted.

PR #946, opened 2019-10-07
Allow use of podman or buildah by using the BUILD_COMMAND variable

akram:
This is required to allow: 
- build with RHEL8 (which has not docker anymore)
- build multistage to get rid of dumb-init

/assign @waveywaves
/assign @gabemontero

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ subscribed RHEL machine.
     $ make build TARGET=rhel7 VERSION=2
     ```
 
+You can also run the build using Podman. However, e2e tests does not run  
+with Podman (which is the default).
+    ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,6 +56,12 @@ subscribed RHEL machine.
     $ make build TARGET=rhel7 VERSION=2
     ```
 
+You can also run the build using Podman. However, e2e tests does not run  
+with Podman (which is the default).
+    ```
+    $ make build TARGET=rhel7 VERSION=2 BUILT_COMMAND="podman build"
```
gabemontero: looks like it is `BUILD_COMMAND` below, not `BUILT_COMMAND`

akram: good catch, I will push it in joint PR thanks @gabemontero

Issue #659, opened 2018-07-19
Add support of  HTTP_PROXY_HOST  to the maven agent image


akram:
Do the same to  #264 but for proxies.

The following variables have to be managed

HTTP_PROXY_HOST and $HTTP_PROXY_PORT environment variables. Optionally, you can also set the $HTTP_PROXY_USERNAME, HTTP_PROXY_PASSWORD, and HTTP_PROXY_NONPROXYHOSTS variables.

akram:
should be fixed by #660

bparees:
reverted https://github.com/openshift/jenkins/pull/660 to further discuss the proxy var we want to use.

gabemontero:
See https://github.com/openshift/jenkins/pull/660#issuecomment-416244946 for @bparees 's latest analysis wrt copying what was done with the s2i wildfly image here

akram:
closed in favour of #949

/close

PR #949, opened 2019-10-08
Manage proxy configuration for the maven agent 

akram:
Manage proxy configuration for the maven agent using http_proxy, https_proxy and no_proxy


akram:
Fixes https://github.com/openshift/jenkins/issues/659

akram:
Per discussions on https://github.com/openshift/jenkins/issues/659 , and to follow wildfly team conventions we decided to go http_proxy (in lower case).
However, discussion should be re-opened, because, all our env vars are upper case.

PR #952, opened 2019-10-09
Install jq on slave base image only

akram:


akram:
some server flake again :-( 
/retest

PR #958, opened 2019-10-17
Upgrade some dependencies to allow use of earlier kubernets clien api…

akram:
… on plugins


akram:
As of we found a bug on our CI, it seems that this one was not needed, because, we did a previous one which was sufficient.

/close

PR #961, opened 2019-10-23
Set env LC_ALL and LANG to en_US.UTF-8 in slave images

### LINKED ISSUE #960: Set slave locale to UTF-8 too
In #279 the locale for the jenkins master was set, but slaves still use ASCII as default, this can lead to errors when using e.g. the `unstash` step with non-ASCII filenames

akram:
Fixes #960


akram:
/retest

flakes with Terraform failing on AWS

PR #973, opened 2019-11-04
Bug 1764460: Re-ordering some plugins in alphabetical order

akram:
This is a fake fix to allo the tracker for  bz-1764460 to be updated


akram:
some big flakes with AWS volumes and pssibily preventing router metrics to be available 

/retest

PR #978, opened 2019-11-13
Bump jenkins to version 2.190.3

akram:


akram:
despite my manual tests with this version didn't show issues: all the plugins I tested works correctly, it seems however that our e2e tests here are failing.

I will put it on 
/hold
/retitle WIP Bump jenkins to version 2.190.2

PR #979, opened 2019-11-13
java-1.8 on centos

akram:
Revert to 1.8 for alternatives was missing centos


akram:
/hold 

this one looks to be blocked by ruby-22 e2e tests deprecation.
We are keepign an eye on it.

PR #982, opened 2019-11-18
Bug 1764466: 4.1 bz 1764466 cve 2019 10431

akram:


akram:
As ruby-22 e2e tests are failing because of the image deprecation, I am overriding these tests
The other tests are passing.

PR #983, opened 2019-11-19
Bug 1771337: Bump openshfit-login-plugin to 1.0.21

akram:
Bump openshfit-login-plugin to 1.0.21


akram:
We are still waiting for the jenkinsci team to fix updates.jenkins-ci.org
it does not sync properly new versions of the plugin

PR #1007, opened 2020-01-09
Disable rhel-fast-datapath-beta repo

akram:
Disable rhel-fast-datapath-beta repo as it has silently be added to origin-cli image


akram:
was flake
/test e2e-aws

PR #1016, opened 2020-01-24
Use OpenJDK 11 on s390x [4.2]

yselkowitz:
OpenJDK 8 does not have JIT enabled on s390x, and Jenkins has moved
unconditionally to OpenJDK 11 in 4.3.

### DIFF
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -46,7 +46,8 @@ EXPOSE 8080 50000
 # /usr/lib64/jenkins will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images 
 # do *NOT* have a /usr/lib64/jenkins path
 RUN ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
-    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk java-1.8.0-openjdk-devel" && \
+    JDK_VER=$(if [ "$(uname -m)" == "s390x" ]; then echo -n 11; else echo -n 1.8.0; fi) && \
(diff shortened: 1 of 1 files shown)


akram:
Hi @yselkowitz  , 
Please have a look at master branch  here https://github.com/openshift/jenkins/blob/master/2/Dockerfile.rhel7 and https://github.com/openshift/jenkins/blob/master/2/Dockerfile.localdev to see the implementation we are using.
For a transition period, we will embed the 2 JDK versions in the image and set the openjdk-11 as default.
Could you please change your PR to rely on the same principle?

2/Dockerfile.rhel7
```diff
@@ -46,7 +46,8 @@ EXPOSE 8080 50000
 # /usr/lib64/jenkins will subsequently get redirected to /usr/lib/jenkins; it is confirmed that the 3.7 jenkins RHEL images 
 # do *NOT* have a /usr/lib64/jenkins path
 RUN ln -s /usr/lib/jenkins /usr/lib64/jenkins && \
-    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-1.8.0-openjdk java-1.8.0-openjdk-devel" && \
+    JDK_VER=$(if [ "$(uname -m)" == "s390x" ]; then echo -n 11; else echo -n 1.8.0; fi) && \
```
akram: see https://github.com/openshift/jenkins/blob/master/2/Dockerfile.rhel7 to calc the implementation.

2/Dockerfile.rhel7
```diff
@@ -46,7 +46,8 @@ EXPOSE 8080 50000
...
+    INSTALL_PKGS="dejavu-sans-fonts wget rsync gettext git tar zip unzip openssl bzip2 dumb-init  java-$JDK_VER-openjdk java-$JDK_VER-openjdk-devel" && \
```
akram: could you use non abbreviated variable names s/JDK_VER/JDK_VERSION/g

yselkowitz:
At this point, I think we'll just leave this, as it's already in 4.3.

PR #1019, opened 2020-01-31
Bug 1971537: Support cgroups v2

cardil:
In code there are embedded paths that are valid only for cgroups v1. This PR creates functions that calculate memory in container either for v1 or v2 of cgroups.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -238,8 +238,33 @@ function get_java_proxy_config() {
   echo $config
 }
 
+function linux_available_memory_bytes() {
+  local kbytes
+  kbytes=$(cat /proc/meminfo | grep MemAvailable | awk '{print $2}')
+  echo $((kbytes*(2**10)))
+}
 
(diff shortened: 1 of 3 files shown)


waveywaves:
@cardil Thank you for the contribution. As you have mentioned that we support cgroupsv1 only, based on the same I will review this for v2 support.
cc @akram

olivergondza:
@waveywaves, just to clarify I understand, do you mean this will be postponed and addressed as a part of some larger effort later? I am asking to set some expectation on when this can get integrated. Thanks!

akram:
thanks @cardil for this contribution. We need more time to test it ; as it seem non-intrusive for OpenShift, it could be merged hopefully.
Please not that for now we don't support cgroups v2 in OpenShift.

akram:
/retitle [main] Support cgroups v2 (Podman on Fedora 31)

akram:
/retitle [head] Support cgroups v2 (Podman on Fedora 31)

PR #1021, opened 2020-02-04
Bug 1793322: [4.3] Remove use of /etc/passwd and upgrade plugins versions

akram:
Related to critical CVE

Makes use of /etc/passwd g=rw not required
Updates Jenkins version
Updates Jenkins plugins


akram:
https://updates.jenkins.io/ starts to be flaky
/retest

PR #1023, opened 2020-02-04
Bug 1793320: Remove use of /etc/passwd in favor of cri-o and upgrade plugins versions [4.1]

akram:
* remove setting /etc/passwd to g=rw in favor of cri-o
* updates jenkins to 2.204.1
* updates plugins versions

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -1,0 +1,1 @@
+Dockerfile.localdev
--- a/2/Dockerfile.localdev
+++ b/2/Dockerfile.localdev
@@ -18,7 +18,7 @@ ENV JENKINS_VERSION=2 \
     HOME=/var/lib/jenkins \
     JENKINS_HOME=/var/lib/jenkins \
     JENKINS_UC=https://updates.jenkins.io \
-    OPENSHIFT_JENKINS_IMAGE_VERSION=4.0 \
+    OPENSHIFT_JENKINS_IMAGE_VERSION=4.1 \
     LANG=en_US.UTF-8 \
     LC_ALL=en_US.UTF-8 \
     INSTALL_JENKINS_VIA_RPMS=false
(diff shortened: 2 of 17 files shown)


gabemontero:
@akram I thought we confirmed that cri-o was doing its thing in 4.1 as well

akram:
@gabemontero , IIRC on my verification it is from 4.2 only . can you double check?

gabemontero:
@akram I found the forum-containers thread where @nalind said the cri-o change was backported to k8s 1.13 and 4.1

`looks like 1.14 (for kube 1.14, so openshift 4.2, i think) as PR#2022, backported to 1.13 (4.1) in PR#2302`

I am not posting the slack thread link as that thread had sensitive information.

Since the jenkins image is coming from the payload anyway, the latest 4.1.z, which would pick this up, would have the appropriate cri-o change

and our e2e's with you doing the same removal would prove it as well.

So go with the depend on crio change first, and if the e2e proves otherwise, we can fall back to an nss_wrapper related change.

akram:
https://updates.jenkins.io/ starts to be flaky
/retest

gabemontero:
passing tests @akram !!  Congratulations :-)

you should change the PR title to reflect the dependence on cri-o, etc.

akram:
@gabemontero yes! finally, we did it.... after corn flakes, snowflakes, aws flakes, I can proudly say that I eat also jenkins infra flakes now.

The 4.4 payload has been built by ART yesterday, and hopefully the dependent BZ can be set to VERIFIED soon.

mfojtik:
@akram @gabemontero the release-4.1 is nearly EOL and only critical fixes are accepted now. Do you consider this as critical fix?

gabemontero:
@akram @mfojtik and I discussed offline and we are closing this given 4.1 EOL.

getting addressed in 4.2 and 4.3

PR #1024, opened 2020-02-04
Bug 1793324: Use nss_wrapper [3.11]

akram:
Revert to use of nss_wrapper and update some plugins


akram:
needs https://github.com/openshift/origin/pull/24548

PR #1033, opened 2020-03-09
Bug 1810393: 4.1  2.204.2 and plugins updates

akram:
This is a catch-up PR as we decided to keep 4.1 because of deprecation.
It adds:
- use of go-init instead of dumb-init that is not resolvable in epel
- use of Jenkins 2.204.2
- use of latest plugins

### DIFF
--- a/2/Dockerfile
+++ b/2/Dockerfile
@@ -1,0 +1,1 @@
+Dockerfile.localdev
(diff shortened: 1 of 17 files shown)


akram:
squashed the 3 commits into 1 as I can't figure out why the e2e-aws-jenkins test is failing.

PR #1039, opened 2020-03-24
Revert to wildcard install of Jenkins 2 rpm

akram:


waveywaves:
@akram why wildcard install ?

PR #1043, opened 2020-04-02
Support Nodejs 12 Agent

waveywaves:
- Add Support for Nodejs 12 Agent for Jenkins

### DIFF
--- a/agent-nodejs-12/Dockerfile.localdev
+++ b/agent-nodejs-12/Dockerfile.localdev
@@ -1,0 +1,32 @@
+FROM quay.io/openshift/origin-jenkins-agent-base:v4.0
+
+MAINTAINER Akram Ben Aissi <abenaiss@redhat.com>
+
+ENV NODEJS_VERSION=12 \
+    NPM_CONFIG_PREFIX=$HOME/.npm-global \
+    PATH=$HOME/node_modules/.bin/:$HOME/.npm-global/bin/:$PATH \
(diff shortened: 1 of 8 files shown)


agent-nodejs-12/Dockerfile.localdev
```diff
@@ -0,0 +1,32 @@
+FROM quay.io/openshift/origin-jenkins-agent-base:v4.0
+
+MAINTAINER Akram Ben Aissi <abenaiss@redhat.com>
```
akram, line 3: I would like to change all these by:
MAINTAINER OpenShift Developer Services <openshift-dev-services+jenkins@redhat.com>

can we start with with this one?

agent-nodejs-12/test/agent_nodejs_test.go
```diff
@@ -0,0 +1,116 @@
...
+var imageName string
+
+var _ = BeforeSuite(func() {
+	var err error
+	dockercli, err = docker.NewEnvClient()
+	Expect(err).NotTo(HaveOccurred())
+
+	imageName = os.Getenv("IMAGE_NAME")
+	if imageName == "" {
+		imageName = "vbobade/openshift-jenkins-base-agent"
```
akram, line 29: it is only visible in the e2e test right?

nielsbril:
When will this image be available on Quay.io?

akram:
/retitle Bug 1848284: Support Nodejs 12 Agent

PR #1045, opened 2020-04-09
Bug 1804345: handle ingress certificate signed by custom ca

jkhelil:


2/contrib/s2i/run
```diff
@@ -268,6 +268,11 @@ if [[ -z "${JAVA_TOOL_OPTIONS}" ]]; then
   export JAVA_TOOL_OPTIONS
 fi
 
+# update system java keystore with custom ca bundle coming from jenkins-trust-ca-bundle configmap
+if [ -s "/etc/pki/ca-trust/source/anchors" ]; then
```
akram, line 272: /etc/pki/ca-trust/source/anchors is supposed to be a directory.
the -s test work with file and test zero size , then we need another test

jkhelil: This a file where the ca trust configmap is mounted, and yes i am testing it exist and more than zero bytes.
You are may referecing the /etc/pki/ca-trust/source which is the directory for system wide certificates for a rhel os where it looks for certs https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-shared-system-certificates

jkhelil: fixed

akram:
/retitle Bug 1817436: handle ingress certificate signed by custom ca

akram:
/retitle Bug 1804345: handle ingress certificate signed by custom ca

PR #1046, opened 2020-04-14
Bug 1820677: Update Description on Monitored Templates reflecting permission issues

waveywaves:
To create the Service Monitor custom resource in the Templates,
provided by the Prometheus Operator in Openshift,
users needs specific RBAC permissions to be able to create them.
This necessity needs to be understood by them, hence adding this to the
description of the templates.


waveywaves:
@adambkaplan I will update accordingly then. Thanks for helping out on this.

akram:
I updated the documentation here:
https://github.com/openshift/openshift-docs/pull/21173

we cannot directly reference the KCS article as we use also the templates for origin.

PR #1051, opened 2020-04-22
Bug 1827081: Adds imagestream for jenkins-agent-base and migrates others to yaml

akram:


gabemontero:
@akram - how does adding an jenkins agent base imagestream have bearing on have the nodejs v10 agent image on registry.redhat.io, which the referenced bugzilla refers to?

akram:
Hi @gabemontero , I completely missed to reply.

As the release of nodejs10 and 12 agent images was taking long because of the release process and because it is resources consuming accross teams, we wanted to give users the ability to benefit more easily from any image provided by Red Hat by using the ContainerTemplate feature of the kubernetes plugin.

Doing such would require to have the jenkins-base-image to be available as an imagestream which is not the case actually.

PR #1068, opened 2020-06-04
Bug 1843945: bump jenkins to 2.222.1 and update plugins

jkhelil:


jkhelil:
/retitle Bug 1843945: bump jenkins to 2.222.1 and update plugins

PR #1078, opened 2020-06-09
Bug 1780125: fix proxy settings in jenkins

jkhelil:


2/contrib/openshift/configuration/proxy.yaml.tpl
```diff
@@ -0,0 +1,8 @@
+jenkins:
+  proxy:
+    name: "${PROXY_HOST}"
+    port: ${PROXY_PORT}
+    userName: "${PROXY_USER}"
+    secretPassword: "${PROXY_PASSWORD}"
+    noProxyHost: "${NO_PROXY}"
+    testUrl: "http://redhat.com"
```
akram: if the parameter is not mandatory, maybe we should remove it, or have also a way to set it; because it may cause issues with disconnected environment or with proxies that does filter this url or any test url that we would specify.

2/contrib/s2i/run
```diff
@@ -75,6 +75,27 @@ function generate_administrative_monitors_config() {
...
+        export PROXY_HOST=$( echo $json | jq -r .host)
+        export PROXY_PORT=$( echo $json | jq -r .port)
+        export PROXY_USER=$( echo $json | jq -r .username)
+        export PROXY_PASSWORD=$( echo $json | jq -r .password)
+        if [ -n "$no_proxy" ]; then
+          export NO_PROXY=$(convert_no_proxy $no_proxy)
+        fi
+      fi 
+      # Fix the envsubst trying to substitute the $Hash inside credentials.xml
+      export Hash="\$Hash"
```
akram, line 93: use upper case env var name, maybe PROXY_CREDS_HASH or PROXY_PASSWORD_HASH ?

2/contrib/s2i/run
```diff
@@ -97,6 +118,8 @@ function create_jenkins_config_from_templates() {
         create_jenkins_config_xml
       elif [[ "${target_path}" == "${image_config_dir}/credentials.xml" ]]; then
         create_jenkins_credentials_xml
+      elif [[ "${target_path}" == "${image_config_dir}/proxy.yaml" ]]; then
```
akram, line 121: starts to be many if/elif .
Can you rewrite it using swtich/case/esac ?

openshift/templates/jenkins-ephemeral-monitored.json
```diff
@@ -150,6 +150,10 @@
                   {
                     "name": "JENKINS_UC_INSECURE",
                     "value": "${JENKINS_UC_INSECURE}"
+                  },
```
akram, line 153: why exposing it?

2/contrib/openshift/configuration/proxy.yaml.tpl
jkhelil: fixed

openshift/templates/jenkins-ephemeral-monitored.json
jkhelil: because the casc plugin needs that env var to load the config

2/contrib/s2i/run
jkhelil: fixed

2/contrib/s2i/run
jkhelil: fixed

akram:
/retest

aws flake

PR #1083, opened 2020-06-10
Bug 1847452: Set OPENSHIFT_JENKINS_IMAGE_VERSION

akram:


akram:
image build flake

/retest

bparees:
4.3.z patch window closed earlier due to an urgent issue, this will be considered for merge next week.

akram:
@bparees ok, noted.

shawn-hurley:
Hey folks,

Having a hard time following the BZ: this is the depends on BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1836914

it references this closed PR: https://github.com/openshift/jenkins/pull/1057/files which does not seem to be the same changes is the BZ wrong?

akram:
@shawn-hurley good catch, I just set the dependant to correct one. We had to do the cherry-pick manually for this one as you can see that every release needs a different version in the Dockerfile.

shawn-hurley:
@akram confused because the bug is now verified: https://bugzilla.redhat.com/show_bug.cgi?id=1847452#c4 for 4.3 so wondering why we need to merge this?

sdodson:
Need to understand what's up with the state of the linked bug please.

akram:
@sdodson; regarding the status, we asked QA to run the tests on it before the ART build happens. It seems there is a confusion from our QA resetting it to VERIFIED again. So, I will reset to POST and notify them.

PR #1089, opened 2020-06-19
Bug 1849094: Use quay.io/openshift/origin-cli:4.2 as base image for upstream

akram:
This image contains a version of oc run that still creates
deploymentconfig and this is required by our e2e tests


gabemontero:
/retest

Ben's disablement of the problematic client plugin tests with `oc.run` has merged in openshift/origin

PR #1135, opened 2020-08-11
Bug 1848611: Remove execution of rpm -e java-1.8.0.

akram:
This rpm -e command was added to make rh-maven35 package installable as it requires
java1.8, but then we delete java 1.8 and maven can run with openjdk-11.
With the revert and opt-in strategy for maven35 and openjdk we reverted this one only partially.


akram:
needs https://github.com/openshift/origin/pull/25383

PR #1137, opened 2020-08-12
[openshift-3.11] Bug 1850528: Login plugin 1.0.24: incorrect detection of openshift 3

akram:


akram:
ART request for plugins update: https://issues.redhat.com/browse/ART-2143

PR #1157, opened 2020-09-10
Bug 1883836: Update nodejs-agent imagestream to v12

waveywaves:
Reference location for image : https://catalog.redhat.com/software/containers/openshift4/jenkins-agent-nodejs-12-rhel7/5ea81a4ad70cc54b02d62294?container-tabs=gti

### DIFF
--- a/openshift/imagestreams/jenkins-agent-nodejs-rhel.json
+++ b/openshift/imagestreams/jenkins-agent-nodejs-rhel.json
@@ -37,7 +37,7 @@
         },
         "from": {
           "kind": "DockerImage",
-          "name": "registry.redhat.io/openshift4/ose-jenkins-agent-nodejs:v4.0"
+          "name": "registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12-rhel8:v4.6"
         },
         "referencePolicy": {
           "type": "Local"
--- a/openshift/imagestreams/jenkins-agent-nodejs-rhel.yaml
+++ b/openshift/imagestreams/jenkins-agent-nodejs-rhel.yaml
@@ -28,7 +28,7 @@ spec:
       tags: jenkins
     from:
       kind: DockerImage
-      name: registry.redhat.io/openshift4/ose-jenkins-agent-nodejs:v4.0
+      name: registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12-rhel8:v4.6
     referencePolicy:
       type: Local
 


openshift/imagestreams/jenkins-agent-nodejs-rhel.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       tags: jenkins
     from:
       kind: DockerImage
-      name: registry.redhat.io/openshift4/ose-jenkins-agent-nodejs:v4.0
+      name: registry.redhat.io/openshift4/jenkins-agent-nodejs-12-rhel7:v4.5
```
gabemontero, line 31: I'll have to defer to @yselkowitz on the `registry.redhat.io/openshift4/jenkins-agent-nodejs-12-rhel7` portion of this, as I believe we'll want to  replace this with the ubi8 version once his downstream/brew related stuff is allowed to merge into 4.6.  I think we would want to update this just once.

Now, on whether to use `latest` or `v4.5` and then bumping this as we move forward.  After the slack discussion yesterday I revisited the various elements of this.  I've boiled it down to these elements.

Supported scenario 1:  running this image within an OCP cluster

- I'll point you to 2 spots in the kube-slave-common.sh script

https://github.com/gabemontero/jenkins/blob/118426855357d16aa4548c65c5e8849333c442ac/2/contrib/jenkins/kube-slave-common.sh#L31

and https://github.com/gabemontero/jenkins/blob/118426855357d16aa4548c65c5e8849333c442ac/2/contrib/jenkins/kube-slave-common.sh#L126

when running in an OCP cluster, unless the user has overrode the value by setting the NODEJS_SLAVE_IMAGE env var per the doc at https://github.com/openshift/jenkins#environment-variables and https://docs.openshift.com/container-platform/4.5/openshift_images/using_images/images-other-jenkins.html#images-other-jenkins-env-var_images-other-jenkins

I had forgotten about this "improvement" over 3.x when we ended up in migration hell.  Since the kubernetes plugin config is using the imagestream image reference vs. the registry.redhat.io image reference, this protects us from the update of the `latest` tag at registry.redhat.io from inadvertently updating the agent image used on clusters with older releases.

Supported scenario 2:  running this image as part of a standalone jenkins instance

In this case, the default at https://github.com/gabemontero/jenkins/blob/118426855357d16aa4548c65c5e8849333c442ac/2/contrib/jenkins/kube-slave-common.sh#L31 will not work, unless the user goes through a lot of work to make the host `image-registry.openshift-image-registry.svc:5000` DNS accessible in their environment.

Most likely they either manually edit the config after the fact, or set the NODEJS_SLAVE_IMAGE env var when launching the image.

**IF** they set it to say `registry.redhat.io/openshift4/jenkins-agent-nodejs-12-rhel7:latest` they then end up with the same problem I described above, where the version of the image they use in say their 4.4 cluster will change on them whenever we change `registry.redhat.io/openshift4/jenkins-agent-nodejs-12-rhel7:latest` as part of new releases going out.

So how best to protect them?  Either:
- we update the doc at https://github.com/openshift/jenkins#environment-variables and at https://docs.openshift.com/container-platform/4.5/openshift_images/using_images/images-other-jenkins.html#images-other-jenkins-env-var_images-other-jenkins to explain to them that overriding NODEJS_SLAVE_IMAGE and MAVEN_SLAVE_IMAGE should be done using specific version tags and not the `latest` tag
- if they are using these imagestream definitions as their guide to picking the image to use, we make them version specific and keep them up to date as we see fit
- or both

@akram @waveywaves I'll ultimately defer to you all since you have to support folks with this stuff, as to which image tag to use here .... `latest` (and don't have to worry about updating release to release) or `v4.x` (where you decide if the images have changed significantly enough to warrant a change release to release).

I believe I've given you all the facts you need for deciding :-)

yselkowitz: The nodejs agent is not being renamed, it will still be `registry.redhat.io/openshift4/ose-jenkins-agent-nodejs` and will be based on node v12 in 4.6.

akram: @gabemontero honestly, I still do not understand all the impacts, as it seems to me that there is is a hidden part that we (@waveywaves  or @akram or @jkhelil )  don't have control on: the replacement by the cluster sample operator of the tag to be matching the version.

As @yselkowitz  said, since the image is unchanged for 4.6 (except the tag); we can remain with the old value. right?

regarding the execution of the image in standalone mode: is this really happening? especially with 4.x , where the registry is authenticated, I am willing to think that users are mostly running in ocp.

So, should we just close this PR?

gabemontero: There certainly are aspects beyond yours or my direct control @akram 

And there are multiple scenarios to consider.  Running in openshift.  Running outside openshift.

That said, I do still feel there is a warranted change for this file that needs to be sorted out.

However, before it can, I believe the PR https://github.com/openshift/ocp-build-data/pull/671 needs to be sorted out. 

I've got an update there that tries to explain things again, but the explanation is lengthy like the one above.

Perhaps a conference call to walk through the particulars is called for at this point.  Let's see how the discussion in  https://github.com/openshift/ocp-build-data/pull/671 goes today and then I'll see about setting a call up for either Friday or early next week based on availability.

gabemontero: End of the day, getting jenkins out of the payload (I know you all are motivated to do that but have hit some obstacles beyond your control there as well) would simplify this a bit.

yselkowitz: Change in plans, for 4.6 use: `registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12:$TAG` (either `:v4.6` or `:latest`)

openshift/imagestreams/jenkins-agent-nodejs-rhel.json
```diff
@@ -37,7 +37,7 @@
         },
         "from": {
           "kind": "DockerImage",
-          "name": "registry.redhat.io/openshift4/ose-jenkins-agent-nodejs:v4.0"
+          "name": "registry.redhat.io/openshift4/jenkins-agent-nodejs-12-rhel7:v4.5"
```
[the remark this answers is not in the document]
yselkowitz: Change in plans, for 4.6 use: `registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12:$TAG` (either `:v4.6` or `:latest`)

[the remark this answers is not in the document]
gabemontero: registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12:v4.6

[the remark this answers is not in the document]
gabemontero: correction received from @sosiouxme 

registry.redhat.io/openshift4/ose-jenkins-agent-nodejs-12-rhel8:v4.6

PR #1184, opened 2020-11-16
[master] Bug 1897584: add casc plugins

jkhelil:


jkhelil:
https://issues.redhat.com/browse/ART-2351

PR #1214, opened 2021-02-08
Update jenkins-rhel.yaml

yigitpolat:
v4.0 tag is no longer available on the Red Hat Registry. It can be replaced with any tag, but since other tags may also be deleted, I just simply wanted to use the latest tag. However, the pod has never skipped the readiness checks. So that, I tried v4.4.0 and worked perfectly.

This is also related with https://github.com/openshift/cluster-samples-operator/pull/354

Below shows the Import Failure issue.
Screen Shot 2021-02-08 at 10 54 51

### DIFF
--- a/openshift/imagestreams/jenkins-rhel.yaml
+++ b/openshift/imagestreams/jenkins-rhel.yaml
@@ -32,7 +32,7 @@ spec:
       version: 2.x
     from:
       kind: DockerImage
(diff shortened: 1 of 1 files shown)


akram:
Hi @yigitpolat,

This file is actually used by the cluster sample operator to operate a replacement with the real image coordinate.
So v4.0 is acutally just a placeholder.
I will close this PR then but feel free to use a fork of this file for you needs.

PR #1222, opened 2021-02-09
[release 4.5] Bug 1926754: Update maven agent to 3.6

waveywaves:


bparees:
@waveywaves 
1) this appears to be an RFE (per the linked bug), we don't generally backport features to z-streams, especially not 4.5 which is in maintenance support only.

2) seems odd you've changed the content w/o renaming the "maven-3.5" directories.

siamaksade:
@bparees Maven 3.5 is EOL which leaves 4.5 with no supported Maven agents for Jenkins. Therefore we aimed to backport to 4.5 as well, despite being in maintenance mode.

bparees:
not renaming the directories still seems like an odd choice but i don't want to hold you up from landing this, so i'm approving.

siamaksade:
@bparees agree about the directories

@waveywaves any reason for not adding a new directory for maven-3.6?

akram:
@siamaksade the initial fix has been done by a new comer on the project, and it has been LGTMed too quickly while I was not here.
for 3.11 , I have created a different directory, so, we should align. However, 4.x directory should be name `agent-maven`  to reflect the final image name which is version agnostic.

To complete the renaming, a PR within https://github.com/openshift/ocp-build-data/blob/openshift-4.5/images/jenkins-agent-maven-35-rhel7.yml should be done and that also may have impacts on ART build process.

akram:
one of the test pipeline is failing with the following error
```
First time build. Skipping changelog.
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Build WAR)
[Pipeline] sh
+ mvn clean package -Popenshift
/tmp/workspace/e2e-test-jenkins-pipeline-7b6mc/e2e-test-jenkins-pipeline-7b6mc-openshift-jee-sample@tmp/durable-0ed0f482/script.sh: line 1: mvn: command not found
[Pipeline] }
[Pipeline] // stage
```

agent-maven-3.5/Dockerfile.localdev
```diff
@@ -2,25 +2,33 @@ FROM quay.io/openshift/origin-jenkins-agent-base:v4.0
...
+ENV MAVEN_VERSION=3.6.3 \
     BASH_ENV=/usr/local/bin/scl_enable \
     ENV=/usr/local/bin/scl_enable \
     PROMPT_COMMAND=". /usr/local/bin/scl_enable" \
     LANG=en_US.UTF-8 \
     LC_ALL=en_US.UTF-8 \
     MAVEN_OPTS="-Duser.home=$HOME"
+    M2_HOME="/opt/maven"
+    MAVEN_HOME="/opt/maven"
+    PATH="${M2_HOME}/bin:${PATH}"
```
akram, line 14: `\` are missing at the end of the lines.
The directory containg `mvn` must be added to PATH. It is if we take it from community.
see here https://github.com/openshift/jenkins/pull/1213/files#diff-6fa4f3817d9f36f09234c2a759b521340bd27d92c3f4a56e172330482f5574a3R10

The e2e test uses the community build for testing.

akram (changes requested):

PR #1254, opened 2021-04-15
[WIP]add proxy system properties as expected by kubernetes-client

jkhelil:
system properties are expected in this format , keeping the old ones because it is the standard way for java application to receive the proxy settigns, so some other plugins may use the standard way

https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L119

Putting this in WIP because it needs https://github.com/fabric8io/kubernetes-client/pull/3000
I will then bump the plugin in the current PR


jkhelil:
Not relevant replaced by https://github.com/openshift/jenkins/pull/1263

PR #1258, opened 2021-04-20
Bug 1940558: Update Jenkins to 2.277.3 and disable startup wizard

waveywaves:


2/contrib/openshift/configuration/init.groovy.d/setup_wizard_disable.groovy
```diff
@@ -0,0 +1,9 @@
+#!groovy
+
+import jenkins.model.*
+import hudson.util.*;
+import jenkins.install.*;
+
+def instance = Jenkins.getInstance()
+
+instance.setInstallState(InstallState.INITIAL_SETUP_COMPLETED)
```
jkhelil, line 9: Is this needed, if we have the system property, is it suffisant ?

waveywaves: System property is sufficient, I tested it.

PR #1261, opened 2021-04-22
Bug 1891694: Update Jenkins to 2.277.3 and disable setup wizard

waveywaves:


jkhelil:
https://issues.redhat.com/browse/ART-3020

adambkaplan:
@waveywaves @jkhelil @akram after this change, Jenkins is failing QA.

Given that the CVEs in the security advisory are rated "Moderate" and we are now in maintenance support, we should revert this change and roll back the ART bump.

PR #1263, opened 2021-04-23
Bug 1956220: add proxy system properties as expected by kubernetes-client

jkhelil:


jkhelil:
https://issues.redhat.com/browse/ART-2916

PR #1269, opened 2021-05-11
Bug 1972451: Correctly declare Jenkins URL with trailing slash

olivergondza:
This is normally enforced by Jenkins itself, though since this is injected through direct modification of the XML file, the incorrect version stays in. There are rare situations when plugins do not expect the values does not adhere to invariants imposed by Jenkins and fail when run in OpenShift:

https://github.com/jenkinsci/jenkins/blob/a1175b932e1ac97085d69752f5821141d012314f/core/src/main/java/jenkins/model/JenkinsLocationConfiguration.java#L144


akram:
/retitle Bug 1972451: Correctly declare Jenkins URL with trailing slash

olivergondza:
Can you guys please take a look at the tests? From the little I understand about the test output, it seems unrelated. Thanks for any pointers!

PR #1273, opened 2021-05-28
Bug 1973109: Fixed MAVEN _VERSION env var

pittar:
The MAVEN_VERSION env var was incorrectly set to "3.5"


akram:
/retitle Bug 1973109: Fixed MAVEN _VERSION env var

Issue #1292, opened 2021-06-18
kubernetes plugin has been updated but not pipeline DSL


max-allan-surevine:
Someone seems to have upped the version of the k8s plugin but not the "Pipeline: DSL" plugin.
So, you get a failure to load the k8s plugin on startup. If you're running in an openshift cluster or have a k8s cluster defined in Clouds, this is pretty catastrophic and Jenkins will not start.
(Seen in quay.io/openshift/origin-jenkins@sha256:435b59b8ef826cbc06f6e2e1ca3507b50390fd9318ddedfa60767bb832d49343 )

```
2021-06-18 11:26:29 SEVERE  jenkins.InitReactorRunner$1 onTaskFailed Failed Loading plugin Kubernetes plugin v1.29.7 (kubernetes)
java.io.IOException: Failed to load: Kubernetes plugin (1.29.7)
 - Update required: Pipeline: Groovy (2.87) to be updated to 2.89 or higher
	at hudson.PluginWrapper.resolvePluginDependencies(PluginWrapper.java:963)
```

Looking at 2/contrib/openshift/base-plugins.txt :
```
...
kubernetes:1.29.7
...
workflow-cps:2.87
...
```

akram:
Thanks Allan for reporting it.
Because of a hard dependency on openshift-client plugin, this dependency has been erratically pull during our tests.
We had to update openshit client plugin to make the workflow-cps pinning to work.

Our build script were reporting that workflow-cps:2.90 was used, but, 2.87 was remained packaged.
This is pending fix.

akram:
it has been fixed over the weekend by #1294 
can you please confirm that it is also working for you @max-allan-surevine ?
Please re-open if it is not.
/close

PR #1294, opened 2021-06-18
Upgrade various plugins to prepare openshift-sync 1.0.47

akram:


akram:
Fixes #1292

max-allan-surevine:
Please see issue  #1292
The current released image of Jenkins onn quay.io is broken for anyone using a kubernetes cloud. It urgently needs workflow-cps:2.87 which is required by kubernetes:1.29.7 (which is already IN the plugins list)

PR #1296, opened 2021-06-22
[helm] Chart for Jenkins

waveywaves:
Convert jenkins template(s) to helm chart
- Init helm chart
- Convert ephemeral template to chart
- Add in values.yaml to set persistence after which PVC would be used for Jenkins volume (replicate jenkins persistent template usage)


akram:
closed in favor of https://github.com/openshift-helm-charts/charts/tree/main/charts/redhat/redhat/jenkins
/close

PR #1302, opened 2021-07-08
adding persistent template test and updating some

jitendar-singh:
- Adding test for persistent template
- Adding scenario to tear down the test project for all resources created while testing the ephemeral template
- Adding maven agent test for persistent template
- Adding nodejs agent test for persistent template

### DIFF
--- a/smoke/features/persistent-jenkins.feature
+++ b/smoke/features/persistent-jenkins.feature
@@ -1,0 +1,22 @@
+Feature: Deploy Jenkins with persistent volume on openshift using template based install
+
+    As a user of openshift
+    I want to deploy Jenkins with persistent volume and configure my CI/CD on openshift cluster
+
+    Background:
+    Given Project [TEST_NAMESPACE] is used
+
+    Scenario: Create jenkins  using persistent template
+        Given we have a openshift cluster
+        When User enters oc new-app jenkins-persistent command
+        Then route.route.openshift.io "jenkins" created
+        And  configmap "jenkins-trusted-ca-bundle" created
+        And  persistentvolumeclaim "jenkins" created
+        Then we check the pvc status is "Bound"
+        And  deploymentconfig.apps.openshift.io "jenkins" created
+        And  serviceaccount "jenkins" created
+        And rolebinding.authorization.openshift.io "jenkins_edit" created
+        And service "jenkins-jnlp" created
+        And service "jenkins" created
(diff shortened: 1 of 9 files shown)


smoke/features/persistent-jenkins.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Deploy Jenkins with persistent volume on openshift using template based install
```
gabemontero, line 1: So best as I can tell, this test assumes the persistent volumes are created prior to the invocation of the test.

If that is correct @jitendar-singh I believe that should be explicitly called out here and in the other persistent*.feature files.

gabemontero: Or, ideally, you replicate the NFS based PV setup we do in the current extended tests at https://github.com/openshift/origin/blob/master/test/extended/builds/pipeline_jenkins_e2e.go#L60-L72

Then we could be on a path of removing those persistent template tests in openshift/origin and replace them with CI invocations of your new smoke tests.

gabemontero (commented):
Aside from the PV related suggestions below @jitendar-singh 

- outside of this PR, but ideally sooner rather than later, if you are not already on this path, I'd love to see all this cool stuff you are doing invoked from CI for any PRs against this repo
- I could also see use defining periodic / batch jobs of the smoke tests against various versions of openshift
- a few pointers from what we do in other repos that could help get you started
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/builder/openshift-builder-master.yaml#L52-L59
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/csi-driver-projected-resource/openshift-csi-driver-projected-resource-master.yaml#L44-L65

and add analogous entries for the smoke tests to

https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/jenkins/openshift-jenkins-master-presubmits.yaml
and
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/jenkins/openshift-jenkins-master.yaml

Otherwise, I'm certainly not fluent enough in python to offer you stylistic / detailed comments on that actual code.  To some extent, being able to run them from CI with repetition and consistency would be the true validation for me.

smoke/features/persistent-jenkins.feature
jitendar-singh: Thanks @gabemontero I will work this next.

PR #1306, opened 2021-07-13
Test the data persistence of jenkins persistent template

jitendar-singh:
- Adding feature to test the data persistence of jenkins persistent template
- Adding a function to get the jenkins master pod
- Adding function to exec into master pod and run command inside the pod to get the job count
- Refactored the nodejs agent test to check for the builds to complete
- Removed some unwanted code

### DIFF
--- a/smoke/features/jenkins-nodejs-agent.feature
+++ b/smoke/features/jenkins-nodejs-agent.feature
@@ -11,5 +11,6 @@ Feature: Testing jenkins agent nodejs image
     Given The jenkins pod is up and runnning
     When The user enters new-app command with nodejs_template
     Then Trigger the build using oc start-build
-    Then nodejs-postgresql-example pod must come up
+    Then verify the build status of "nodejs-postgresql-example-1" build is Complete
+    Then verify the build status of "nodejs-postgresql-example-2" build is Complete
     And route nodejs-postgresql-example must be created and be accessible
--- a/smoke/features/steps/steps.py
+++ b/smoke/features/steps/steps.py
@@ -31,7 +31,7 @@ config.load_kube_config()
 v1 = client.CoreV1Api()
 oc = Openshift()
 podStatus = {}
-
+buildconfigs = {'sample-pipeline':'1','openshift-jee-sample':'1'}
 # Parse the base plugins from the file and store them in a dictonary with key=plugin-name & value=plugin-version
 
 baseplugins = './2/contrib/openshift/base-plugins.txt'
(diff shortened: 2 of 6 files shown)


smoke/features/steps/steps.py
```diff
@@ -454,12 +437,53 @@ def del_pods(context):
...
+    for jobnames,_ in buildconfigs.items():
+        exec_command = 'cat /var/lib/jenkins/jobs/'+current_project+'/jobs/'+current_project+'-'+jobnames+'/nextBuildNumber'
+        count = oc.exec_in_pod(jenkins_master_pod,exec_command)
+        buildconfigs[jobnames] = str(count)
+    print(buildconfigs)
+
+@when(u'We delete the jenkins master pod')
+def deletemaster(context):
+    master_pod = getmasterpod(current_project)
+    res = oc.delete("pods",master_pod,current_project)
```
gabemontero, line 453: curious why you delete the master pod vs. scaling it down like in https://github.com/openshift/origin/blob/master/test/extended/builds/pipeline_jenkins_e2e.go#L261

gabemontero: not saying the pod delete, let the deployment config function recreate pod, is a bad thing per se

but in theory in the future if we wanted to examine things while jenkins was down, using scale down and scale up would allow for that

smoke/features/jenkins-nodejs-agent.feature
```diff
@@ -11,5 +11,6 @@ Feature: Testing jenkins agent nodejs image
     Given The jenkins pod is up and runnning
     When The user enters new-app command with nodejs_template
     Then Trigger the build using oc start-build
-    Then nodejs-postgresql-example pod must come up
+    Then verify the build status of "nodejs-postgresql-example-1" build is Complete
+    Then verify the build status of "nodejs-postgresql-example-2" build is Complete
```
gabemontero, line 15: unclear how this is related to the persistence template

did you just decide to add more verification, regardless of persistent vs. ephemeral?

if so, I think the PR title should capture that.

If this is somehow related to persistence, please clarify.

gabemontero (commented):
A few minor questions @jitendar-singh with perhaps some minor changes needed to the PR title

Also, am I correct in assuming the smoke test still requires the PVs being established before running this?  If not, was PV creation done elsewhere already?

jitendar-singh:
@gabemontero Yes the test assumes that the cluster has PV created. As asked earlier by you I am yet to start implementing your suggestion on PV.

smoke/features/jenkins-nodejs-agent.feature
jitendar-singh: This test the agent images on jenkins installed using persistent template. O already have test of agent on ephemeral templates. But wanted to make sure the agents are tested on persistent templates too

gabemontero: ok thanks for clarifying ... I'm good with things as is

gabemontero:
ok @jitendar-singh the only remaining question is https://github.com/openshift/jenkins/pull/1306#discussion_r668896954

smoke/features/steps/steps.py
jitendar-singh: @gabemontero I was thinking of adding that along side some more test cases to test the performance & robustness of jenkins template based install.

gabemontero: so you mean in a subsequent PR, where in addition to those perf and robustness tests you possibly change this test to use scale vs. delete ?

jitendar-singh: Yes

PR #1312, opened 2021-08-09
Add support for websocket connection to Jenkins

GolanTrevize10:
When you configure Kubernetes cloud configuration in Jenkins to use Websocket connection, the Kubernetes plugin adds the JENKINS_WEB_SOCKET env variable, but that variable is ignored by the agent base image and it does not add the webSocket flag.

### DIFF
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -101,6 +101,12 @@ if [[ $# -eq 0 ]]; then
   if [ ! -z "$JENKINS_NAME" ]; then
     set -- "${@}" "$JENKINS_NAME"
   fi
+  
+  # If Kubernetes plugin is configured to use WebSocket, add the -webSocket flag to the arguments
+  if [ ! -z "$JENKINS_WEB_SOCKET" ]; then
+    set -- "${@}" "-webSocket"
+  fi
+
 fi
 
 # if `docker run` has 2 or more arguments the user is passing jenkins launcher arguments
--- a/slave-base/contrib/bin/run-jnlp-client
+++ b/slave-base/contrib/bin/run-jnlp-client
@@ -103,8 +103,10 @@ if [[ $# -eq 0 ]]; then
   fi
   
   # If Kubernetes plugin is configured to use WebSocket, add the -webSocket flag to the arguments
-  if [ ! -z "$JENKINS_WEB_SOCKET" ]; then
-    set -- "${@}" "-webSocket"
+  if [[ "$@" != *"-webSocket"* ]]; then
+    if [ ! -z "$JENKINS_WEB_SOCKET" ]; then
+      set -- "${@}" "-webSocket"
+    fi
   fi
 
 fi


akram:
@GolanTrevize10 you need to expose the JENKINS_WEB_SOCKET flag in the Jenkins template as well to be sure that we can pass it as an environment variable in the jenkins pod.
Otherwise, user experience may not be sastisfying as it will require to manually update the `DeploymentConfig` after first deployment occurs.

As-is, also, this change will not get test by our e2e as it will be completely transparent.

GolanTrevize10:
@akram the JENKINS_WEB_SOCKET env variable is passed by the Kubernetes plugin when you click on the Websocket checkbox in Kubernetes cloud configuration in Jenkins

slave-base/contrib/bin/run-jnlp-client
```diff
@@ -101,6 +101,12 @@ if [[ $# -eq 0 ]]; then
   if [ ! -z "$JENKINS_NAME" ]; then
     set -- "${@}" "$JENKINS_NAME"
   fi
+  
+  # If Kubernetes plugin is configured to use WebSocket, add the -webSocket flag to the arguments
+  if [ ! -z "$JENKINS_WEB_SOCKET" ]; then
+    set -- "${@}" "-webSocket"
+  fi
```
jemerald, line 108: Consider add a check to see if the parameter already includes `-webSocket`?

```suggestion
  if [[ "$@" != *"-webSocket"* ]]; then
    if [ ! -z "$JENKINS_WEB_SOCKET" ]; then
      set -- "${@}" "-webSocket"
    fi
  fi
```

jemerald:
I raised https://github.com/openshift/jenkins/issues/1343 before finding this PR. Looking forward to progress of this.

@maximili0n FYI - it's still possible to enable web socket without this PR by adding `-webSocket` to the slave container template's command argument, e.g. 

image

maximili0n:
@jemerald I just tried your proposed solution, but I'm always getting the following error in the agent:

`flag provided but not defined: -webSocket
Usage of /usr/bin/go-init:
  -main string
    	Main command
  -post string
    	Post-stop command
  -pre string
    	Pre-start command
  -version
    	Display go-init version`

jemerald:
@maximili0n the `-webSocket` parameter needs to be passed to the `/usr/local/bin/run-jnlp-client` script.

I've changed the command to bypass `go-init` like this
image

GolanTrevize10:
@jemerald I added your changes. Your workaround works

jemerald (approved):

jemerald:
Any update on this PR? Anything blocking this to be merged?

akram:
@jemerald can you test a version using go-init ?
We are using go-init because of a known issue related to zombie processes remaining on node when agent container is not using an init process.

jemerald:
@akram I have tested it with the following setting and it works fine

image

akram:
@jemerald I can't see the modification of the templates to use `go-init` 
Can you please explicit that point?

jemerald:
@akram for containerTemplate, it's the `command` and `args`, try something like:

```yaml
command: "/usr/bin/go-init",
args: "-main \"/usr/local/bin/run-jnlp-client -webSocket ${computer.jnlpmac} ${computer.name}\""
```

akram:
@jemerald this has to be reflected in these various locations, but also some other ones:

https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L103
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L134
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L165
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L180
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/kube-slave-common.sh#L226


This is what I mean when I say, that we cannot release it as-is. The integration of this new option must be enable by default so it can also be tested by our e2e tests.

jemerald:
@akram I don't quite get why those default pod templates needs to change.

Whether the `-webSocket` parameter should be added, really depends on whether user chooses the `WebSocket` option for the Kubernetes cloud:

image

This is how I see this PR works:

1. By default the Kubernetes cloud is created with `WebSocket` disabled - slaves would be using JNLP for communication
2. User enables the `WebSocket` option, and sets the `Jenkins tunnel` option to blank (and if need to run slaves from different namespace/cluster, change the `Jenkins URL` to use the route instead of IP)
3. The [Kubernetes plugin](https://github.com/jenkinsci/kubernetes-plugin/pull/661/files#diff-07d3a3100d41642c232eaf8786fc1e3f480e581662df30d0ec568bdfd7877a7dR349-R351) sees the `WebSocket` option is enabled on the cloud, and adds the `JENKINS_WEB_SOCKET` environment variable to the slave pod
4. The slave run script sees the `JENKINS_WEB_SOCKET` being set, adds the `-webSocket` option when running the jar, see example implementation in [docker-inbound-agent](https://github.com/jenkinsci/docker-inbound-agent/pull/130/files#diff-b28e3fa5181c0f270db6a279de41018f3685140ea88759182cb053c5b8b07a7fR72-R75)

Unless you are suggesting that the Kubernetes config generated by `kube-slave-common.sh` should enable WebSocket by default?

akram:
@jemerald OK, I got it now. 
I was indeed thinking that this PR was intending to enable `websocket` by default for all agents. And I wanted to test that.
Given your arguments, it is more reasonable indeed to only allow the websocket option to be enabled and not swallowed.

/lgtm

akram (approved):

akram:
Fixes #1343

PR #1332, opened 2021-11-04
Bug 2020616: update jenkins version to 2.303.3 (security update)

alicerum:
In order to fix `CVE-2014-3577` and a bunch of other security issues, jenkins version is bumped to a current LTS, `2.303.3`.

### DIFF
--- a/2/contrib/jenkins/install-jenkins-core-plugins.sh
+++ b/2/contrib/jenkins/install-jenkins-core-plugins.sh
@@ -18,8 +18,12 @@ if [[ "${INSTALL_JENKINS_VIA_RPMS}" == "false" ]]; then
       rm -fr /var/cache/yum/x86_64/7Server/*
       rm -fr /var/cache/yum/x86_64/7Server/ # Clean yum cache otherwise, it will fail if --disablerepos are specified
     fi
-    yum -y $YUM_FLAGS --setopt=tsflags=nodocs --disableplugin=subscription-manager install jenkins-2.289.2
-    rpm -V jenkins-2.289.2
+    # Since the recent LTS jenkins update we need to install the 'daemonize' package
+    # which is only available in EPEL, so enable it here
+    yum -y --setopt=tsflags=nodocs --disableplugin=subscription-manager install \
+	    https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+    yum -y $YUM_FLAGS --setopt=tsflags=nodocs --disableplugin=subscription-manager install jenkins-2.303.3
+    rpm -V jenkins-2.303.3
     yum $YUM_FLAGS clean all
     /usr/local/bin/install-plugins.sh $PLUGIN_LIST
 else
--- a/2/Dockerfile.localdev
+++ b/2/Dockerfile.localdev
@@ -30,7 +30,7 @@ ENV JENKINS_VERSION=2 \
     HOME=/var/lib/jenkins \
     JENKINS_HOME=/var/lib/jenkins \
     JENKINS_UC=https://updates.jenkins.io \
-    OPENSHIFT_JENKINS_IMAGE_VERSION=4.9 \
+    OPENSHIFT_JENKINS_IMAGE_VERSION=4.10 \
     LANG=en_US.UTF-8 \
     LC_ALL=en_US.UTF-8 \
     INSTALL_JENKINS_VIA_RPMS=false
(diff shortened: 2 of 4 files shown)


akram (commented):
@alicerum It is also needed to change the label to match the Jenkins version.

akram:
ART team notified me a while ago about issues on the ubi image and the available repo. I wonder if it is not this affecting the build of the image.
Investigating it need to restart

/retest

akram:
@alicerum I have created another PR similar to yours to be able to investigate the ci issue. As a PR creator, one can `oc login https://api.build02.gcp.ci.openshift.org:6443` and better see what happens. That helped me to better see that the image build is failing because now jenkins requires a package named `daemonize` which is available in the EPEL repository.

We can add that for the community image build process that is produced acutally by CI, but, I wonder if the daemonize package is available in some repos for the Red Hat bits. ART may help in this case.

akram:
@alicerum can you PTAL at https://github.com/openshift/jenkins/pull/1333 ?
using epel indeed fixes the image build issue.

2/contrib/jenkins/install-jenkins-core-plugins.sh
```diff
@@ -18,8 +18,12 @@ if [[ "${INSTALL_JENKINS_VIA_RPMS}" == "false" ]]; then
...
       rm -fr /var/cache/yum/x86_64/7Server/ # Clean yum cache otherwise, it will fail if --disablerepos are specified
     fi
-    yum -y $YUM_FLAGS --setopt=tsflags=nodocs --disableplugin=subscription-manager install jenkins-2.289.2
-    rpm -V jenkins-2.289.2
+    # Since the recent LTS jenkins update we need to install the 'daemonize' package
+    # which is only available in EPEL, so enable it here
+    yum -y --setopt=tsflags=nodocs --disableplugin=subscription-manager install \
+	    https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+    yum -y $YUM_FLAGS --setopt=tsflags=nodocs --disableplugin=subscription-manager install jenkins-2.303.3
+    rpm -V jenkins-2.303.3
```
adambkaplan, line 26: @akram I am pretty certain this will break our downstream build with ART. Do we need to coordinate with the test platform team and/or ART to address this?

adambkaplan (commented):
/hold

For confirmation on how we can get the `daemonize` package into our build.

2/contrib/jenkins/install-jenkins-core-plugins.sh
gabemontero: yeah this has come up before ... we have to use rhel/ubi packages not fedora 

as @akram noted in https://github.com/openshift/jenkins/pull/1332#issuecomment-961316852

including reaching out to the ART team in their channel to find out where in rhel land the `daemonize` package lives .... they have been helpful with that in the past

yselkowitz: I don't see any such package internally, only Fedora and EPEL (which, as you noted, is not acceptable for OCP builds).

akram: @adambkaplan ART is setting the INSTALL_JENKINS_VIA_RPMS env var in osbs/brew, so they are passing here:
https://github.com/openshift/jenkins/blob/663a4b706c50aae7dddc6d11d7bdfc79ff4af7a6/2/contrib/jenkins/install-jenkins-core-plugins.sh#L29

we should be ok

adambkaplan: @yselkowitz so for our downstream build we make our own RPM spec that has nothing to do with what upstream does. We don't need `daemonize`: https://pkgs.devel.redhat.com/cgit/rpms/jenkins/tree/jenkins.spec?h=rhaos-4.9-rhel-8 (needs VPN).

@alicerum please update this PR so that the enablement of EPEL happens inside the `[[ "${INSTALL_JENKINS_VIA_RPMS}" == "false" ]];` if block.

gabemontero:
@alicerum - go ahead and update subversion to 2.15.1 at https://github.com/openshift/jenkins/blob/master/2/contrib/openshift/base-plugins.txt#L35 per the notice from the same advisory https://www.jenkins.io/security/advisory/2021-11-04/#fix

do it in a separate commit

2/contrib/jenkins/install-jenkins-core-plugins.sh
alicerum: @adambkaplan it already is inside this `if` block, which begins earlier in the code. in the current state `EPEL` should not be enabled for our own downstream build.

adambkaplan: ack - misread the code then

adambkaplan:
/hold cancel

Our upstream build process can use EPEL. The downstream process bypasses this by passing the `INSTALL_JENKINS_VIA_RPMS=true` environment variable.

adambkaplan:
Note - cherry-pick will need to be done manually since we are updating the OCP image version label here.

adambkaplan (approved):
/lgtm

gloria-zmm:
@alicerum @adambkaplan 
When can we expect the release of this PR?
We are waiting for the release to fix the critical vulnerabilities.

adambkaplan:
ART request: https://issues.redhat.com/browse/ART-3513

PR #1334, opened 2021-11-08
Bug 2003204: Fix jbcrypt_path lookup

akram:
In resulting image, jbcrypt is not extracted into /tmp/war but rather it is present under directories in /var/lib/jenkins/war/


adambkaplan (commented):
@akram do we have a test case where we deploy the Jenkins template with the oauth plugin disabled? We need to ensure that we don't regress if there are upstream Jenkins changes in the future.

akram:
@adambkaplan I didn't see a such a test, I will prepare one.

akram:
closing in favor of #1345

PR #1344, opened 2021-11-16
Fixes #1241

### LINKED ISSUE #1241: Bug: Jenkins will failed to notify stash/bitbucket when there is multiple routes in namespace
Hi team, we met an issue that Jenkins will failed to notify stash with below error:

```
[Pipeline] step
Cannot notify Stash! (Jenkins Root URL not configured)
[Pipeline] sh
```
As we found the `JENKINS_URL` was not set correctly in `/var/lib/jenkins.model.JenkinsLocationConfiguration.xml` file.

After digging into log and source code, we found a bug in bellow file:

https://github.com/openshift/jenkins/blob/84056bf7c0fc2b42b118761e610953e9b2fc89c5/2/contrib/s2i/run#L19

```
--> JENKINS_URL="$JENKINS_SCHEME$(oc get routes -o json | jq -c -r '.items[0].spec | select (.to.name == env.JENKINS_SERVICE_NAME ) | .host ')"
```
It was trying to get `JENKINS_URL` from 1st route in namespace, however, we have multiple routes in namespace currently, jenkins route is not the first one.

Could you please confirm the issue and get it fixed?

aiman-alsari:
Rather than picking the first route then checking if it matches, select from *all routes* then grab the host of the matching one.

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -574,7 +574,7 @@ create_jenkins_location_configuration_xml() {
   if [ $( oc get routes -o json | jq -c -r '.items[0].spec.tls')  == "null" ]; then
     JENKINS_SCHEME="http://"
   fi
-  JENKINS_URL="$JENKINS_SCHEME$(oc get routes -o json | jq -c -r '.items[0].spec | select (.to.name == env.JENKINS_SERVICE_NAME ) | .host ')"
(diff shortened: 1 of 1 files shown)


akram:
Hi @aiman-alsari, thank you for your contribution. That looks ok to be merged; unless you wanted more tests?

aiman-alsari:
no, I'm good. Thanks @akram

PR #1345, opened 2021-11-18
Bug 2003204: Fix jbcrypt_path lookup and add openshift enable oauth env support

akram:


gabemontero:
when you think this is ready @akram just do a "/assign" on my github ID and I can tag

one of these days, we probably should run `shellcheck` against our bash scripts, but fine with not worrying about that for this bug that needs to be expedited

akram:
/assign gabemontero

Gabe that's ok now. We have a working solution.

PR #1352, opened 2021-11-23
Bug 2003204: Fix the admin startup on ephemeral volume initial password case

akram:


adambkaplan (commented):
/hold

Do we have a test to ensure that we don't regress on this capability in the future?

akram:
@adambkaplan actually, after checking, it seems that there was:
https://github.com/openshift/jenkins/blob/master/2/test/jenkins_test.go#L148
These smoke tests are supposed to set `JENKINS_PASSWORD` to a given value and then perform authenticated operations using this password.
Despite the feature was not working, the e2e were not failing. I need to check where they were supposed to be run.

adambkaplan:
Following up from our conversation this morning:

- As you mentioned, the e2e tests in this repo rely on talking to a Docker socket. We would need to significantly refactor these tests to deploy the Jenkins template on OpenShift, and then test using `oc exec` or similar/equivalent commands.
- Getting the smoke tests that @jitendar-singh and @prietyc123 into CI might be a simpler path. The OC cli images come bundled with Python3, so it shouldn't be too difficult to modify what we have so it runs against a CI cluster.

akram:
@adambkaplan here are those changes https://github.com/openshift/jenkins/pull/1353
can you ptal ?

akram:
when #1354 https://github.com/openshift/jenkins/pull/1354 will be merged, we will have the safety net.

gabemontero:
So the test @akram noted in https://github.com/openshift/jenkins/pull/1352#issuecomment-977160944 were the old unit tests we employed in v3 when all our CI still had the docker socket available (I now understand your question in the team meeting last week about this @akram :-) )

Those tests of course are nowadays only valid for local validation, if you happen to have docker running.  They can no longer be called from our CI.  Early in 4.x days the test platform team had hoped to provide a special env to allow for docker with ci-operator, but that ship has sailed and that is not arriving.  I *think* the only hope to redo those tests are to take what @adambkaplan has curated / mainted for the s2i repo and add that to the jenkins repo.  I'll defer to him on the advisability and desire to do that.

Next, I see @akram 's https://github.com/openshift/jenkins/pull/1352#issuecomment-983129234 and the addition of the new smoke tests.

Cool deal, but my question there is simple:  we are not calling the smoke tests from CI, so until that happens we are not getting the PR verification that I believe @adambkaplan was asking for, right?

Or are you ok with folks manually verifying from time to time @adambkaplan ?

adambkaplan:
@akram @gabemontero given the challenges of setting up the smoke tests in CI, does it make sense to merge this as is and separately fix the CI job?

akram:
> I'm curious if such an e2e exercise might motiviate us to expose JENKINS_PASSWORD as a template parameter

I don't know indeed why it has not been done previously, especially that the `OPENSHIFT_OAUTH_ENABLE` is exposed. And disabling oauth without changing the default password is kind of risky.

Based on the current status of the project, I don't think it is worth introducing that now. The `JENKINS_PASSWORD` fix here was following a case where the customer was directly using the image and not the template, which is not documented. So, this is for sure for advanced usages and corner cases.

gabemontero:
> > I'm curious if such an e2e exercise might motiviate us to expose JENKINS_PASSWORD as a template parameter
> 
> I don't know indeed why it has not been done previously, especially that the `OPENSHIFT_OAUTH_ENABLE` is exposed. And disabling oauth without changing the default password is kind of risky.
> 
> Based on the current status of the project, I don't think it is worth introducing that now. The `JENKINS_PASSWORD` fix here was following a case where the customer was directly using the image and not the template, which is not documented. So, this is for sure for advanced usages and corner cases.

fair enough re parameters

that said, an e2e could be added where the env var is set on the deployment, and you still verify your change that way

akram:
> that said, an e2e could be added where the env var is set on the deployment, and you still verify your change that way
agree. The smoke tests does that actually. `oc new-app`  accepts `-e` parameter which is then passed to the resources created by the template and which accepts env car (`Deployment` or `Pod` etc...) .
I will port the smoke test verification part in a separate PR then.

adambkaplan (approved):
/lgtm

PR #1353, opened 2021-11-26
Change smoke tests so they can run in openshift-ci

akram:
Changing the smoke test so they can be run in openshift-ci i.e in an environment where creating new namespaces is not allowed
* Adding detection of OPENSHIFT_CI env var and in this case run the smoke tests in the current namespace
* Adds resources cleanup on startup to allow re-run of smoke tests in the local project
* Adds a step that cleans the test resources between launches, the existing one was not cleaning postgres resources and it was failing.
* Adds while loop waiting in many @then steps to avoir unecessary wait time
* Re order some events waiting that were breaking the tests

/assign @jitendar-singh  
can you PTAL @jitendar-singh and @adambkaplan ?
The next coming PR would contain the additional smoke test verifying JENKINS_PASSWORD env var usage.

### DIFF
--- a/scripts/test-jenkins-template-install.sh
+++ b/scripts/test-jenkins-template-install.sh
@@ -17,12 +17,23 @@ ${PYTHON_VENV_DIR}/bin/pip install --upgrade setuptools
 ${PYTHON_VENV_DIR}/bin/pip install --upgrade pip
 # -- Generating a new namespace name
 echo "test-namespace-$(uuidgen | tr '[:upper:]' '[:lower:]' | head -c 8)" > ${OUTPUT_DIR}/test-namespace
+if [[ $OPENSHIFT_CI == true ]]; then
+    # openshift-ci will not allow to create different namespaces, so will do all in the same namespace.
+    oc project -q > ${OUTPUT_DIR}/test-namespace
+fi
 export TEST_NAMESPACE=$(cat ${OUTPUT_DIR}/test-namespace)
-echo "Assigning value to variable TEST_NAMESPACE:"${TEST_NAMESPACE}
+echo "Assigning value to variable TEST_NAMESPACE=${TEST_NAMESPACE}"
 # -- Do clean up & create namespace
 echo "Starting cleanup"
-kubectl delete namespace ${TEST_NAMESPACE} --timeout=45s --wait
-kubectl create namespace ${TEST_NAMESPACE}
+if [[ $OPENSHIFT_CI == true ]]; then
+    oc delete all,rolebindings.authorization.openshift.io,bc,cm,is,pvc,sa,secret -l app=jenkins-ephemeral
+    oc delete all,rolebindings.authorization.openshift.io,bc,cm,is,pvc,sa,secret -l app=jenkins-persistent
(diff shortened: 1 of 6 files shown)


scripts/test-jenkins-template-install.sh
```diff
@@ -17,12 +17,23 @@ ${PYTHON_VENV_DIR}/bin/pip install --upgrade setuptools
 ${PYTHON_VENV_DIR}/bin/pip install --upgrade pip
 # -- Generating a new namespace name
 echo "test-namespace-$(uuidgen | tr '[:upper:]' '[:lower:]' | head -c 8)" > ${OUTPUT_DIR}/test-namespace
+if [[ $OPENSHIFT_CI == true ]]; then
+    # openshift-ci will not allow to create different namespaces, so will do all in the same namespace.
+    oc project -q > ${OUTPUT_DIR}/test-namespace
+fi
```
adambkaplan, line 23: Odd that we can't call `oc new-project` against a CI cluster by default...

scripts/test-jenkins-template-install.sh
```diff
@@ -35,7 +46,9 @@ oc project ${TEST_NAMESPACE}
 # -- Trigger the test
 echo "Starting local Jenkins instance"
 ${PYTHON_VENV_DIR}/bin/pip install -q -r smoke/requirements.txt
-echo "Running smoke tests"
-TEST_NAMESPACE=${TEST_NAMESPACE}
-${PYTHON_VENV_DIR}/bin/behave --junit --junit-directory ${TEST_SMOKE_OUTPUT_DIR} --no-capture --no-capture-stderr smoke/features
-echo "Logs collected at "${TEST_SMOKE_OUTPUT_DIR}
\ No newline at end of file
+echo "Running smoke tests in namespace with TEST_NAMESPACE=${TEST_NAMESPACE}"
+echo "Logs will be collected in "${TEST_SMOKE_OUTPUT_DIR}
```
adambkaplan, line 50: Line duplicated below
```suggestion

```

smoke/features/steps/steps.py
```diff
@@ -324,26 +324,29 @@ def verifyRoute(context):
...
 def verifyJenkinsBuildStatus(context):
-    buildState = oc.get_resource_info_by_jsonpath('build','openshift-jee-sample-1',current_project,json_path='{.status.phase}')
-    if not 'Complete' in buildState:
-        raise AssertionError
-    else:
-        print("Build openshift-jee-sample-1-deploy status:{buildState}")
+    verify_status('build', 'openshift-jee-sample-1', 2, 100, 'Complete', '{.status.phase}')
+
 
+def verify_status(object_type, object_name, sleep_delay, sleep_count, expected_status,json_path='{.status.phase}'):
```
adambkaplan, line 338: This function effectively polls the object's status to check a specific condition. I'd rename the parameters to be a bit clearer:

```suggestion
def verify_status(object_type, object_name, poll_interval_seconds, max_retries, expected_status,json_path='{.status.phase}'):
```

adambkaplan (changes requested):
A few bits of cleanup, otherwise looks good.

PR #1356, opened 2021-12-06
refactoring smoke test

jitendar-singh:
- Cleaning up the resources before running the jenkins-ephemeral.feature as it was causing the test scenario to fail
- Removed the plugin feature file and steps as its not yet implemented
- Updated the README.Md


smoke/features/jenkins-ephemeral.feature
```diff
@@ -5,6 +5,15 @@ Feature: Deploy Jenkins on openshift using template based install
...
     Given Project [TEST_NAMESPACE] is used
+    Then we delete deploymentconfig.apps.openshift.io "jenkins"
+        And  we delete route.route.openshift.io "jenkins"
+        And  delete configmap "jenkins-trusted-ca-bundle"
+        And  delete serviceaccount "jenkins"
+        And  delete rolebinding.authorization.openshift.io "jenkins_edit"
+        And  delete service "jenkins-jnlp"
+        And  delete service "jenkins"
+        And delete all deploymentconfig
+        And delete all remaining test resources
```
akram, line 16: What about creating a single @step that does that in one shot?

akram (commented):
Except my single comment, that looks good to me.

smoke/features/jenkins-ephemeral.feature
jitendar-singh: All the step definitions for the above are already present so didnt create a new one, as we use the same before starting the persistent template test

Issue #924, opened 2019-09-13
Container image should include Jenkins version as a label


jlebon:
So that a e.g. `skopeo inspect` is enough to know what the version is.

waveywaves:
Hello @jlebon , 
Thanks for opening this issue. 

This is a good idea. 
/assign waveywaves

akram:
Hi @waveywaves 
as discussed on slack, that's good idea. And we will have a specific label since we didn't find a generic one amongst openshift images.

akram:
This is fixed here https://github.com/openshift/jenkins/pull/925

/close

PR #1374, opened 2022-01-27
JKNS-269: Pre-compute bundle-plugins.txt

akram:
This PR has 2 goals:

- the primary goal is to avoid the strong requirement of a midstream repository while running automated productisation (CPaaS) by providing a convenient way to generate the comprehensive list of plugins used by the Jenkins image. 

- the secondary goal is to remove supply chain errors which could rarely happen in case of a commit of the `base-plugins.txt` only, and in the meanwhile, if a transitive dependency changes, we could end-up release a different version compared to what we have test in local development environment or in CI.

Generating the comprehensive list of plugins at build time allows its strict usage during the CI build and the CPaaS build. 

Before this PR, the comprehensive list of plugins was provided manually from the CI build logs to the ART team.
By having the list forcibly generated at commit time (a git pre-commit hook can now be added), and being checked at CI time, ensures that the commited list of plugins is the exactly one that will be tested during CI https://github.com/openshift/release/pull/25817 ; and then, it can be used as-is in CPaaS as well without having to re-compute the list.
If a drift between the committed list of plugins and its locally computed comprehensive list is detected, the CI build fails by asking the developer to re-compute the list locally and push it to the code repository.


akram:
/assign @gabemontero 

this will be used by this: https://github.com/openshift/release/pull/25817
Can PTAL @gabemontero ?

gabemontero:
@akram - could you update this PR's description with more detail on what this PR is trying to achieve.  And even after having read the BZ text, I feel like maybe we ought to discuss this one voice to voice (next office hours maybe).  I'm not sure I fully understand the pros and cons, the risk / reward, with what is going on here.

But take a stab at providing detail in the PR description and we'll go from there.

akram:
@gabemontero I updated the description. We can discuss that during office hours sure.

gabemontero (changes requested):
I believe https://github.com/openshift/jenkins/blob/master/CONTRIBUTING_TO_OPENSHIFT_JENKINS_IMAGE_AND_PLUGINS.md should be updated @akram to capture some essence of what you have in this PR's description

I also suspect there is a new update to our process wrt what a contributor must do (i.e. when in the existing flow to execute your new make target) that should be described there.

Tangentially, I suspect will start needing to document ART vs. CPass diffs in our release processes.  Some of that could certainly be in a separate PR.  But I mention it since you mention the CPaas vs. ART flow in the PR's description.

gabemontero:
I also wonder @akram @adambkaplan if this PR should be tracked via a Jira instead of a BZ.

akram:
> I believe https://github.com/openshift/jenkins/blob/master/CONTRIBUTING_TO_OPENSHIFT_JENKINS_IMAGE_AND_PLUGINS.md should be updated @akram to capture some essence of what you have in this PR's description
> 
> I also suspect there is a new update to our process wrt what a contributor must do (i.e. when in the existing flow to execute your new make target) that should be described there.
> 
> Tangentially, I suspect will start needing to document ART vs. CPass diffs in our release processes. Some of that could certainly be in a separate PR. But I mention it since you mention the CPaas vs. ART flow in the PR's description.

I think https://github.com/openshift/jenkins/pull/1382 adresses the concern about developer onboarding;

gabemontero:
/label docs-approved

a no-docs story

gabemontero:
/label px-approved

a no px story

gabemontero (changes requested):

hack/check-plugins-supply-chain-change.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
akram: wow, it was running from the first shot.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -329,6 +330,10 @@ function resolveDependencies() {
 }
 
 function bundledPlugins() {
+    echo "Checking JENKINS_WAR=$JENKINS_WAR"
+    if [ ! -f $JENKINS_WAR ]; then
+	curl -L -C - -o $JENKINS_WAR https://get.jenkins.io/war-stable/latest/jenkins.war
```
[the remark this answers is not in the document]
akram: you are true. It seems that the analysis of the `JENKINS_WAR` is intended determine the list of detached-plugins in the war. i.e. those whose functionality used to be part of Jenkins core but was moved into a plugin

hack/check-plugins-supply-chain-change.sh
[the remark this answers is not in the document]
gabemontero: :-)

2/contrib/jenkins/install-plugins.sh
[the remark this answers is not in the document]
gabemontero: ok cool I'll wait for you to update this then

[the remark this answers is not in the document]
akram: I also ended up externalizing the jenkins version in `jenkins-version.txt` . 
Now, to update the Jenkins version it should be sufficient to update the version in `2/contrib/openshift/jenkins-version.txt`

akram:
@gabemontero should be good now.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -329,6 +330,11 @@ function resolveDependencies() {
 }
 
 function bundledPlugins() {
+    echo "Checking JENKINS_WAR=$JENKINS_WAR"
+    if [ ! -f $JENKINS_WAR ]; then
+        jenkins_version=$(cat $PWD/2/contrib/openshift/jenkins-version.txt )
```
gabemontero, line 335: add an echo that displays the value of `jenkins_version`

2/contrib/jenkins/install-jenkins-core-plugins.sh
```diff
@@ -2,6 +2,8 @@
 
 set -o pipefail
 
+
+jenkins_version=$(cat /opt/openshift/jenkins-version.txt)
```
gabemontero, line 6: add an echo that display the value of `jenkins_version`

gabemontero (changes requested):
just a couple of minor asks @akram

2/contrib/jenkins/install-jenkins-core-plugins.sh
akram: done

2/contrib/jenkins/install-plugins.sh
akram: done

2/contrib/jenkins/install-jenkins-core-plugins.sh
gabemontero: ok now see

```
2022-02-02T16:54:45.838915482Z Jenkins version: 2.319.2
```

in the image build log.

2/contrib/jenkins/install-plugins.sh
gabemontero: ok now see 

```
2022-02-02T16:54:45.838915482Z Jenkins version: 2.319.2
```

in the image build log.

and https://get.jenkins.io/war-stable/2.319.2/ rendors

gabemontero:
install fail on sync plugin e2e

/retest

jlebon:
I think this PR caused a regression on https://github.com/openshift/jenkins/issues/924. Now the label on the image is `"io.jenkins.version": "latest"`, which isn't very useful. Is there a way to go back to the previous behaviour?

PR #1384, opened 2022-02-03
JNKS-269: Use the generated bundle-plugins.txt file while building the image

akram:
Now that the bundle-plugins.txt is correctly generated and checked using CI, this PRs enables its strict usage while building the images.

### DIFF
--- a/2/Dockerfile.localdev
+++ b/2/Dockerfile.localdev
@@ -62,7 +62,7 @@ COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 ADD release.version /tmp/release.version
 
-RUN /usr/local/bin/install-jenkins-core-plugins.sh /opt/openshift/base-plugins.txt "--disablerepo=rhel-server-extras --disablerepo=rhel-server --disablerepo=rhel-fast-datapath  --disablerepo=rhel-server-optional --disablerepo=rhel-server-ose --disablerepo=rhel-server-rhscl" && \
+RUN /usr/local/bin/install-jenkins-core-plugins.sh /opt/openshift/bundle-plugins.txt "--disablerepo=rhel-server-extras --disablerepo=rhel-server --disablerepo=rhel-fast-datapath  --disablerepo=rhel-server-optional --disablerepo=rhel-server-ose --disablerepo=rhel-server-rhscl" && \
     rmdir /var/log/jenkins && \
     chmod -R 775 /etc/alternatives && \
     chmod -R 775 /var/lib/alternatives && \
--- a/2/Dockerfile.rhel7
+++ b/2/Dockerfile.rhel7
@@ -73,7 +73,7 @@ COPY ./contrib/jenkins /usr/local/bin
 ADD ./contrib/s2i /usr/libexec/s2i
 ADD release.version /tmp/release.version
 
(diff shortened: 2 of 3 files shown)


akram:
AWS Terraform flakes
/retest

gabemontero:
/label docs-approved

no docs impact

gabemontero:
/retitle JNKS-269 Use the generated bundle-plugins.txt file while building the image

gabemontero:
/retitle JNKS-269: Use the generated bundle-plugins.txt file while building the image

gabemontero:
/assign @jitendar-singh 

to apply the qe approved label, where he may
want to make any adjustments to smoke tests or perform any manual testing

gabemontero:
/retitle JNKS-269: Use the generated bundle-plugins.txt file while building the image

gabemontero:
hey @akram looking at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_jenkins/1384/pull-ci-openshift-jenkins-master-images/1489174802982768640/artifacts/build-logs/jenkins.log though the net result I'm sure is usually going to be the same, the way the curent script currently works, it is still downloading dependencies from inspecting the pom.xml of the 

For example:

`2022-02-03T09:43:33.794397523Z Downloading dependency plugin branch-api version 2.0.11 that has yet to be installed`

I think to truly "lock in", we need a new variant on the install plugins script that does not examine dependencies, and just downloads what is in the list.

So we then use the existing script to generate the bundle-plugins.txt file .... essentially we've taken our snapshot,
and then in the Dockerfiles we call a new script that just installs what is in the file.

It might be as simple as augmenting the existing script so that if it is called with the correct parameter, it bypasses the call to `resolveDependencies` here at https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/install-plugins.sh#L190

Give that a go and let's see what happens.

akram:
@gabemontero let's try to do it in a second step. Once, the whole chained has been proved enough to not hiding some corner cases we can remove this additional check.
In the meanwhile, we can add it in the story as another task. wdyt ?

akram:
Here it is https://issues.redhat.com/browse/JKNS-270 .
I kept it unassigned, as it can also be a first issue task for someone who wants to become familiar with that parts; as fixing it touches ci and plugin dependencies

akram:
Technically, we are locking right now.  This is ensure by `ci/prow/jenkins-check-plugins-supply-chain-change`. Just as you are mentioning, the build of the image is however re-calculating the plugins dependencies.
In case of a difference, the `make plugins-list` job would fail preventing to merge something different.

gabemontero:
> @gabemontero Technically, we are locking right now. This is ensured by `ci/prow/jenkins-check-plugins-supply-chain-change`. Just as you are mentioning, the build of the image is however re-calculating the plugins dependencies. In case of a difference, the `make plugins-list` job would fail preventing to merge something different.

Ah forgot  that validation you recently added would catch this.  Yeah OK I'd still like to discussion the task vs. story and making sure JNKS-270 gets properly groomed in the next appropriate team meeting, but yeah I'm good with this getting lgtm'ed

If you have not already @akram move JNKS-269 to qe review on the board and let's get @jitendar-singh to review and decide what if anything else is needed for him to apply the qe approved label.

akram:
/assign @jitendar-singh 

Jitendar can you please add the qe-approved label ?

Issue #1342, opened 2021-11-15
pipeline-utility-steps plugin removed from base plugins


chriscarpenter12:
In the process of updating our Jenkins install from 4.7 -> 4.8 I noticed a few plugins originally in the base plugins list were removed we relied on, specifically the `pipeline-utility-steps` plugin.

According to #1294 it removed these plugins entirely:

```txt
parameterized-trigger:2.36
pipeline-build-step:2.12
pipeline-github-lib:1.0
pipeline-input-step:2.11
pipeline-model-api:1.7.2
pipeline-model-definition:1.6.0
pipeline-utility-steps:2.5.0
workflow-aggregator:2.6
workflow-cps-global-lib:2.15
workflow-cps:2.87
workflow-multibranch:2.22
workflow-remote-loader:1.5
workflow-support:3.6
```

Assuming some of these plugins are pulled in as dependencies of others our updated 4.8 installation is missing the following which we previously had:

```text
parameterized-trigger
pipeline-github-lib
pipeline-utility-steps
workflow-remote-loader
```

And it looks like release/4.8 branch was backported with [this commit](https://github.com/openshift/jenkins/commit/bcf60dec55daf4b0dee2f932c3a0f5b87f460bf5#diff-2a93d798ca112ffa17e2d1d713ef42d02dab81e4704b8a783d9b03635b6470b3).

Is this an oversight or were these removed because of a specific reason?

PR #1387, opened 2022-02-15
Revert used plugins by the OCS team

akram:
This PR reintroduces the plugin `pipeline-utility-steps` which is used by the OCS team and also some other comminity users as reported here https://github.com/openshift/jenkins/issues/1342


akram:
/assign @gabemontero 
Gabe, can you PTAL ?

gabemontero:
@akram - please add "Fixes #1342 " into the description so that issue is closed out when this merges

Also, can you elaborate more on the relationship with https://bugzilla.redhat.com/show_bug.cgi?id=2052568  ?

Based on the answer to that, it may make sense to re-open that BZ, associate it with this PR, and backport to the supported z streams.

gabemontero:
> Good point @gabemontero , I was hesitant in doing this, because technically the cause of the BZ was really the caffeine plugin missing. But, Oded (the reporter) found it by chance while investigating the bug more globally. When the caffeine plugin fix was done, the sync plugin was starting well. But, Oded pipelines was failing because of a `mavenLoad` function that coudn't be found, and this is brought by `pipeline-utility-steps`
> 
> I am OK re-opening it, but maybe changing the title of the BZ. And backporting. Or we can create another BZ and do the backport in all cases.
> 
> wdyt?

OK I think a new BZ should be opened for adding pipeline-utility-steps back in so things like `mavenLoad` are available.  Then we can see about backporting, since https://github.com/openshift/jenkins/issues/1342 notes they discovered the issue in a 4.7 to 4.8 upgrade.  Go ahead an open the BZ for that please @akram and associate with this PR.

I'll then lgtm.

As to the caffeine plugin piece, per my response on the separate email thread with Oded and team I believe that is related to plugins the customer has manually upgraded when they run Jenkins on PVs. then upgrade OpenShift, and what we currently do wrt "leaving the plugin alone", per https://docs.openshift.com/container-platform/4.9/openshift_images/using_images/images-other-jenkins.html#images-other-jenkins-env-var_images-other-jenkins and the `OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS` parameter.  I suspect we may need to pivot on the choice we made back in 3.x and default our behavior to what occurs when that parameter is true.

Most likely we should discuss this in an upcoming architecture meeting, and if everyone agrees with my idea here, then get the appropriate tracking items opened, work started, KCS articles written, etc.

akram:
fair enough, here is the bz then https://bugzilla.redhat.com/show_bug.cgi?id=2054822

akram:
ART request: https://issues.redhat.com/browse/ART-3798

akram:
To be cherry-picked, we will have to cherr-pick also: https://github.com/openshift/jenkins/pull/1374 
Otherwise, we have to make the changes manually

gabemontero:
> To be cherry-picked, we will have to cherr-pick also: #1374 Otherwise, we have to make the changes manually

I am not ready to cherry-pick #1374 at this time ... you'll have to manual pick to 4.10.  After that, you should be able to use the bot to pick the changes from 4.10 back to 4.9, etc.

PR #1390, opened 2022-02-17
removing reference to plugin class

jitendar-singh:
- Removing reference to plugin class which was causing the smoke test not to run


jitendar-singh:
/label docs-approved

QE only

jitendar-singh:
/label px-approved

QE only

PR #1394, opened 2022-02-23
Sort final plugins list alphabetically in `make plugins-list`

akram:
To have a consistent plugins list comparison, we now sort the plugins list alphabetically

### DIFF
--- a/2/contrib/jenkins/install-plugins.sh
+++ b/2/contrib/jenkins/install-plugins.sh
@@ -367,10 +367,12 @@ function versionFromPlugin() {
 
 function installedPlugins() {
     echo "# Generated file: To generate this file run 'make plugins-list' before pushing your changes" > $BUNDLE_PLUGINS
+    temp_bundle=$(mktemp)
     for f in "$REF_DIR"/*.jpi; do
         echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")"
-        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $BUNDLE_PLUGINS
+        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $temp_bundle
     done
+    sort -u $temp_bundle >> $BUNDLE_PLUGINS
 }
 
 function jenkinsMajorMinorVersion() {


2/contrib/jenkins/install-plugins.sh
```diff
@@ -367,10 +367,12 @@ function versionFromPlugin() {
 
 function installedPlugins() {
     echo "# Generated file: To generate this file run 'make plugins-list' before pushing your changes" > $BUNDLE_PLUGINS
+    temp_bundle=$(mktemp)
     for f in "$REF_DIR"/*.jpi; do
         echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")"
-        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $BUNDLE_PLUGINS
+        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $temp_bundle
     done
+    sort $temp_bundle >> $BUNDLE_PLUGINS
```
gabemontero, line 375: @akram given the discrepancy we are seeing in our local ubuntu / fedora systems, I want to confirm whether or not the supply chain check in this PR is in fact using you change here.

So add a `echo "generated plugin list $temp_bundle before sort"` followed by a `cat $temp_bundle` before you call to `sort` here.

Then after the call to `sort` add a `echo "generated plugin list $BUNDLE_PLUGINS after sort"` followed by a `cat $BUNDLE_PLUGINS` file

And then we can review the logs of the supply chain e2e

I am wondering if that e2e is *NOT* picking up these changes and is using the master branch copy, though maybe it is with the `clone: true` field you have at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/jenkins/openshift-jenkins-master.yaml#L81-L82

Depending on the result, I'm wondering if we merge this now, then after this merges, we'll need a one time `make plugins-list` update to get the ordering correct

akram: done

akram:
@gabemontero I see no differences between before and after :-)
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_jenkins/1394/pull-ci-openshift-jenkins-master-jenkins-check-plugins-supply-chain-change/1498640911049953280/artifacts/test/build-log.txt

gabemontero:
On Tue, Mar 1, 2022 at 11:02 AM Akram Ben Aissi ***@***.***>
wrote:

> @gabemontero <https://github.com/gabemontero> I see no differences
> between before and after :-)
>
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_jenkins/1394/pull-ci-openshift-jenkins-master-jenkins-check-plugins-supply-chain-change/1498640911049953280/artifacts/test/build-log.txt
>
yep which is what I expected.  It is just that running the sort on my
system appears to produce a different result

I will confirm that and if so will experiment with specifying different
args on the sort

2/contrib/jenkins/install-plugins.sh
```diff
@@ -367,10 +367,17 @@ function versionFromPlugin() {
...
     for f in "$REF_DIR"/*.jpi; do
         echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")"
-        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $BUNDLE_PLUGINS
+        echo "$(basename "$f" | sed -e 's/\.jpi//'):$(get_plugin_version "$f")" >> $temp_bundle
     done
+    echo "generated plugin list in $temp_bundle before sort"
+    cat $temp_bundle
+
+    echo "generated plugin list $BUNDLE_PLUGINS after sort"
+    sort $temp_bundle >> $BUNDLE_PLUGINS
```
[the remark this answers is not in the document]
akram: ok @gabemontero maybe a good news...so `sort -d` on my system creates a different output. Where `sort -u` keeps the file unchanged.

So, I will update in that direction and set `sort -u` to keep us consistent.

PR #1414, opened 2022-03-23
JNKS-254: update sample imagestreams with latest 4.10 image (using specific image tag reference)

gabemontero:
In order to enable samples operator testing where we disable the use of the OCP payload, and while the CPaas images are not yet ready, we can update the sample imagestreams with our latest 4.10 image up on the TBR.

We'll then go through the openshift/library flow, and in the subsequent samples operator PR, where we disable use of the payload images, these images will be used (instead of the super old out of date 4.0 or 4.6 ones), and we can properly validate via the latest jenkins e2e's, which the 4.10 image should satisfy.

PR #1421, opened 2022-03-29
Bug 2066019: set necessary JVM args to allow jenkins JVM to come up on a FIPS node

gabemontero:
without defaulting so, none of our Jenkins related JVMs will come up on a FIPS enabled node, failing with something like

```
SEVERE: FIPS mode: only SunJSSE TrustManagers may be used
java.security.KeyManagementException: FIPS mode: only SunJSSE TrustManagers may be used
    at java.base/sun.security.ssl.SSLContextImpl.chooseTrustManager(SSLContextImpl.java:133)
    at java.base/sun.security.ssl.SSLContextImpl.engineInit(SSLContextImpl.java:95)
    at java.base/javax.net.ssl.SSLContext.init(SSLContext.java:297)

```

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -640,7 +640,11 @@ if [ -n "$https_proxy" ]; then
   JAVA_HTTP_PROXY_OPTIONS="$JAVA_HTTP_PROXY_OPTIONS $proxy"
 fi
 
-JENKINS_JAVA_OPTIONS="$JENKINS_JAVA_OPTIONS $JAVA_HTTP_PROXY_OPTIONS"
+if [[ -z "${JAVA_FIPS_OPTIONS}" ]]; then
+  JAVA_FIPS_OPTIONS="-Dcom.redhat.fips=false"
+fi
+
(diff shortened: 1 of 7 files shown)


gabemontero:
fyi @jitendar-singh I launched a fips cluster with this PR via cluster-bot and I was able to execute the test plan I spelled out in the bugzilla

gabemontero:
also @akram I think there may be something wrong with one of the sample DBs that the client plugin e2e's are using

I saw it locally in my cluster-bot fips cluster as well

trying it out in #1414 as well, see if the results compare

gabemontero:
heh ... client e2e plugin is now passing  here @akram 

and it is passing for me locally off of master branch now

perhaps a momentary hiccup with some upstream service ??

rolfedh:
> /lgtm
> 
> ok, that's a good news @gabemontero . I have seen your message on slack only after reading the PR's comments. I think we also need to update @rolfedh to have a proper documentation update

I'm tracking the docs work for this with https://issues.redhat.com/browse/RHDEVDOCS-3931

PR #1422, opened 2022-03-31
Bug 2069498:  [release-4.9] update bundle plugins

akram:
Update bundle-plugins.txt and allow supply chain check 
required by https://github.com/openshift/release/pull/27254

/assign @adambkaplan 
/assign @gabemontero


adambkaplan (changes requested):
/lgtm cancel

There appears to be several plugin downgrades in the bundle-plugins.txt file. Is there an explanation for this?

adambkaplan:
/hold

Due to downgrades in the bundle-plugins.

akram:
@adambkaplan it is really weird, I had to force some plugins now to get the similar results. It seems that there is some fine dependencies tuning that we need to double check because they are pulled transitevly and older versions get preferred.

This should be an additional check that we need to pay attention to in the future.
So well catch, but I fear that this may have been happened in the past before we had the `bundle-plugins.txt` computation done.

akram:
flake
```
  exec failed: error running /usr/bin/kubectl --server=https://api.ci-op-727i6wd5-e66a4.origin-ci-int-aws.dev.rhcloud.com:6443 --kubeconfig=/tmp/kubeconfig-3195469076 --namespace=e2e-test-cli-idling-ns5c4 exec execpodxc8vz -- /bin/sh -x -c curl --max-time 120 -s -g http://172.30.214.253:8675/echo?msg=It+is+time+to+TCP.:
```

adambkaplan (approved):
/lgtm

Agree that in the future we need to detect if we are downgrading a plugin.
Noted this in https://issues.redhat.com/browse/JKNS-290

coreydaley:
/assign @jitendar-singh
Could you add the cherry pick approved label if appropriate?

PR #1438, opened 2022-04-25
JNKS-287: cannot use specific image refs with okd; quay.io prunes too often

gabemontero:
long term if we need this for okd we'll push permanent tags to quay.io/redhat-developer


gabemontero:
/label qe-approved

no impact

gabemontero:
/label docs-approved

no impact

Issue #1224, opened 2021-02-10
Warning or Error?: _/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)_


mbecca:
Hello,

I have the following warning/error? when extending the base image of the slave: quay.io/openshift/origin-jenkins-agent-base:4.8.0

_/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)_

Complete docker build log:

```
Sending build context to Docker daemon  9.499MB
Step 1/4 : FROM quay.io/openshift/origin-jenkins-agent-base:4.8.0
 ---> 9a95eabfa25a
Step 2/4 : LABEL io.k8s.description="Agente Slave de jenkins para el nodo maven con OpenJDK y NodeJS"       io.k8s.display-name="jenkins-maven-slave"       io.openshift.tags="jenkins,node maven, openjdk, nvm, nodejs, npm"       maintainer="Matías Beccaria mbecca@mecon.gov.ar"
 ---> Using cache
 ---> d4d76a12aec7
Step 3/4 : RUN mkdir -p /tmp/prueba
 ---> Running in f3e416b87375
/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
Removing intermediate container f3e416b87375
 ---> 9512ba54833e
Step 4/4 : ENTRYPOINT ["/usr/local/bin/run-jnlp-client"]
 ---> Running in 9d50ea2b215f
Removing intermediate container 9d50ea2b215f
 ---> f2353f4187ae
Successfully built f2353f4187ae
Successfully tagged slave:latest
```

My Dockefile (simplifided)

> FROM quay.io/openshift/origin-jenkins-agent-base:4.8.0
> 
> LABEL io.k8s.description="Agente Slave de jenkins para el nodo maven con OpenJDK y NodeJS" \
>       io.k8s.display-name="jenkins-maven-slave" \
>       io.openshift.tags="jenkins,node maven, openjdk, nvm, nodejs, npm" \
>       maintainer="Matías Beccaria"
> 
> RUN mkdir -p /tmp/prueba
> 
> ENTRYPOINT ["/usr/local/bin/run-jnlp-client"]

Thanks in advance!

M.

PR #1451, opened 2022-05-24
Bug 2096207: Add missing glibc-langpack-en for specified locale

cobexer:


cobexer:
See issues https://github.com/openshift/jenkins/issues/960 and https://github.com/openshift/jenkins/issues/1224

@akram maybe you can have a look?

akram:
/retitle Bug 2096207: Add missing glibc-langpack-en for specified locale

sll552:
This fix would be very welcome, but it seems that it didn't get merged in any release branch, and therefore is not available in an actual container. Is there any reason for this? @akram

PR #1452, opened 2022-05-24
asdf

gabemontero:


gabemontero:
ok sanity check with regard to being able to vet plugin updates vetted @akram @coreydaley 

I do think we are on a path to resolution sooner rather than later for getting our plugin e2e's working with https://github.com/openshift/release/pull/28845 followed by updates to the plugin repo Dockerfiles so that the sidecar e2e's work.

Issue #1446, opened 2022-05-11
Failed to download plugin with provided INSTALL_PLUGINS


gloria-zmm:
Hi,
We are using OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS=true and INSTALL_PLUGINS={list of plugins} to install plugins during jenkins start up. We encountered this issue:
```
Downloading plugins...
Using provided plugin: ace-editor
 > ace-editor has no dependencies
Downloading plugin: analysis-model-api from https://updates.jenkins.io/download/plugins/analysis-model-api/10.9.1/analysis-model-api.hpi
Downloading plugin: analysis-model-api-plugin from https://updates.jenkins.io/download/plugins/analysis-model-api-plugin/10.9.1/analysis-model-api-plugin.hpi
Failed to download plugin: analysis-model-api or analysis-model-api-plugin
```
It happened at a very high frequency. Usually after several times of restart it can run successfully finally.
The image we are using is :
registry.redhat.io/openshift4/ose-jenkins:v4.9.0-202202080146.p0.ga71a115.assembly.stream

Please suggest us what to do to improve the efficiency.

chengjun0124:
Hi
I suffered the same issue. The issue is intermittent.

chengjun0124:
The issue is caused by network issue. When jenkins tries to download plugin, the network is not ready.  

It is not an issue from openshift jenkins. Openshift jenkins uses parameter `--retry` of curl to make sure a successful download:
`curl $curl_options --connect-timeout "${CURL_CONNECTION_TIMEOUT:-20}" --retry "${CURL_RETRY:-5}" --retry-delay "${CURL_RETRY_DELAY:-0}" --retry-max-time "${CURL_RETRY_MAX_TIME:-60}" -s -f -L "$url" -o "$jpi"`.
However `--retry` only retries for HTTP 408, 429, 500, 502, 503 or 504 response code.

From fault-tolerance perspective, I would be better if openshift jenkins were to retry download against other network issues, like `curl: (7) couldn't connect to host`.

akram:
@chengjun0124 could you please propose a fix through a PR? 
Network flakes and issues can be of a different types, so we still have choice on what to test. And testing on every iteration of download may be unproductive. what do you think?

ggareth:
The issue I was having was not with a flakey network, but with mirrors not being complete. 
Have proposed a PR which reduces (makes it work for me) the effects

coreydaley:
[...]

PR #1461, opened 2022-07-13
Update install-plugins.sh

ggareth:
This adds an ability to retry downloads of plugins when mirrors are incomplete/returning 404s
This doesn't unfortunately fix when a file only downloads 16mb of a 40mb plugin, but at least reduces some of the errors

This is in reference to #1446


ggareth (commented):
/assign @sbose78

coreydaley:
/hold
The `xmlstarlet` package not being available is a known issue and we are working to resolve it, putting a hold on this pull request until it is fixed, no sense in consuming resources re-running tests that have to chance to pass.

coreydaley:
/assign @akram 
Can you please take a look at this and see if it looks O.K. to you?

PR #1465, opened 2022-08-10
Revert xmlstarlet usage

coreydaley:
Revert xmlstarlet usage so that we can merge/backport multiple CVE fixes
Tracking in [Bug 2117742](https://bugzilla.redhat.com/show_bug.cgi?id=2117742)


adambkaplan (approved):
/lgtm

I would use the BZ to track when xmlstarlet is restored. Alternatively, if you discover that xmlstarlet isn't needed after all, then you can simply close the BZ.

coreydaley:
I will open another bugzilla to track restoring it, I thought I might need these bugzillas for the backports.

PR #1473, opened 2022-08-29
OCPBUGS-655: Update blueocean-autofavorite to 1.2.5

coreydaley:
Update to newer version which does not use the groovy-sandbox library


coreydaley:
/assign @akram 
for lgtm

PR #1480, opened 2022-09-12
[release-4.10] OCPBUGS-1037: Install glibc language package to remove `setLocale` warning

divyansh42:
Whenever jenkins jobs and commands are run, logs are filled with below warning:
 
 ```
 # podman run -it --entrypoint /bin/bash registry.redhat.io/openshift4/ose-jenkins-agent-base:v4.10.0-202207291908.p0.gc5b7159.assembly.stream 
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8) <---
```

Installing the glibc language package solves the problem.

Bug https://issues.redhat.com/browse/OCPBUGS-1037

Signed-off-by: divyansh42 <diagrawa@redhat.com>


coreydaley:
I think that we should keep the images as they were, pointing to 4.10 and see if you can just install the glibc language package and see if that fixes the issue.

coreydaley:
/assign @jitendar-singh 
for cherry-pick-approved
/assign @akram 
for lgtm

PR #1482, opened 2022-09-14
Fix install-plugins.sh to not override locked plugin versions

coreydaley:
- Do no override locked plugin versions
 - Verify that locked plugin versions exist in the bundle-plugins.txt file


coreydaley:
@akram All tests are passing, please take a look and let me know if you have any questions about what the issue was.

akram (commented):
is there any reason to change the lockfile from a directory to a file?
I don't know the historical reason of having a directory, but, if we can minimize changes to this old code, that would be more reinsuring.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -172,9 +173,8 @@ function download() {
     plugin="$1"
     version="${2:-latest}"
     ignoreLockFile="${3:-}"
-    lock="$(getLockFile "$plugin")"
 
-    if [[ $ignoreLockFile ]] || mkdir "$lock" &>/dev/null; then
+    if [[ $ignoreLockFile ]] || ! test -f $(getLockFile $plugin); then
```
akram, line 177: is there any reason to change the lockfile from a directory to a file?
I don't know the historical reason of having a directory, but, if we can minimize changes to this old code, that would be more reinsuring.

2/contrib/jenkins/install-plugins.sh
```diff
@@ -194,6 +194,13 @@ function download() {
         fi
 
         resolveDependencies "$plugin"
+    else
+        lockFile=$(getLockFile "$plugin")
+        lockedVersion=$(cat $lockFile)
+        echo "Plugin $plugin locked to version $lockedVersion, ignoring, requested version $version"
+        if versionLT "${lockedVersion}" "${version}"; then
```
akram, line 201: what will we do with the warning message?
and during CI, we will not check it for sure. Should we fail in case of warning ?

2/contrib/jenkins/install-plugins.sh
```diff
@@ -310,7 +317,7 @@ function resolveDependencies() {
             # download the dependence; passing "true" is needed for "download" to replace the existing dependency
             if versionLT "${versionInstalled}" "${minVersion}"; then
                 echo "Upgrading bundled dependency $d ($minVersion > $versionInstalled)"
-                download "$plugin" "$minVersion" "true"
+                download "$plugin" "$minVersion"
```
akram, line 320: Was the bug here? I don't see a signature change to the download function, however, it was used only with 2 arguments previously, the 2nd one being a boolean. no regression ?

2/contrib/jenkins/install-plugins.sh
coreydaley: I am storing the locked plugin version in the `<plugin>.lock` file for use when checking the versions that are being requested.

2/contrib/jenkins/install-plugins.sh
coreydaley: The problem is that sometimes the versions are like `2.6.5` and sometimes they switch to `1123.456.asdflasdf`, and the script does not account for `2.6.5` could be newer than `1123.456.asdflasdf`, so if we show the warnings when someone updates the `base-plugins.txt` and then runs the install script, they can sort it out manually and update the locked version if needed, which it probably is. But specific plugins we want locked to a newer version that has CVE fixes.

2/contrib/jenkins/install-plugins.sh
coreydaley: Since this was passing `true`, it was overriding the locked versions of plugins and potentially installed a different version, which may or may not be correct, or what we wanted.

PR #1484, opened 2022-09-16
[release-4.10] OCPBUGS-1440: fix install plugins script

coreydaley:


2/contrib/jenkins/install-plugins.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#! /bin/bash -eu
```
akram, line 1: is that a typo ?

coreydaley: Yep! Fixed.

PR #1504, opened 2022-10-17
Correctly identify the jenkins route

josephearl:
Support namespaces where Jenkins is not the first route (at my current client they also have an Argo route in the same namespace which comes first, and causes a warning on Jenkins startup that the URL is not correctly configured because the file just contains `https://` as the URL).


josephearl:
Sorry realised the code already did this, was looking at an old PR

PR #1507, opened 2022-10-19
[release-4.11] OCPBUGS-913: Add maven plugin dependency in base plugins

divyansh42:
Signed-off-by: divyansh42 <diagrawa@redhat.com>


coreydaley:
@divyansh42 I think you need to check on the JIRAs and make sure that they are linked properly. If there is an existing bug open, then we should manually create the pull request instead of usin the cherry-pick functionality so that it does not create a duplicate bug.

coreydaley:
/assign @jitendar-singh 
for cherry-pick-approved

PR #1526, opened 2022-11-25
[release-4.8] OCPBUGS-4131: Externalize jenkins-version

divyansh42:


2/contrib/openshift/jenkins-version.txt
```diff
@@ -0,0 +1 @@
+2.319.3
```
jkhelil, line 1: why dont we have same version as for 4.9, and >?
https://github.com/openshift/jenkins/blob/release-4.9/2/contrib/openshift/jenkins-version.txt

2/contrib/openshift/bundle-plugins.txt
```diff
@@ -0,0 +1,118 @@
...
+workflow-api:2.47
+workflow-basic-steps:2.20
+workflow-cps:2660.vb_c0412dc4e6d
+workflow-cps-global-lib:566.vd0a_a_3334a_555
+workflow-durable-task-step:2.35
+workflow-job:2.41
+workflow-multibranch:711.vdfef37cda_816
+workflow-scm-step:2.13
+workflow-step-api:622.vb_8e7c15b_c95a_
+workflow-support:3.8
```
jkhelil, line 118: You changed directly the bundle but We need to change the baseplugin file then generate the bundle with make plugin-list, Is that because the PR doesnt hold this mecanism, in this case, can we add it to 4.8 too please ?

divyansh42: Bundle mechanism was there earlier in 4.8, I added that to 4.8 in this PR.
That's why new file is created, and it is generated from base-plugins only.

2/contrib/openshift/jenkins-version.txt
divyansh42: I am planning to change jenkins version in other PR.
For this PR i have not changed jenkins version, just used what was there in 4.8 earlier https://github.com/openshift/jenkins/blob/release-4.8/2/Dockerfile.rhel8#L46

divyansh42:
/assign @coreydaley
for backport-risk-assessed

PR #1532, opened 2022-12-02
OCPBUGS-4520: Update jenkins master branch to use 4.13 base image

jkhelil:


divyansh42:
@jkhelil can we change the openshift cli image version in localdev Dockerfile also, same as we did in `rhel8` Dockerfile.

PR #1588, opened 2023-02-16
Bump jenkins image version to use 4.12

coreydaley:


openshift/imagestreams/jenkins-rhel.json
```diff
@@ -55,7 +55,7 @@
         },
         "from": {
           "kind": "DockerImage",
-          "name": "registry.redhat.io/ocp-tools-4/jenkins-rhel8:v4.11.0-1663086392.1666642635"
+          "name": "registry.redhat.io/ocp-tools-4/jenkins-rhel8:jenkins-rhel8:v4.12.0-1675704988"
```
divyansh42, line 58: is this a typo?

coreydaley: yes! good catch.

coreydaley: fixed in current push

gabemontero:
I'm going to try an summarize (but also detail) what I attempted to convey to @coreydaley in https://redhat-internal.slack.com/archives/C02258G4S79/p1676574278725019 and what you have to worry about here

1) docs at https://docs.openshift.com/container-platform/4.12/cicd/jenkins/important-changes-to-openshift-jenkins-images.html need to be updated
2) while deploying jenkins initially should be the same, you all will want to verify the following upgrade OCP scenario:
  a) deploy jenkins where the imagestream ref points to `registry.redhat.io/ocp-tools-4/jenkins-rhel8:v4.12` ... it will be OCP 4.12.x
  b) note the the precise sha of the image the jenkins pod is using 
  c) now, say it is a week later, and you all make a change to `registry.redhat.io/ocp-tools-4/jenkins-rhel8:v4.12`, what happens if you (as the customer) upgrade from OCP 4.12.x to OCP 4.12.x+2
  d) as part of the OCP upgrade, when the samples operator applies all the imagstreams and ****IF***** this causes a re-import of the imagestream, even if the image ref is still `registry.redhat.io/ocp-tools-4/jenkins-rhel8:v4.12`, and that fires the image change trigger in the jenkins deployment config, and the customer's jenkins instance is restarted, pulling in the newer version of `registry.redhat.io/ocp-tools-4/jenkins-rhel8:v4.12`, you all are good.  The customer will get the CVE update as part of their OCP upgrade.

But, if ^^ does not happen, the customer will have to manually update the jenkins imagestream via `oc import-image`.  There will undoubtedly be a ton of customer bugs coming in about this, complaining about changing behavior, why are the CVE updates not automatic as part of the OCP upgrade.

Now, if the undesirable outcome is what happens, one idea did occur to me since @coreydaley and I talked, that might help here, and get what you all what you want, as in not having to update samples operator / jenkins imagestreams on each CVE.

If you all update the sample operator to special case the jenkins image stream, and have it always issue the equivalent of an `oc import-image` after it applies the jenkins imagestream, you'll get those updates, and the same behavior obtained when we used sha specific images.

The `oc import-image` logic is already there (it is what sample operator uses when imagestream imports fail).  https://github.com/openshift/cluster-samples-operator/blob/master/pkg/stub/retry.go

And special casing jenkins in samples operator is nothing new (we used to special case jenkins imagestream when we were in the payload, patching in the payload image into the imagestream's ref).

coreydaley:
/assign @divyansh42 @mbharatk 
for lgtm

PR #1662, opened 2023-04-14
[JKNS-344] Second phase of self-verification process for Jenkins

apoorvajagtap:
The script `verify-jenkins-on-openshift.sh` takes the following inputs:
- Cluster's API Server URL
- OpenShift Username & Password
- Jenkins Image
- SHA of the relevant commit (This is requested at a later stage, let me know, if we should input this at the beginning itself (with other variables)).

& Automates the following tasks:
- Pulling the (provided) CPaas image, tagging & pushing it to `quay.io/pipeline-integrations`.
- Logging in to an OpenShift cluster, creating a project & deploying jenkins pod using the provided image.
- Once the pod is ready, executes [verify-jenkins.sh](https://github.com/openshift/jenkins/blob/master/scripts/verify-jenkins.sh) within the pod & redirects the output in a file on local machine.
- Checks for errors in the output file, if found, reports.

### DIFF
--- a/scripts/verify-jenkins-on-openshift.sh
+++ b/scripts/verify-jenkins-on-openshift.sh
@@ -43,24 +43,6 @@ vars(){
     QUAY_IMAGE="quay.io/pipeline-integrations/openshift-ose-jenkins:${COMMIT_SHA}"
 }
 
-# get_image_ready() 
-# pulls the brew image locally, tags & pushes it to quay.
-# 
-get_image_ready(){
-    printf "\n-------------------------------------------------------\n>INFO || Pulling the Jenkins Image\n"
-    IMAGE_ID=$(podman pull "${JENKINS_IMAGE}")
-    if [ -z "${IMAGE_ID}" ]
-    then
-        printf "\n>ERR || Image Pull Failed\n"
-        exit 1
-    else
-        podman tag "${JENKINS_IMAGE}" "${QUAY_IMAGE}"
-        printf "\n-------------------------------------------------------\n>INFO || Pushing the image to 'quay.io/pipeline-integrations'\n"
-        podman push "${QUAY_IMAGE}"
-        printf "\n-------------------------------------------------------\n>INFO || Image %s pushed successful\n" "${QUAY_IMAGE}"
-    fi
-}
-
 # deploy_on_openshift()
 # Deploys the pod with the provided CPaas image, 
 # 
@@ -70,7 +52,8 @@ deploy_on_openshift(){
     oc version > /dev/null
     if [ $? != 0 ]
     then
-        printf "\n ERR || Please install 'oc' client to continue. https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/"
+        printf "\n>ERR || Please install 'oc' client to continue. https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/\n"
+        printf ">ERR || And/OR ensure that the user has been logged in to the cluster.\n"
         exit 1
     fi
 
@@ -93,11 +76,32 @@ deploy_on_openshift(){
         IFS="/" read -r POD POD_NAME <<< "${ELEMENTS[0]}"
     fi
 
-    printf "\n-------------------------------------------------------\n>INFO || Pod %s is available now\n" "${POD_NAME}"    
+    printf "\n-------------------------------------------------------\n>INFO || Pod %s is now available. Any additional tests can be performed against it.\n" "${POD_NAME}"
 }
 
-# Creates '/tmp/download_script.sh' script which is executed within the pod.
-# The `download_script.sh` downloads & executes the verify_jenkins.sh inside the pod. 
+# get_image_ready() 
+# pulls the brew image locally, tags & pushes it to quay.
+# 
+get_image_ready(){
+    printf "\n-------------------------------------------------------\n>INFO || Pulling the Jenkins Image\n"
+    IMAGE_ID=$(podman pull "${JENKINS_IMAGE}")
+    if [ -z "${IMAGE_ID}" ]
+    then
+        printf "\n>ERR || Image Pull Failed\n"
+        exit 1
+    else
+        podman tag "${JENKINS_IMAGE}" "${QUAY_IMAGE}"
+        printf "\n-------------------------------------------------------\n>INFO || Pushing the image to 'quay.io/pipeline-integrations'\n"
+        podman push "${QUAY_IMAGE}"
+        printf "\n-------------------------------------------------------\n>INFO || Image %s pushed successful\n" "${QUAY_IMAGE}"
+    fi
+
+    # now that image is available, deploy a pod on OpenShift cluster using same image.
+    deploy_on_openshift 
+}
+
+# Creates '/tmp/download_script.sh' script which is executed within the container.
+# The `download_script.sh` downloads & executes the verify_jenkins.sh inside the container. 
 # 
 execute_verification_script(){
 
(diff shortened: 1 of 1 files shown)


apoorvajagtap:
@coreydaley @divyansh42 @mbharatk Please feel free to test & share your inputs (Needs a cluster ready):
~~~
$ curl -o /tmp/second-phase.sh https://raw.githubusercontent.com/apoorvajagtap/jenkins/deploy_jenkins_openshift_script/scripts/verify-jenkins-on-openshift.sh && chmod +x /tmp/second-phase.sh
$ /tmp/second-phase.sh
~~~

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,117 @@
...
+    # ensure oc exists
+    oc version > /dev/null
+    if [ $? != 0 ]
+    then
+        echo "\n-------------------------------------------------------> Please install `oc` client to continue. https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/"
+        exit 1
+    fi
+
+    # if oc exists, continue
+    echo -e "\n-------------------------------------------------------\n>INFO || Logging in as " ${USER_NAME} " to " ${API_SERVER} "\n"
```
coreydaley, line 61: printf is usually preferred instead of echo, especially when you are including variables in the output.

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,117 @@
...
+set -o pipefail
+
+# Takes input & stores in the respective variables
+# 
+get_input(){
+    read -p "OpenShift's API URL: " API_SERVER
+    read -p "OpenShift's User Name (kubeadmin): " USER_NAME
+    read -p "OpenShift's User Password: " USER_PASSWORD
+    read -p "Jenkins Image built by CPaas: " JENKINS_IMAGE  
+}
```
coreydaley, line 14: This seems like a pain to input one at a time, how about

1. Eliminate the API_SERVER, USER_NAME, USER_PASSWORD
2. Have the user do an export KUBECONFIG
3. have the JENKINS_IMAGE and SHA either be positional arguments to the script, or ENV VARS that are specified (similar to how the verify script does it)

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,117 @@
...
+# 
+get_image_ready(){
+    echo -e "\n-------------------------------------------------------\n>INFO || Pulling the Jenkins Image\n"
+    IMAGE_ID=$(podman pull ${JENKINS_IMAGE})
+    if [ -z $IMAGE_ID ]
+    then
+        echo -e "\n>ERR =======> Image Pull Failed\n"
+        exit 1
+    else
+        podman tag ${JENKINS_IMAGE} ${QUAY_IMAGE}
```
coreydaley, line 40: If we all use the same image we could stomp on each others work. maybe use the commit SHA as the tag when the image is pushed?

coreydaley (changes requested):

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,117 @@
...
+
+    # Create a script, that'll execute the `verify-jenkins.sh` within the pod
+    read -e -p "Enter the relevant commit's SHA: " COMMIT_SHA
+    cat > /tmp/download_script.sh << EOL
+curl https://raw.githubusercontent.com/openshift/jenkins/master/scripts/verify-jenkins.sh | SHA=$COMMIT_SHA sh
+EOL
+
+    # Run jenkins verification script inside the pod & redirect the output to a file.
+    echo -e "\n-------------------------------------------------------\n>INFO || Running the verify-jenkins.sh within the pod\n"
+    oc exec -i $POD_NAME -- bash -s < /tmp/download_script.sh &> /tmp/result.out
```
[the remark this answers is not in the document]
apoorvajagtap: So, when it tries to run the curl command directly doing an `exec`, it kind of performs those downloads on the local machines (and probably tries to look for the plugin files locally instead of within the pod).
I believe this is most likely due to the networking intricacies of `oc exec` (currently learning on that), but, I could see clear differences in the output of:
- Curl via Exec:
~~~
$ oc exec -i jenkins-1-f5wzn -- curl https://raw.githubusercontent.com/openshift/jenkins/master/scripts/verify-jenkins.sh | SHA=$COMMIT_SHA sh
~~~
vs 
- Curl after rsh:
~~~
$ oc rsh jenkins-1-f5wzn                                                                          [±deploy_jenkins_openshift_script ]
sh-4.4$ curl https://raw.githubusercontent.com/openshift/jenkins/master/scripts/verify-jenkins.sh | SHA=$COMMIT_SHA sh
~~~

scripts/verify-jenkins-on-openshift.sh
apoorvajagtap: suree, eliminating those several inputs, would be a better option. On it.

scripts/verify-jenkins-on-openshift.sh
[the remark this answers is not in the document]
coreydaley: That is very weird ...

apoorvajagtap:
@coreydaley thank you for looking through!
I've made the suggested changes. Let me know in case I've missed anything.

apoorvajagtap:
The script would need two args now:
~~~
$ curl -o /tmp/second-phase.sh https://raw.githubusercontent.com/apoorvajagtap/jenkins/deploy_jenkins_openshift_script/scripts/verify-jenkins-on-openshift.sh && chmod +x /tmp/second-phase.sh
$ /tmp/second-phase.sh -i <jenkins_image> -s <commit_sha>
~~~

coreydaley:
I like the script so far, but I think we can improve it

1. Have it run the image locally first and run the verify script inside of it
2. if that passes, THEN have it run the image on an openshift cluster and run the e2e tests against it from the openshift/jenkins repo (make e2e).

Thoughts?

apoorvajagtap:
With latest commit, the script does the following:
1. Runs a container with CPaas image, executes verify_jenkins.sh within the container. 
2. If all tests succeed in 1, deploys a pod on OpenShift using the same CPaas image, & shares the pod details with the user.
3. If the tests fail in 1, exits, and shares the details about a file with the output(errors).

@coreydaley please let me know, if this meets the expected criteria? Or, in order to run e2e tests against the pod, we'll need some additional configurations in the pod?

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,145 @@
...
+
+    if [ -z "${OUTPUT}" ]
+    then
+        exit 1
+    else
+        IFS=" " read -ra ELEMENTS <<< "${OUTPUT}"
+        IFS="/" read -r POD POD_NAME <<< "${ELEMENTS[0]}"
+    fi
+
+    printf "\n-------------------------------------------------------\n>INFO || Pod %s is now available. Any additional tests can be performed against it.\n" "${POD_NAME}"
```
coreydaley, line 79: Can we expand this comment to mention needing to run the `make e2e` command and verify that all tests pass?

scripts/verify-jenkins-on-openshift.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+
+# This script deploys the jenkins pod with provided image (from CPaas) on an 
+# OpenShift cluster & executes verify_jenkins.sh within the newly deployed pod.
```
coreydaley, line 4: This description seems out of date now since it runs the verify script locally

scripts/verify-jenkins-on-openshift.sh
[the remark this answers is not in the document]
apoorvajagtap: I did a little research & this seems to be an expected behavior for containers given the difference in the env within container vs env in a container namespace. As the `curl` via exec is performed within the network namespace, it could not use the complete network configurations of the container, hence the difference.

scripts/verify-jenkins-on-openshift.sh
apoorvajagtap: Did not change it yet, as still contemplating what could be the best way to use this pod. Let me know your thoughts :)

PR #1664, opened 2023-04-21
Updating Jenkins Go tests to use Podman instead of Docker

coreydaley:
The agent-base tests can be run locally, but the server tests must be run on an openshift cluster (or against one, but i haven't investigated how yet). These used to run using a jenkins server that OpenShift had setup which had docker installed, but that instance went away and we had to disable the test. I think with the current changes we should be able to get them running again via some configuration in openshift/release. If these tests look OK we will need to do some coordination between openshift/jenkins and openshift/release to get them enabled.


coreydaley:
/hold
for discussion

/assign @divyansh42 @mbharatk @apoorvajagtap
for discussion

PR #1697, opened 2023-06-26
OCPBUGS-15530: Bump Jenkins , Jenkins Plugins and Go dependencies

coreydaley:
FYI: I have updated the script that lists the plugins in the Jenkins Script Console, so there are some plugins in this pull request that are the same version, but just rearranged in the list.


coreydaley:
/assign @divyansh42 @mbharatk @apoorvajagtap
for lgtm

coreydaley:
/hold
While I research if this fixes any CVEs

apoorvajagtap:
@coreydaley I have been looking through the golang specific CVEs, and I think most of them are fixed with go v1.20.6. Should we consider bumping the go version for 1.20.6 or above?

coreydaley:
/close
in favor of #1739

PR #1708, opened 2023-08-16
Installing Java 17 instead of Java 8

AlbertoPimpo:
Since Java 8 is no more compatible with Jenkins since version 2.361.1 (see [jenkins docs](https://www.jenkins.io/blog/2022/06/28/require-java-11/)), so there is no more reason to install it in the image. 

Instead, since the same version (see [jenkins docs](https://www.jenkins.io/blog/2023/08/01/documentation-transition-to-java-17/)), Jenkins supports Java 17, so it makes sense to install it in the image so that early adopters can already try it.

### DIFF
--- a/2/Dockerfile.localdev
+++ b/2/Dockerfile.localdev
@@ -49,7 +49,7 @@ EXPOSE 8080 50000
 RUN rm /etc/yum.repos.d/*
 COPY contrib/openshift/CentOS-Base.repo /etc/yum.repos.d/CentOS-Base.repo
 RUN curl http://mirror.centos.org/centos/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS && \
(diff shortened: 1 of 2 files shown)


coreydaley:
We can add Java 17 probably, but we can not remove Java 8 yet. There are some users which may still use Java 8 in their jobs so we want it to continue to be available.

AlbertoPimpo:
> We can add Java 17 probably, but we can not remove Java 8 yet. There are some users which may still use Java 8 in their jobs so we want it to continue to be available.

I'm not sure how many people are going to use it. Consider that by default the execution of jobs in master node is de facto disable (configured with 0 executors) and in general is discouraged (see [jenkins docs](https://www.jenkins.io/doc/book/security/controller-isolation/#not-building-on-the-built-in-node)).

coreydaley:
I believe that in our images that is not disabled by default yet.

AlbertoPimpo:
> I believe that in our images that is not disabled by default yet.

You're actually right. I checked and the built-in node is not disabled by default, it is used if it is the only node that you have. 

So, the only case in which you could benefit having Java 8 is the case in which you don't have any external node or k8s integration and you need to run a specific job that can only run on Java 8. Given that, I think it is safe to drop it. 
If you don't think so I will put it back in the PR.

PR #1739, opened 2024-01-17
OCPBUGS-15530: Update Jenkins version, bump plugins, update image versions

### LINKED ISSUE #1733: Upgrade jenkins-version to 2.426.1 LTS
Hi, would it be possible to get a release of the latest LTS release 2.426.1? Have some problems with plugins complaining they are not compatible with 2.414.3. Would prefer to use the openshift image if we can. Thank you!

coreydaley:
Fixes #1733


coreydaley:
/hold
need to add the openshift plugins to the plugins lists once the new ones are published.

PR #1769, opened 2024-04-18
JKNS-384: Update Jenkins base images to RHEL9

avinal:
- update ubi7 to ubi9

Signed-off-by: Avinal Kumar <avinal@redhat.com>

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.21-openshift-4.16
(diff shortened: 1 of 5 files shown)


avinal:
I think the references to RHEL9 should be updated as well in this PR. Please let me know if that should be a different PR.

adambkaplan:
We should aim to use RHEL 9/UBI9 in as many spots as possible. Note that there is a lot in this repo that appears to be way out of date, and isn't used either in OpenShift CI or our downstream build systems.

One thing I'd advise is to copy the `Dockerfile.rhel8` files to `Dockerfile.rhel9`, where you then update the base UBI images. That should help make the transition a bit smoother.

PR #1706, opened 2023-08-10
Updating the oc client and jenkins templates for fixing the smoke tests

ayushsatyam146:
Failing smoke tests

- smoke/features/jenkins-ephemeral.feature:19 Create jenkins using ephemeral template : [JKNS-02]
- smoke/features/persistent-jenkins.feature:10 Create jenkins using persistent template : [JKNS-06]
- smoke/features/persistent-test.feature:10 Test persistence of jenkins while using persistent template : [JKNS-09]

Updated the relevant oc client to solve JKNS-09 and updated templates for jenkins ephemeral and jenkins persistent apps to solve JKNS-02 and JKNS-06

PR #1773, opened 2024-04-22
OCPTOOLS-393: Revert Incomplete Template Update

adambkaplan:
Reverts openshift/jenkins#1706, which switched the Jenkins templates to use `Deployment` from `DeploymentConfig`. This was an incomplete and left the template with an invalid pod template spec (container image was empty).


sayan-biswas (approved):

adambkaplan:
/hold

We need to investigate why the sync plugin tests are failing.

PR #1841, opened 2024-08-22
feat: update Java to JDK 21

avinal:
## Changes

- use JDK version 21 in builder images

Signed-off-by: Avinal Kumar <avinal@redhat.com>


sayan-biswas:
#1851 Provides Java 17 update, which supported by the base image right now, Java 21 is not supported. Will address this issue once the base image receives an update. Hence closing this now.

PR #1853, opened 2024-10-04
JKNS-529: Remove java 11 usage

kevydotvinu:
- Remove java-11-openjdk and java-11-openjdk-headless from Dockerfiles


ramessesii2 (commented):
Looks good to me.
There are few instances of 'Java 11' in docs/comments. Can you please check using grep over `Java 11` if they need to be removed as well?  Thanks @kevydotvinu.

ramessesii2 (approved):
/lgtm

PR #2052, opened 2025-11-03
KFLUXMIG-936: Build rhel9 ECP generator

k37y:
- Generate the ECP violation staza from conforma log file


k37y:
@red-hat-konflux

PR #2055, opened 2025-11-04
KFLUXMIG-927: Fix rhel9 UBI repository names

k37y:
Use the ubi9 known rpm repositories from the below links:
https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/known_rpm_repositories.yml#L11627-L11650

Update Red Hat Konflux tekton tasks


k37y:
@red-hat-konflux

PR #2370, opened 2026-06-19
JKNS-1063,JKNS-1060: Remove JDK 17 and Update Java to use JDK 25 by default

kunalmemane:
Changes:

- Remove JDK 17 as it has reached End Of Life
- Update JDK version to 25 in builder images and all other references
- Use JDK 25 as default

### DIFF
--- a/2/contrib/s2i/run
+++ b/2/contrib/s2i/run
@@ -293,15 +293,9 @@ function container_max_memory_bytes() {
 CONTAINER_MEMORY_IN_BYTES=$(container_max_memory_bytes)
 CONTAINER_MEMORY_IN_MB=$((CONTAINER_MEMORY_IN_BYTES/2**20))
 
-export JAVA_VERSION=${USE_JAVA_VERSION:=java-21}
-if [[ "$(uname -m)" == "x86_64" ]]; then
-	alternatives --set java $(alternatives --display java | grep $JAVA_VERSION | awk '/family.*x86_64/ { print $1; }')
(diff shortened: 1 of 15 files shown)


rpms.in.yaml
```diff
@@ -1,5 +1,5 @@
 contentOrigin:
   repofiles:
     - ./ubi.repo
-packages: [dejavu-sans-fonts, wget, rsync, gettext, git, git-lfs, tar, zip, unzip, openssl, bzip2, java-21-openjdk, java-21-openjdk-devel, java-17-openjdk, java-17-openjdk-devel, jq, glibc-locale-source, xmlstarlet, glibc-langpack-en, bc]
+packages: [dejavu-sans-fonts, wget, rsync, gettext, git, git-lfs, tar, zip, unzip, openssl, bzip2, java-25-openjdk, java-25-openjdk-devel, jq, glibc-locale-source, xmlstarlet, glibc-langpack-en, bc]
```
k37y: Keep `java-21-openjdk` as we use it in `INSTALL_PKGS`.

2/contrib/s2i/run
k37y: Shouldn't we use `"${JAVA_VERSION}-openjdk.*x86_64.*priority"` kinda regex like above changes?

slave-base/contrib/bin/run-jnlp-client
k37y: Shouldn't we use `"${JAVA_VERSION}-openjdk.*x86_64.*priority"` kinda regex like above changes?

k37y (approved):

